PR #2, opened 2015-09-24
Add name-server API

hunner:
The netdev_stdlib types
(https://forge.puppetlabs.com/puppetlabs/netdev_stdlib) have a resource
`name_server` that does not have a provider in the cisco puppet module.
The node utils need a library to change the name servers first, so this
commit adds the ability to query, add, and remove nameservers.

It does not yet have vrf support, but if you would like me to try and add it before merging I can.


lib/cisco_node_utils/name_server.rb
```diff
@@ -0,0 +1,57 @@
...
+      create if instantiate
+    end
+
+    def self.nameservers
+      hosts = config_get('dnsclient', 'name_server')
+      return {} if hosts.nil?
+
+      hash = {}
+      hosts.each do |name|
+        hash[name] = name
```
glennmatthews: What's the purpose of this line?

tests/test_name_server.rb
```diff
@@ -0,0 +1,78 @@
...
+  def test_router_create_destroy_multiple
+    id1 = '8.8.8.8'
+    ns1 = Cisco::NameServer.new(id1)
+    id2 = '8.8.4.4'
+    ns2 = Cisco::NameServer.new(id2)
+
+    s = @device.cmd("show runn | i 'name-server'")
+    assert_match(s, /^ip name-server #{id1}$/)
+    assert_match(s, /^ip name-server #{id2}$/)
+
```
glennmatthews, line 81: Somewhere in this test you should exercise the Cisco::NameServer.nameservers API and check that its contents are as expected.

robert-w-gries:
:+1:

lib/cisco_node_utils/command_reference_common.yaml
```diff
@@ -98,6 +98,13 @@ aaa_server_group:
   vrf:
     default_value: "default"
 
+dnsclient:
+  name_server:
+    # NOTE: This does not account for vrf ip name-server
+    config_get: 'show hosts'
```
mikewiebe: When possible I would prefer we use the running-config instead of a feature show command.

n9k# show running-config all | include 'ip name-server'
ip name-server 5.5.5.5 7.7.7.7 55:44::3
n9k#

tests/test_name_server.rb
```diff
@@ -0,0 +1,78 @@
...
+    ns.destroy
+    s = @device.cmd("show runn | i 'name-server #{id}'")
+    refute_match(s, /^ip name-server #{id}$/,
+                 "Error: failed to destroy nameserver #{id}")
+  end
+
+  def test_router_create_destroy_multiple
+    id1 = '8.8.8.8'
+    ns1 = Cisco::NameServer.new(id1)
+    id2 = '8.8.4.4'
```
mikewiebe: Could you add an ipv6 test?

lib/cisco_node_utils/command_reference_common.yaml
```diff
@@ -98,6 +98,13 @@ aaa_server_group:
...
+    config_get_token: '/Name servers.* (\S+)$/'
+    config_set: '<state> ip name-server <ip>'
+
```
mikewiebe, line 107: I will let others comment on this as well, but I think we want a default here even it it's set to empty string indicating not configured.

default_value: ''

lib/cisco_node_utils/name_server.rb
hunner: Whoop, I just forgot to delete it when adding the following line.

lib/cisco_node_utils/command_reference_common.yaml
hunner: Oh yeah, that's much better.

lib/cisco_node_utils/command_reference_common.yaml
```diff
@@ -98,6 +98,18 @@ aaa_server_group:
   vrf:
     default_value: "default"
 
+dnsclient:
+  name_server:
+    config_get: "show running-config all | include 'ip name-server'"
+    config_get_token: '/^ip name-server ([\s\d\.:]+)$/'
+    config_set: '<state> ip name-server <ip>'
+    default_value: ''
+  name_server_vrf:
```
glennmatthews: Doesn't look like name_server_vrf is actually used?

tests/test_name_server.rb
```diff
@@ -0,0 +1,120 @@
...
+    @device.cmd('conf t ; no ip name-server 8.8.8.8 ; end')
+    @device.cmd('conf t ; no ip name-server 2001:4860:4860::8888 ; end')
+    # Flush the cache since we've modified the device outside of the node_utils APIs
+    node.cache_flush
+  end
+
+  # TESTS
+
+  def test_nameserver_create_destroy_single_ipv4
+    assert_equal({}, Cisco::NameServer.nameservers,
```
glennmatthews: Perhaps `assert_empty(Cisco::NameServer.nameservers...`

However, this will fail if the node has any pre-existing DNS configuration, correct? This may not be a safe or desirable assumption to make.

lib/cisco_node_utils/command_reference_common.yaml
hunner: Nope, not yet. I figured I'd add it while I have a regex though.

tests/test_name_server.rb
hunner: I was wondering about how to handle that. I could assume `nameservers` returns a hash containing X, then in each test to add Y verify that X+Y is returned, and after destroy only X is returned again...

glennmatthews: That'd be the most thorough approach (we have something similar in `test_snmp_user.rb` for a similar reason). Alternately as a simpler approach, you could not worry about whatever is pre-existing and only check the specific instances you're testing - something like:

```
refute_includes(Cisco::NameServer.nameservers, id)
...
# create instance ...
...
assert_includes(Cisco::NameServer.nameservers, id)
assert_equal(Cisco::NameServer.nameservers[id], ns)
...
# destroy instance
...
refute_includes(Cisco::NameServer.nameservers, id)
```

mikewiebe:
:+1:

PR #7, opened 2015-10-01
Feature/tunnel

mikewiebe:
New tunnel feature.

Please do not merge this.. this is for developer training only.


lib/cisco_node_utils/tunnel.rb
```diff
@@ -0,0 +1,43 @@
+#
+# NXAPI implementation of Tunnel class
+#
+# Copyright (c) 2014-2015 Cisco and/or its affiliates.
```
chrisvanheuveln, line 4: This should be 2015 only

PR #5, opened 2015-09-28
Added ntp_server

jonnytdevops:
Hey Everyone,

Here is my initial code for ntp_server for the cisco node util gem. I know I'll need to write a test before this can be merged.

Comments are very welcome.

Cheers

JT (Jonny)
Puppet Labs

PR #10, opened 2015-10-03
Added ntp_config

jonnytdevops:


jonnytdevops:
Hi There

Thanks for your feedback.

Do you still need me to make the changes regarding getters/setting, and also the NTP enabling feature now that you are aware that this is a NetDev type for a Puppet Type?

My code is with the Puppet type in mind, which makes no mention of enabling/disabling Cisco features, hence my work was limited to the parameters that were in the type (i.e. source-interface).

Also, regarding the getter/setter query, I was following the design patters from my previous PR which got merged:
https://github.com/cisco/cisco-network-node-utils/pull/5
You will see there that as the CLI configuration is all on one line, I simply did everything in the create method.

Thanks!

glennmatthews:
I think it would be desirable on the node-utils level to have a single 'ntp' class rather than separate 'ntp_server' and 'ntp_config' classes with only a single property each. There may be two separate netdev providers for 'ntp_server' and 'ntp_config' but they can share a single underlying class.

jonnytdevops:
The current design pattern is that the Puppet provider creates an instance of an "equivalent" class in the util library. I'm unsure on how to change to a different pattern, without deviating from our core focus that is Puppet netdev integration, as change this would require a significant formal design undertaking. Perhaps not a decision for me to make personally :)

lib/cisco_node_utils/command_reference_common.yaml
```diff
@@ -459,6 +459,13 @@ inventory:
     config_get: "show inventory"
     config_get_token: ["TABLE_inv", "ROW_inv", 0, "vendorid"]
 
+ntp_config:
+  config:
+    config_get: "show running-config | include 'ntp source-interface'"
```
mikewiebe: include '^ntp source-interface'

tests/test_ntp_config.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def teardown
+    # teardown runs at the end of each test
+    no_ntpconfig
+    super
+  end
+
+  def no_ntpconfig
+    # Turn the feature off for a clean test.
+    @device.cmd('conf t ; no ntp source-interface ethernet 2/1; end')
```
mikewiebe: config('no ntp source-interface ethernet 2/1')

tests/test_ntp_config.rb
```diff
@@ -0,0 +1,56 @@
...
+
+# TestNtpConfig - Minitest for NtpConfig node utility.
+class TestNtpConfig < CiscoTestCase
+  def setup
+    # setup runs at the beginning of each test
+    super
+    no_ntpconfig
+  end
+
+  def teardown
```
mikewiebe: I don't think you need this teardown method since the setup method is called at the beginning of each test and there is no difference between the two.

chrisvanheuveln:
getters/setters: The problem here is that ntp supports several other configurations that should (ideally) simply be additional properties in the ntp class. This class is not easily extensible as it is currently written. When those additional properties are added it will mean rewriting the core of this node utils class and breaking the puppet provider. If you make it extensible now then additional properties can be added piecemeal without breaking the provider.

tests/test_ntp_config.rb
```diff
@@ -0,0 +1,56 @@
...
+    @device.cmd('conf t ; no ntp source-interface ethernet 2/1; end')
+    # Flush the cache since we've modified the device outside of
+    # the node_utils APIs
+    node.cache_flush
+  end
+
+  # TESTS
+
+  def test_ntpconfig_create_destroy_single
+    id = 'ethernet 2/1'
```
mikewiebe: We try not to hardcode the ethernet interface name.  Instead you can use interfaces[0].  Our interface tests use this convention.

https://github.com/cisco/cisco-network-node-utils/blob/develop/tests/test_interface.rb

lib/cisco_node_utils/command_reference_common.yaml
chrisvanheuveln: Mike, see my comment about using 'show running-config ntp all' instead of this pipe.

mikewiebe: Ok, I see that now.  No need to anchor then if using the show command that @chrisvanheuveln  suggested.

jonnytdevops:
Please hold off merging this...few other changes coming

jonnytdevops:
@glennmatthews @chrisvanheuveln 

Folks, I came across an issue where we weren't able to _remove_ settings in my previous code design. So, I've made some changes based on your feedback, as well as my tinkering to see what works. Your feed would be appreciated on my following design changes/issues:
- I've removed the create and destroy methods, as there is nothing to create and destroy (everything is individual parameters)
- I've implemented using getters and setters.
- When there is no source_interface configured on the device, I'm returning a string set to 'none'. I needed to make 'none' a special keyword as this is the only way we can remove the source_interface later on. I.e. Puppet does not play well with an empty strings with arbitrary spaces. Now we can simply do source_interface => 'none'

I look forward to your feedback, but this works well with our providers.

Cheers

JT

hunner:
I would suggest having `nil` be the value that does `no ntp source-interface` rather than a keyword like `'none'` or something. The puppet module can translate `'none'` from the catalog into passing a `nil` to the node utils call, and vice versa.

jonnytdevops:
@hunner agreed. This has now been implemented. And I've replaced 'none' with 'unset' in the Puppet provider as per previous patterns in other modules.

lib/cisco_node_utils/ntp_config.rb
```diff
@@ -0,0 +1,65 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require File.join(File.dirname(__FILE__), 'node')
+
+module Cisco
+  # NtpConfig - node utility class for NTP Config configuration management
+  class NtpConfig < NodeUtil
+    attr_reader :name, :source_interface
```
glennmatthews: You don't need an attr_reader for `:source_interface` since you are explicitly defining the function with `def source_interface` below.

lib/cisco_node_utils/ntp_config.rb
```diff
@@ -0,0 +1,65 @@
...
+      source_interface
+    end
+
+    def source_interface=(val)
+      if val.nil? && !source_interface.nil?
+        config_set('ntp_config',
+                   'source_interface',
+                   state:            'no',
+                   source_interface: source_interface)
+      else
```
glennmatthews: should this be `elsif !val.nil?` In the case where val is nil and source_interface is also nil, this setter function should be a no-op.

tests/test_ntp_config.rb
```diff
@@ -0,0 +1,44 @@
...
+class TestNtpConfig < CiscoTestCase
+  def setup
+    # setup runs at the beginning of each test
+    super
+    no_ntpconfig
+  end
+
+  def no_ntpconfig
+    # Turn the feature off for a clean test.
+    config('no ntp source-interface ethernet 2/1')
```
glennmatthews: I think this was a comment on one of the other PRs but it applies here as well - instead of hard-coding an interface name here, do `"no ntp source-interface #{interfaces[0]}"`

tests/test_ntp_config.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  # TESTS
+
+  def test_ntpconfig_create_destroy_single
+    id = 'default'
+    refute_includes(Cisco::NtpConfig.NtpConfigs, id)
```
glennmatthews: This seems like it will fail - NtpConfigs always returns a hash with key 'default' now, right?

tests/test_ntp_config.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    ntp = Cisco::NtpConfig.new(id, false)
+    assert_includes(Cisco::NtpConfig.NtpConfigs, id)
+    assert_equal(Cisco::NtpConfig.NtpConfigs[id], ntp)
```
glennmatthews, line 41: Needs some testing for the `source_interface` and `source_interface=` methods.

glennmatthews:
New design seems much better - thanks!

lib/cisco_node_utils/ntp_config.rb
jonnytdevops: Sure no probs

lib/cisco_node_utils/ntp_config.rb
jonnytdevops: Good catch! I'll get this changed

tests/test_ntp_config.rb
jonnytdevops: No probs

tests/test_ntp_config.rb
jonnytdevops: You are correct. I'll get this removed

jonnytdevops:
@glennmatthews done :)

jonnytdevops:
@glennmatthews can this be merged now? :)

mikewiebe:
:+1:  but looks like you need to resolve some branch conflicts before we can merge.

jonnytdevops:
@mikewiebe @chrisvanheuveln @glennmatthews I have fixed the conflicts. Can this now be merged please? :) Thanks

PR #14, opened 2015-10-05
Adds domain_name API

bmjen:


lib/cisco_node_utils/command_reference_common.yaml
```diff
@@ -108,8 +108,9 @@ dnsclient:
 domain_name:
   domain_name:
     # NOTE: This does not account for vrf ip domain-name
-    config_get: "show running-config"
+    config_get: "show running-config | include 'ip domain-name' | exclude 'use-vrf'"
```
glennmatthews: The 'ip domain-name' command doesn't take a 'use-vrf' parameter. The configuration you want to look out for is:

```
ip domain-name cisco.com
vrf context red
  ip domain-name red.com
vrf context green
  ip domain-name green.com
```

lib/cisco_node_utils/domain_name.rb
```diff
@@ -0,0 +1,54 @@
...
+# limitations under the License.
+
+require File.join(File.dirname(__FILE__), 'node_util')
+
+module Cisco
+  # DomainName- node utility class for domain name configuration
+  class DomainName < NodeUtil
+    attr_reader :name
+
+    def initialize(name, instantiate=true)
```
glennmatthews: In the interest of forward compatibility this should take an optional vrf parameter:

`initialize(name, vrf='default', instantiate=true)`

lib/cisco_node_utils/domain_name.rb
```diff
@@ -0,0 +1,54 @@
...
+      create if instantiate
+    end
+
+    def self.domainnames
+      hash = {}
+      domains = config_get('domain_name', 'domain_name')
+      return hash if domains.nil?
+
+      domains.each do |name|
+        hash[name] = DomainName.new(name, false)
```
glennmatthews: Without VRF support, there will only ever be a single domain-name value. If you want this to be extensible to support VRFs (preferable!), then the hash key should be the VRF (currently only 'default').

tests/test_domain_name.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def teardown
+    # teardown runs at the end of each test
+    no_domainname_test_xyz
+    super
+  end
+
+  def no_domainname_test_xyz
+    # Turn the feature off for a clean test.
+    @device.cmd('conf t ; no ip domain-name test.xyz; end')
```
glennmatthews: We streamlined the underlying code a bit here - you can change this line to simply:

config('no ip domain-name test.xyz')

bmjen:
@glennmatthews Thanks for the feedback! In the interest of getting the current requirements achieved, is it possible to treat this PR as an implementation to match that of the netdev library? We can stage in the provider enhancements to support VRFs in a separate task and PR?

glennmatthews:
My concern is that adding in these hooks for future work is a lot easier to do in the initial implementation than in a subsequent refactor - for example, if the Puppet provider is looking up the existing domain-name entry by name, then when we change the node_util class to index by VRF instead, this will break Puppet. Better to get things right the first time so we don't have to change the API in a non-compatible way.

tests/test_domain_name.rb
```diff
@@ -0,0 +1,56 @@
...
+    no_domainname_test_xyz
+    super
+  end
+
+  def no_domainname_test_xyz
+    # Turn the feature off for a clean test.
+    config('no ip domain-name test.xyz')
+    # Flush the cache since we've modified the device outside of
+    # the node_utils APIs
+    node.cache_flush
```
hunner: I believe the cache_flush is no longer needed.

robert-w-gries: @hunner That's correct.  Since this testcase is inheriting from CiscoTestCase, [cache_flush is called inside the config method](https://github.com/cisco/cisco-network-node-utils/blob/838d95a8e2b771d84b7a63bebfc36d53f843ac69/tests/ciscotest.rb#L58)

lib/cisco_node_utils/domain_name.rb
```diff
@@ -0,0 +1,54 @@
...
+      create if instantiate
+    end
+
+    def self.domainnames
+      hash = {}
+      domains = config_get('dnsclient', 'domain_name')
+      return hash if domains.nil?
+
+      domains.each do |name|
+        hash[name] = DomainName.new(name, false)
```
glennmatthews: Still would prefer this to index by VRF name rather than by domain-name value - for now, just:

``` ruby
hash['default'] = DomainName.new(name, false)
```

tests/test_domain_name.rb
```diff
@@ -0,0 +1,53 @@
...
+  def no_domainname_test_xyz
+    # Turn the feature off for a clean test.
+    config('no ip domain-name test.xyz')
+  end
+
+  # TESTS
+
+  def test_domainname_create_destroy
+    name = 'test.xyz'
+    refute_includes(Cisco::DomainName.domainnames, name)
```
glennmatthews: Since there's only one domain-name at most (and you removed any existing domain-name in setup), a stronger test would be:

`assert_empty(Cisco::DomainName.domainnames)`

tests/test_domain_name.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    domain = Cisco::DomainName.new(name)
+    assert_includes(Cisco::DomainName.domainnames, name)
+    assert_equal(Cisco::DomainName.domainnames[name], domain)
+
+    domain.destroy
+    refute_includes(Cisco::DomainName.domainnames, name)
```
glennmatthews: Here too, `assert_empty` would be a stronger test.

lib/cisco_node_utils/domain_name.rb
mikewiebe: Agreed.  This feature should definitely be vrf aware.  The config below can coexist.

```
ip domain-name mike.cisco.com
```

or

```
vrf context blue
  ip domain-name mike.cisco.com
```

This change would also need a few test cases.

bmjen:
Thanks @glennmatthews! I'll make those updates and work in the VRF compatibility.

lib/cisco_node_utils/domain_name.rb
hunner: I am looking into this for #23; when looking at the README_YAML.md spec on how `config_get` and `config_get_token` keys work, there doesn't seem to be a way to extract both non-vrf and vrf values. It's either/or, and the vrf must be specified ahead of time. Thus we would need to gather all the non-vrf domains, then get a list of vrfs and loop over them to get all of the vrf domains. There is no node-utils api for getting the vrfs yet, so that would have to be written as a prerequisite.

lib/cisco_node_utils/domain_name.rb
hunner: What about `hash["#{vrf}:#{name}"] = DomainName.new(name, vrf, false)` when a vrf is specified? This would eliminate any duplicate entries (since domains will be keyed with their vrfs).

hunner: Actually, I guess for domain names there can only be one per vrf... so my question relates to domain-lists in #23 in which there can be multiple per vrf...

lib/cisco_node_utils/domain_name.rb
mikewiebe: @hunner  Actually, thinking about this more, the _template feature will not work for this so ignore the comment below.

You can use the `_template` feature for this along with `config_get_append`, `config_get_token_append`, and `config_set_append`.  Just search for _template in the README_YAML.md spec.  This way you can extract the same kinds of information under different levels of router config.

lib/cisco_node_utils/domain_name.rb
glennmatthews: With regard to #23, I would prefer `hash[vrf][name]` over `hash["${vrf}:${name}"]` - that's the style used in most of our other types/providers and avoids the possibility of any issues with special characters. `vrf` will just be `'default'` when no vrf is specified.

hunner: I will respond to this in #23, since it pertains to that.

hunner:
```
Finished in 35.483321s, 0.0564 runs/s, 1.0146 assertions/s.
2 runs, 36 assertions, 0 failures, 0 errors, 0 skips
```

PR #15, opened 2015-10-06
Code optimization and improvement

glennmatthews:
Run tests and code coverage through Rake:
- Use SimpleCov to measure test coverage
- Get rid of `test_all_cisco.rb` - `rake test` will auto-discover tests.
- minitest can now prompt user to enter node/user/pass if not in argv.

Additionally:
- Fix some warnings detected by Rake test
- Fix bug with `negotiate_auto` in `test_interface.rb`
- Use `config()` API throughout all minitests where appropriate
- Move `platform_info` from `lib/` to `tests/`
- Add `bin/check_metric_limits.rb` helper script
- Refactor minitests to improve code quality metrics:
  - Metrics/AbcSize 84 -> 63
  - Metrics/MethodLength 103 -> 86


bin/check_metric_limits.rb
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env ruby-2.1.1
```
robert-w-gries, line 1: What happens when someone has ruby-1.9.3 or ruby-2.1.6?

robert-w-gries:
Warning from Travis build:

> WARNING:  open-ended dependency on simplecov (>= 0, development) is not recommended
>   if simplecov is semantically versioned, use:
>     add_development_dependency 'simplecov', '~> 0'

We should constrain the version to '>=0.9' per [this line from the simplecov Readme](https://github.com/colszowka/simplecov/blame/master/README.md#L12).

cisco_node_utils.gemspec
```diff
@@ -31,5 +31,6 @@ Currently supports NX-OS nodes.
   spec.add_development_dependency 'bundler', '~> 1.7'
   spec.add_development_dependency 'rake', '~> 10.0'
   spec.add_development_dependency 'rubocop', '~> 0.34'
+  spec.add_development_dependency 'simplecov'
```
robert-w-gries: See other comment about constraining simplecov to '>=0.9'

lib/cisco_node_utils/command_reference.rb
```diff
@@ -142,13 +142,12 @@ def valid?
 
   # Builds reference hash for the platform specified in the product id.
   class CommandReference
-    attr_reader :debug
```
robert-w-gries, line 145: Why are you removing these attr_reader's?

tests/basetest.rb
```diff
@@ -28,60 +34,60 @@
...
+# with all child classes, so that we only need to initialize them once.
+
 # TestCase - common base class for all minitest cases in this module.
 #   Most node utility tests should inherit from CiscoTestCase instead.
 class TestCase < Test::Unit::TestCase
-  # rubocop:disable Style/ClassVars
+  # These variables can be set in one of three ways:
+  # 1) ARGV:
+  #   $ ruby basetest.rb -- address username password
+  # 2) NODE environment variable
```
robert-w-gries, line 47: This is a great feature! Can you include it in the README's?

tests/basetest.rb
```diff
@@ -28,60 +34,60 @@
...
-      puts "Only 'debug' is allowed"
-      exit
-    end
-  end
-
-  # setup-once params
   def address
-    process_arguments unless @@address
+    @@address ||= ARGV[0]
+    @@address ||= ENV['NODE'].split(' ')[0] if ENV['NODE']
```
robert-w-gries, line 59: It might make more sense to have ARGV parameters override the environment variable if parameters are present.

For example:

```
set NODE="10.10.10.10 username password"
ruby basetest.rb -- --address=11.11.11.11
```

Would use the username and password from the NODE variable, but it would replace the address with 11.11.11.11

bin/check_metric_limits.rb
glennmatthews, line 1: It should be possible to run with either of those too. I just used this because the default 'ruby' on our ADS servers is ruby 1.8.7, which wasn't working for me. As this is only a helper script (it's not run automatically by anything, and most users won't need it either) I wasn't too concerned about locking it to a specific Ruby version for now.

lib/cisco_node_utils/command_reference.rb
glennmatthews, line 145: They're actually redundant ('rake test' was warning about this) - we explicitly define a `debug` function below which overwrites this attr_reader.

tests/basetest.rb
glennmatthews, line 59: ARGV takes precedence - that's the magic of `||=`.
- if it's not yet set, try to set it from ARGV
- if it's still not set, try to set it from ENV['NODE']
- if it's still not set, prompt the user

cisco_node_utils.gemspec
glennmatthews: Will do.

tests/basetest.rb
glennmatthews, line 47: Added - see new diff.

bin/check_metric_limits.rb
robert-w-gries, line 1: Alright sounds good.

Out of curiosity, does [this line in our gemspec](https://github.com/cisco/cisco-network-node-utils/blob/develop/cisco_node_utils.gemspec#L27) not force ruby 1.9.3 or higher on the ADS server?

tests/basetest.rb
robert-w-gries, line 59: Neat trick :)

tests/test_router_ospf_vrf.rb
```diff
@@ -196,6 +192,7 @@ def test_routerospfvrf_collection_not_empty_valid
       refute_empty(vrfs, 'Error: Collection is empty')
       hv = ospf_h.fetch(routername.to_s)
       next if hv.nil?
+      hv = hv['default']
```
robert-w-gries, line 195: What is the purpose of this change?

glennmatthews, line 195: Needed because of the above refactoring. Previously `test_routerospfvrf_collection_not_empty_valid` had a hash of the form `ospf_h[ospf][property]` while `test_routerospfvrf_collection_router_multi_vrfs` had a hash of the form `ospf_h[ospf][vrf][property]`. When I refactored to let these two tests share the `config_from_hash` function, I had to add the extra `[vrf]` layer to this hash, even though this test only uses vrf `default`.

robert-w-gries:
:+1:

CONTRIBUTING.md
```diff
@@ -14,9 +14,24 @@ Cisco Network Elements support a rich set of features to make networks robust, e
...
-  * Create a minitest for new APIs or new functionality
-  * Make sure your new minitest file(s) are included in `tests/test_all_cisco.rb`
-  * Run all the tests to ensure there was no collateral damage to existing code
+  * Create a minitest script for any new APIs or new functionality
+  * Run all the tests to ensure there was no collateral damage to existing code. There are two ways you can specify the Nexus switch (virtual or physical) to test against when running the full test suite:
+    1. Use the NODE environment variable to specify the address, username, and password:
+
+        ```bash
+        export NODE="192.168.100.1 user password"
+        rake test
```
mikewiebe, line 23: Here you document 'how to run the tests' using 'rake' but in the developer guide you use the old 'ruby' method.  Is there any rational for choosing one over the other or is the main benefit that we no longer need a test_all_cisco.rb?

glennmatthews, line 23: 'rake test' is the way to run the whole test suite. If you want to run a single test file, that's not currently available through rake - so for running an individual script, the old 'ruby' method is still needed.

mikewiebe:
:+1:   Lot of good stuff here @glennmatthews

PR #16, opened 2015-10-06
Git hook enhancements

glennmatthews:
- commit-msg-enforce_style:
  - do not fail due to contents of comment lines
  - if line length constraints are validated, print the lines that are failing
- pre-commit-rubocop:
  - still only enforce `rubocop --lint`
  - however, also do a full `rubocop` run and warn the user if it reports any non-lint issues.


robert-w-gries:
:+1:

PR #17, opened 2015-10-06
The handle Cisco::Node config_get nil

hunner:
The nxapi change to return empty string on ascii empty response does not
propagate the empty string past Cisco::Node.config_get; config_get
returns nil on an empty response. This change handles nil correctly.

### DIFF
--- a/tests/test_name_server.rb
+++ b/tests/test_name_server.rb
@@ -34,14 +34,14 @@ class TestNameServer < CiscoTestCase
 
   def no_nameserver_google
     # Turn the feature off for a clean test.
-    config('no ip name-server 8.8.8.8',
-           'no ip name-server 2001:4860:4860::8888')
+    config('no ip name-server 7.7.7.7',
+           'no ip name-server 2001:4860:4860::7777')
   end
 
   # TESTS
 
   def test_nameserver_create_destroy_single_ipv4
-    id = '8.8.8.8'
+    id = '7.7.7.7'
     refute_includes(Cisco::NameServer.nameservers, id)
 
     ns = Cisco::NameServer.new(id)
@@ -53,7 +53,7 @@ class TestNameServer < CiscoTestCase
   end
 
   def test_nameserver_create_destroy_single_ipv6
-    id = '2001:4860:4860::8888'
+    id = '2001:4860:4860::7777'
     refute_includes(Cisco::NameServer.nameservers, id)
 
     ns = Cisco::NameServer.new(id)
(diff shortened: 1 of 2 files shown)


glennmatthews:
Is there a minitest case we can add that would catch this issue in the future?

hunner:
I'm looking for that at the moment... I thought this was already handled, but either my nexus had tainted config when I ran it for the original merge, or it doesn't cover it yet. I'll comment back when I find out.

hunner:
It appears that my original PR was before https://github.com/cisco/cisco-network-node-utils/commit/80733a118f68deee60a972ff46e915c1a2d8a828#diff-73ac5adb236895c412f98beb3210f365R592 which changed the way Cisco::Node handles empty values after testing but before merging.

hunner:
Updated the tests to be more strict. `6 tests, 23 assertions, 0 failures, 0 errors, 0 skips`

tests/test_name_server.rb
```diff
@@ -44,8 +44,8 @@ def no_nameserver_google
   # TESTS
 
   def test_nameserver_create_destroy_single_ipv4
-    id = '8.8.8.8'
-    refute_includes(Cisco::NameServer.nameservers, id)
+    id = '7.7.7.7'
+    assert_equal({}, Cisco::NameServer.nameservers)
```
glennmatthews: The argument against this change (and why I'd suggested using `includes` in the initial review instead) is the possibility that the UUT may have a default name-server configuration that already exists - unless this test is taking responsibility for removing any pre-existent config, this may fail.

robert-w-gries: assert_empty(Cisco::NameServer.nameservers) is a bit cleaner.

hunner: This is why the tests questionably cover the "empty" case; either the tests tiptoe around pre-existing name-servers (note that this is non-vrf name servers only) and therefore don't test empty cases, or they brazenly assume this is a throwaway system (which in my case it is; it's vagrant) and `assert_empty()`. Which would you like?

mikewiebe: I am in favor of removing any pre-existent config before running the tests.  This is the only way to ensure predictability when running the tests against physical and virtual devices.

tests/test_name_server.rb
```diff
@@ -65,10 +65,10 @@ def test_nameserver_create_destroy_single_ipv6
   end
 
   def test_router_create_destroy_multiple
-    id1 = '8.8.8.8'
-    id2 = '2001:4860:4860::8888'
+    id1 = '7.7.7.7'
+    id2 = '2001:4860:4860::7777'
+    refute_includes(Cisco::NameServer.nameservers, id1)
```
robert-w-gries, line 70: Copy + paste error?  id1 tested twice

mikewiebe:
Just ran into this problem running my minitests and forgot that this was the fix for it.  Good enough to merge and fix the immediate minitest failures.

tests/test_name_server.rb
mikewiebe, line 70: @robert-w-gries I fixed this 

https://github.com/cisco/cisco-network-node-utils/commit/f2ddb8c74414484eb6a8b1cd479ee4c2d692adff

PR #18, opened 2015-10-07
Tri-state prop changes

chrisvanheuveln:
Change these props to use tri-state values
  additional-paths [receive | send] [disable]
  soft-reconfiguration inbound

This is for CSCuw40269. 
I originally implemented these properties as multiple boolean properties (e.g.: cap_add_paths_send & cap_add_paths_send_disable); Jie had the same problem with log_neighbor_changes but coded it as a tri-state instead (:enable, :disable, :inherit). The team voted for the tri-states so this is just converting my props over to do the same. On the plus side it's reduced a lot of puppet code.

I also changed the prop names from cap_add_paths_\* to additional_paths_\* to match what bgp_af used.

Minitest, rubocop are clean. Puppet PR will follow shortly.

### DIFF
--- a/lib/cisco_node_utils/bgp_neighbor_af.rb
+++ b/lib/cisco_node_utils/bgp_neighbor_af.rb
@@ -211,8 +211,7 @@ module Cisco
     def additional_paths_receive
       val = config_get('bgp_neighbor_af', 'additional_paths_receive', @get_args)
       return default_additional_paths_receive if val.nil?
-      return :disable if /disable/.match(val.first)
-      :enable
+      /disable/.match(val.first) ? :disable : :enable
     end
 
     def additional_paths_receive=(val)
@@ -237,8 +236,7 @@ module Cisco
     def additional_paths_send
       val = config_get('bgp_neighbor_af', 'additional_paths_send', @get_args)
       return default_additional_paths_send if val.nil?
-      return :disable if /disable/.match(val.first)
-      :enable
+      /disable/.match(val.first) ? :disable : :enable
     end
 
     def additional_paths_send=(val)
@@ -515,8 +513,7 @@ module Cisco
     def soft_reconfiguration_in
       val = config_get('bgp_neighbor_af', 'soft_reconfiguration_in', @get_args)
       return default_soft_reconfiguration_in if val.nil?
-      return :always if /always/.match(val.first)
-      :enable
+      /always/.match(val.first) ? :always : :enable
     end
 
     def soft_reconfiguration_in=(val)
--- a/lib/cisco_node_utils/bgp_neighbor_af.rb
+++ b/lib/cisco_node_utils/bgp_neighbor_af.rb
@@ -205,64 +205,54 @@ module Cisco
 
     # -----------------------
     # <state> capability additional-paths receive <disable>
-    # Nvgens as True -OR- True with 'disable' keyword
-    def cap_add_paths_receive_get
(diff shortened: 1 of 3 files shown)


lib/cisco_node_utils/bgp_neighbor_af.rb
```diff
@@ -205,64 +205,54 @@ def default_as_override
...
-      set_args_keys(state:   (state ? '' : 'no'),
-                    disable: (disable ? 'disable' : ''))
-      config_set('bgp_neighbor_af', 'cap_add_paths_receive', @set_args)
-    end
-
-    def default_cap_add_paths_receive
-      config_get_default('bgp_neighbor_af', 'cap_add_paths_receive')
+    #   :enable  = capability additional-paths receive
+    #   :disable = capability additional-paths receive disable
+    #   :inherit = no capability additional-paths receive
```
mikewiebe, line 210: Can you add a note to indicate where we inherit from?

lib/cisco_node_utils/command_reference_common_bgp.yaml
```diff
@@ -368,21 +368,15 @@ bgp_neighbor_af:
...
 
-  cap_add_paths_receive:
+  additional_paths_receive:
     config_get_token_append: '/^capability additional-paths receive(?: disable)?/'
     config_set_append: '<state> capability additional-paths receive <disable>'
-    default_value: false
-
-  cap_add_paths_receive_disable:
-    default_value: false
+    default_value: 'inherit'
```
mikewiebe, line 374: Should this be :inherit instead of 'inherit' ?

chrisvanheuveln, line 374: I considered that but ultimately used strings just like Jie's code. Symbols are a bit puppet-centric anyway.

lib/cisco_node_utils/bgp_neighbor_af.rb
```diff
@@ -205,64 +205,54 @@ def default_as_override
...
+    def additional_paths_receive
+      val = config_get('bgp_neighbor_af', 'additional_paths_receive', @get_args)
+      return default_additional_paths_receive if val.nil?
+      return :disable if /disable/.match(val.first)
+      :enable
```
glennmatthews: Maybe:

`/disable/.match(val.first) ? :disable : :enable`

Same suggestion below

lib/cisco_node_utils/command_reference_common_bgp.yaml
glennmatthews, line 374: This is a point we should document in our developers' guide - I know several of the PuppetLabs PRs were initially using symbols in their node_utils implementation. It would be good to explicitly recommend using strings rather than symbols.

glennmatthews:
:+1:

lib/cisco_node_utils/bgp_neighbor_af.rb
chrisvanheuveln, line 210: That doesn't add any value. The comment already indicates the state of the cli when inherit is used.

lib/cisco_node_utils/command_reference_common_bgp.yaml
chrisvanheuveln, line 374: We already have this documented in our Best Practices wiki. I'd like to see that wiki pushed out to the repo's.

lib/cisco_node_utils/bgp_neighbor_af.rb
chrisvanheuveln: done

PR #22, opened 2015-10-12
Fix another bug in snmpuser

glennmatthews:
Refactor auth_password_equal? and priv_password_equal?
Fix a bug in priv_password_equal? (was exiting early if auth_protocol
unset, instead of if priv_protocol is unset).
Raise errors in several cases that should never be hit.
This fixes some intermittent failures in snmpuser minitest.

### DIFF
--- a/lib/cisco_node_utils/snmpuser.rb
+++ b/lib/cisco_node_utils/snmpuser.rb
@@ -202,82 +202,97 @@ module Cisco
       config_get_default('snmp_user', 'engine_id')
     end
 
-    # passwords are hashed and so cannot be retrieved directly, but can be
-    # checked for equality. this is done by creating a fake user with the
+    # Passwords are hashed and so cannot be retrieved directly, but can be
+    # checked for equality. This is done by creating a fake user with the
     # password and then comparing the hashes
-    def auth_password_equal?(passwd, is_localized=false)
-      passwd = passwd.to_s unless passwd.is_a?(String)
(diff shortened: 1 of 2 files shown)


lib/cisco_node_utils/snmpuser.rb
```diff
@@ -205,17 +205,22 @@ def self.default_engine_id
...
     # password and then comparing the hashes
-    def auth_password_equal?(passwd, is_localized=false)
-      passwd = passwd.to_s unless passwd.is_a?(String)
-      return true if passwd.empty? && _auth_sym_to_str(auth_protocol).empty?
-      return false if passwd.empty? || _auth_sym_to_str(auth_protocol).empty?
-      dummypw = passwd
-      pw = nil
+    def auth_password_equal?(input_pw, is_localized=false)
+      input_pw = input_pw.to_s unless input_pw.is_a?(String)
+      # if we provide no password, and no password present, it's a match!
```
RichWellum: if = If

lib/cisco_node_utils/snmpuser.rb
```diff
@@ -205,17 +205,22 @@ def self.default_engine_id
...
+      return true if input_pw.empty? && auth_protocol == :none
+      # if we provide no password, but a password is present, or vice versa...
```
RichWellum: if = If

RichWellum:
Looks good Glenn - just those two nits.

RichWellum:
Looks good Glenn. :+1:

PR #23, opened 2015-10-12
Add domain-list support

hunner:
Domain support is pending, and nameserver support is merged. Here is
search domain support.

### DIFF
--- a/lib/cisco_node_utils/command_reference_common.yaml
+++ b/lib/cisco_node_utils/command_reference_common.yaml
@@ -99,6 +99,18 @@ aaa_server_group:
     default_value: "default"
 
 dnsclient:
+  domain:
+    config_get: "show running-config all"
+    config_get_token: '/^ip domain-list (\S+)$/'
+    config_set: '<state> ip domain-list <name>'
+    default_value: ''
+
+  domain_vrf:
+    config_get: "show running-config all"
+    config_get_token: ['/vrf context <vrf>/', '/ip domain-list (\S+)/']
+    config_set: ['vrf context <vrf>', '<state> ip domain-list <name>']
+    default_value: ''
+
   name_server:
     config_get: "show running-config all | include 'ip name-server' | exclude 'use-vrf'"
     config_get_token: '/^ip name-server ([\s\d\.:]+)$/'
--- a/lib/cisco_node_utils/dns_domain.rb
+++ b/lib/cisco_node_utils/dns_domain.rb
@@ -1,0 +1,79 @@
+#
+# NXAPI implementation of DnsDomain class
+#
+# September 2015, Hunter Haugen
+#
+# Copyright (c) 2015 Cisco and/or its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# "group" is a standard SNMP term but in NXOS "role" is used to serve the
+# purpose of group; thus this provider utility does not create snmp groups
+# and is limited to reporting group (role) existence only.
+
+require File.join(File.dirname(__FILE__), 'node_util')
+
+module Cisco
+  # DnsDomain - node utility class for DNS search domain config management
+  class DnsDomain < NodeUtil
+    attr_reader :name, :vrf
+
+    def initialize(name, vrf=nil, instantiate=true)
+      unless name.is_a? String
+        fail TypeError, "Expected a string, got a #{name.class.inspect}"
+      end
+      @name = name
+      @vrf = vrf
+      create if instantiate
+    end
+
+    def self.dnsdomains(vrf=nil)
+      if vrf.nil?
+        domains = config_get('dnsclient', 'domain')
+      else
+        domains = config_get('dnsclient', 'domain_vrf', vrf: vrf)
+      end
+      return {} if domains.nil?
+
+      hash = {}
+      domains.each do |name|
+        hash[name] = DnsDomain.new(name, vrf, false)
+      end
+      hash
+    end
+
+    def ==(other)
+      (name == other.name) && (vrf == other.vrf)
+    end
+
+    def create
+      if @vrf.nil?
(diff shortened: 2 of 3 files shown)


hunner:
```
Finished tests in 35.328043s, 0.0566 tests/s, 0.5944 assertions/s.
2 tests, 21 assertions, 0 failures, 0 errors, 0 skips
```

lib/cisco_node_utils/dns_domain.rb
```diff
@@ -0,0 +1,64 @@
...
+require File.join(File.dirname(__FILE__), 'node_util')
+
+module Cisco
+  # DnsDomain - node utility class for DNS search domain config management
+  class DnsDomain < NodeUtil
+    attr_reader :name
+
+    def initialize(name, instantiate=true)
+      unless name.is_a? String
+        fail TypeError, "Expected a string, got a #{name.inspect}"
```
robert-w-gries: Inspect will not print helpful information for some classes.  I recommend using class.name instead.

```
irb
1.9.3-p551 :001 > test = 123
 => 123 
1.9.3-p551 :002 > test.inspect
 => "123" 
1.9.3-p551 :003 > test.class.name
 => "Fixnum"
```

lib/cisco_node_utils/dns_domain.rb
```diff
@@ -0,0 +1,64 @@
...
+      create if instantiate
+    end
+
+    def self.dnsdomains
+      domains = config_get('dnsclient', 'domain')
+      return {} if domains.nil?
+
+      hash = {}
+      # Join and split because config_get returns array of strings separated by
+      # spaces (regexes are a subset of PDA)
```
robert-w-gries: Comment doesn't apply to this code.  Copy+paste remnant?

tests/test_dns_domain.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    ns1 = Cisco::DnsDomain.new(id1)
+    ns2 = Cisco::DnsDomain.new(id2)
+    assert_includes(Cisco::DnsDomain.dnsdomains, id1)
+    assert_includes(Cisco::DnsDomain.dnsdomains, id2)
+    assert_equal(Cisco::DnsDomain.dnsdomains[id1], ns1)
+    assert_equal(Cisco::DnsDomain.dnsdomains[id2], ns2)
+
+    ns1.destroy
+    ns2.destroy
```
glennmatthews: Would be more thorough to do:

ns1.destroy
(make sure ns1 is gone but ns2 still exists)
ns2.destroy
(make sure ns1 and ns2 are both gone)

lib/cisco_node_utils/dns_domain.rb
```diff
@@ -0,0 +1,64 @@
...
+# and is limited to reporting group (role) existence only.
+
+require File.join(File.dirname(__FILE__), 'node_util')
+
+module Cisco
+  # DnsDomain - node utility class for DNS search domain config management
+  class DnsDomain < NodeUtil
+    attr_reader :name
+
+    def initialize(name, instantiate=true)
```
robert-w-gries: Can you add an optional vrf parameter, as suggested [in this PR](https://github.com/cisco/cisco-network-node-utils/pull/14#discussion_r41181734)?

hunner:
```
Finished tests in 65.036177s, 0.0615 tests/s, 0.7688 assertions/s.
4 tests, 50 assertions, 0 failures, 0 errors, 0 skips
```

lib/cisco_node_utils/command_reference_common.yaml
```diff
@@ -99,6 +99,18 @@ aaa_server_group:
...
 dnsclient:
+  domain:
+    config_get: "show running-config all"
+    config_get_token: '/^ip domain-list (\S+)$/'
+    config_set: '<state> ip domain-list <name>'
+    default_value: ''
+
+  domain_vrf:
+    config_get: "show running-config all"
+    config_get_token: ['/vrf context %s/', '/ip domain-list (\S+)/']
```
mikewiebe: For consistency use <vrf> here instead of %s and then of course change the call below from

`domains = config_get('dnsclient', 'domain_vrf', vrf)`

to 

`domains = config_get('dnsclient', 'domain_vrf', vrf: @vrf)`

hunner: Hmm. I tried that, but it said `RuntimeError: Given 1 args, but token ["/^vrf context <vrf>$/", "/^\\s+ip domain-list (\\S+)$/"] requires 0`

mikewiebe: Hmm.  We do this for ospf_vrf.

```
vrf:
    config_get_token: ['/^router ospf <name>$/i', '/^vrf\s+(\S+)$/']
    config_set: ["router ospf <name>", "<state> vrf <vrf>", "end"]

vrf_ids = config_get('ospf', 'vrf', name: name)
```

What does your call to config_get look like?  I suggested you use `@vrf` but in this case since it's in the class method it should be just `vrf`

lib/cisco_node_utils/dns_domain.rb
```diff
@@ -0,0 +1,79 @@
...
+      if vrf.nil?
+        domains = config_get('dnsclient', 'domain')
+      else
+        domains = config_get('dnsclient', 'domain_vrf', vrf)
+      end
+      return {} if domains.nil?
+
+      hash = {}
+      domains.each do |name|
+        hash[name] = DnsDomain.new(name, vrf, false)
```
hunner, line 50: @glennmatthews 

> With regard to #23, I would prefer `hash[vrf][name]` over `hash["${vrf}:${name}"]` - that's the style used in most of our other types/providers and avoids the possibility of any issues with special characters. `vrf` will just be `'default'` when no vrf is specified.

I've instead gone a more simple route and just stuck with `hash[name]` as the vrf is required to enumerate the domains, and any API querying for the domains can do a hash-merge between all vrfs as they choose.

There is currently no node-utils library for getting a list of all the vrfs to iterate over, but I don't particularly need it yet. Is that okay?

lib/cisco_node_utils/command_reference_common.yaml
hunner: Odd; it worked. I must have mucked it up last time.

glennmatthews:
I think this looks good enough for now. @mikewiebe can you also take a look at the latest diff?

mikewiebe:
:+1:

glennmatthews:
Please fix the merge conflict (I'm going to look into ways to make this less of an issue in the future) then you can merge the PR (or I can  do so). Thanks!

hunner:
@glennmatthews Nice solution on the lib/cisco_node_utils.rb to avoid future conflicts :). Should be good to go now.

PR #24, opened 2015-10-15
enforce first as property

andy91315:
add enforce-first-as property 

pass static test with rubocop and pass ruby test


andy91315:
TestRouterBgp#test_routerbgp_set_get_enforce_first_as = 5.25 s = .

Finished tests in 473.659993s, 0.1161 tests/s, 0.4391 assertions/s.
55 tests, 208 assertions, 0 failures, 0 errors, 0 skips

mikewiebe:
You still have some rubocop failures you need to resolve.  Click on the 'details' link in the Travis CI box for details.

tests/test_router_bgp.rb
```diff
@@ -311,6 +311,24 @@ def test_routerbgp_default_cluster_id
     bgp.destroy
   end
 
+  def test_routerbgp_set_get_enforce_first_as
+    asnum = 55
```
mikewiebe, line 317: Please test this property under both the default and non-default vrf.  There are examples of this for the other properties.

glennmatthews:
Note that the Travis CI failures are a baseline issue. As long as you don't introduce new failures, this can be ignored.

andy91315:
delete work around and change to enforce-first-as

lib/cisco_node_utils/bgp.rb
```diff
@@ -311,6 +311,22 @@ def default_confederation_id
       config_get_default('bgp', 'confederation_id')
     end
 
+    # Enforce First As (Getter/Setter/Default)
+    def enforce_first_as
+      match = config_get('bgp', 'enforce_first_as', @get_args)
+      match.nil? ? false : default_enforce_first_as
```
glennmatthews, line 317: This seems wrong. Shouldn't it be:

`match.nil? ? false : true`

Otherwise you're explicitly assuming that default_enforce_first_as is `true`.

andy91315, line 317: Yes it is true. Does that relate to the default in yaml?

chrisvanheuveln, line 317: I spoke with Andi about this offline. The other boolean getters in this file are using the same method so I told him to just leave it as is and we'll fix these all in a separate commit.

PR #25, opened 2015-10-16
Add prefix_list property to bgp_neighbor_af

chrisvanheuveln:
This is a boilerplate prop modeled on filter-list and route-map.

rubocop/minitest are clean.

### DIFF
--- a/lib/cisco_node_utils/bgp_neighbor_af.rb
+++ b/lib/cisco_node_utils/bgp_neighbor_af.rb
@@ -317,7 +317,7 @@ module Cisco
       str.strip! unless str.nil?
       if str == default_filter_list_in
         state = 'no'
-        # Current filter list name is required for removal
+        # Current filter-list name is required for removal
         str = filter_list_in
         return if str.nil?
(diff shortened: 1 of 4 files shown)


lib/cisco_node_utils/bgp_neighbor_af.rb
```diff
@@ -452,6 +452,53 @@ def default_next_hop_third_party
...
+      str.shift.strip
+    end
+
+    def prefix_list_in=(str)
+      str.strip! unless str.nil?
+      if str == default_prefix_list_in
+        state = 'no'
+        # Current prefix-list name is required for removal
+        str = prefix_list_in
+        return if str.nil?
```
mikewiebe, line 468: Do you really need this return statement?  Does not look like prefix_list_in will ever return nil?

chrisvanheuveln, line 468: It's possible -- if it doesn't find the cli configured then it returns the default val, and the default val is nil.

PR #28, opened 2015-10-19
Removal of 'feature tunnel' from developer guide

mikewiebe:
This is the removal of the 'feature' example from our developers guide.  This example does not provide real value over the more complex eigrp example so we are removing it.


mikewiebe:
@chrisvanheuveln , thanks pointing out the missed references.  Also, I added back some important yaml information that was only listed under the yaml section for tunnel.

chrisvanheuveln:
:+1:

glennmatthews:
:+1:

PR #29, opened 2015-10-20
maxas-limit property

andy91315:
1. do not know maxas-limit default value
2. ruby test 5 failure and 1 error not related to this property
3. static analysis error: Class has too many lines

### DIFF
--- a/lib/cisco_node_utils/command_reference_common_bgp.yaml
+++ b/lib/cisco_node_utils/command_reference_common_bgp.yaml
@@ -100,6 +100,11 @@ bgp:
     config_set_append: '<state> log-neighbor-changes'
     default_value: false
 
+  maxas_limit:
+    config_get_token_append: '/^maxas-limit (\d+)$/'
+    config_set_append: '<state> maxas-limit <limit>'
+    default_value: 50
+
   neighbor_fib_down_accelerate:
     config_get_token_append: '/^neighbor-down fib-accelerate$/'
     config_set_append: '<state> neighbor-down fib-accelerate'
--- a/tests/test_router_bgp.rb
+++ b/tests/test_router_bgp.rb
@@ -566,6 +566,36 @@ class TestRouterBgp < CiscoTestCase
     bgp.destroy
   end
 
+  def test_routerbgp_set_get_maxas_limit
+    %w(test_default test_vrf).each do |t|
+      if t == 'test_default'
+        asnum = 55
+        vrf = 'default'
+        bgp = RouterBgp.new(asnum)
+      else
+        asnum = 99
+        vrf = 'yamllll'
+        bgp = RouterBgp.new(asnum, vrf)
+      end
+      bgp.maxas_limit = 55
+      assert_equal(55, bgp.maxas_limit,
+                   "vrf #{vrf}: bgp maxas-limit should be set to '55'")
+      bgp.maxas_limit = 50
+      assert_equal(50, bgp.maxas_limit,
+                   "vrf #{vrf}: bgp maxas_limit" \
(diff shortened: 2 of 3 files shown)


lib/cisco_node_utils/command_reference_common_bgp.yaml
```diff
@@ -100,6 +100,11 @@ bgp:
     config_set_append: '<state> log-neighbor-changes'
     default_value: false
 
+  maxas_limit:
+    config_get_token_append: '/^maxas-limit (\d+)$/'
+    config_set_append: '<state> maxas-limit <limit>'
+    default_value: 50
```
chrisvanheuveln, line 106: Where does 50 come from? Your comments indicated that there was no default value. When I configure 'maxas-limit 50' it nvgens (so it's not the default value). Use a tilde to indicate nil instead:

```
  default_value: ~
```

tests/test_router_bgp.rb
```diff
@@ -566,6 +566,36 @@ def test_routerbgp_default_neighbor_fib_down_accelerate
...
+                   "should be set to default value of '50'")
+      bgp.destroy
+    end
+  end
+
+  def test_routerbgp_default_maxas_limit
+    asnum = 55
+    bgp = RouterBgp.new(asnum)
+    assert_equal(50, bgp.maxas_limit,
+                 "bgp maxas-limit default value should be '50'")
```
chrisvanheuveln, line 595: For this test you should:
1. set maxas to some non default value and check the result
2. set maxas to default using the default api, then check the result
   
   e.g:

```
  bgp.maxas_limit = 3
  assert_equal(3, bgp.maxas_limit, ...
  bgp.maxas_limit = bgp.default_maxas_limit
  assert_equal(bgp.default_maxas_limit, bgp.maxas_limit, ...
```

chrisvanheuveln:
Please elaborate on the test failures you're seeing. You shouldn't be getting any minitest failures. For the rubocop failure make this change to test_router_bgp.rb:

```
 # TestRouterBgp - Minitest for RouterBgp class
+# rubocop:disable Metrics/ClassLength
 class TestRouterBgp < CiscoTestCase
```

glennmatthews:
Rather than disabling the Metrics/ClassLength limit altogether for TestRouterBgp, as Chris suggest, I would recommend updating `.rubocop.yml` with whatever the new limit needs to be - something like:

``` diff
 Metrics/ClassLength:
-  Max: 722
+  Max: 743
```

chrisvanheuveln:
@glennmatthews I'm not sure I see the value in that. You're disabling the cop either way. There are more   properties coming for this provider so this will have to be changed again for the next ones.

glennmatthews:
If we don't see any value in the Metrics/ClassLength cop, we should just turn it off globally - rather than disabling it for each individual file that happens to pass some arbitrary limit.

chrisvanheuveln:
Glenn and I discussed this offline and agreed to disable it globally. Andi, please make this change along with your diff:

```
diff --git a/.rubocop.yml b/.rubocop.yml
index a94e12d..8ac8011 100644
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,9 +4,6 @@
 Metrics/AbcSize:
   Max: 62

-Metrics/ClassLength:
-  Max: 722
-
 Metrics/CyclomaticComplexity:
   Max: 17

@@ -28,6 +25,10 @@ Metrics/PerceivedComplexity:
 # Team preferences different from Rubocop defaults:
 #

+# Test files and large providers can get very large
+Metrics/ClassLength:
+  Enabled: false
+
```

PR #33, opened 2015-10-21
Adding AAA Authorization Service node utils class

hunsbea:
Adding AAA Authorization Service node utils class. There are a few rubocop errors from the other AAA classes that were committed before rubocop checks were enabled - those will be handled in a different commit. 

ruby test_aaa_authorization_service.rb -v -- 10.122.197.195 admin admin
Run options: -v -- --seed 48955
31 runs, 125 assertions, 0 failures, 0 errors, 0 skips

### DIFF
--- a/lib/cisco_node_utils/aaa_authorization_service.rb
+++ b/lib/cisco_node_utils/aaa_authorization_service.rb
@@ -1,0 +1,150 @@
+# NXAPI implementation of AaaAuthorizationService class
+#
+# May 2015, Alex Hunsberger
+#
+# Copyright (c) 2015 Cisco and/or its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative 'node_util'
+
+module Cisco
+  # AaaAuthorizationService - node util class for aaa authorization management
+  class AaaAuthorizationService < NodeUtil
+    attr_reader :name, :type
+
+    def initialize(type, name, create=true)
+      fail TypeError unless name.is_a? String
+      fail TypeError unless type.is_a? Symbol
+      # only console and default are supported currently
+      fail ArgumentError unless %w(console default).include? name
+      fail ArgumentError unless
+        %i(commands config_commands ssh_certificate ssh_publickey).include? type
+      @name = name
+      @type = type
+      type_str = AaaAuthorizationService.auth_type_sym_to_str(type)
+
+      return unless create
+
+      config_set('aaa_authorization_service', 'method', '', type_str, name)
+    end
+
+    def self.services
+      servs = {}
+      servs_arr = config_get('aaa_authorization_service', 'services')
+      unless servs_arr.nil?
+        servs_arr.each do |type, name|
+          type = auth_type_str_to_sym(type)
+          servs[type] ||= {}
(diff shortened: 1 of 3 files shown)


tests/test_aaa_authorization_service.rb
```diff
@@ -0,0 +1,1230 @@
...
+require_relative '../lib/cisco_node_utils/aaa_authorization_service'
+
+# TestAaaAuthorizationService - Minitest for AaaAuthorizationService util
+# rubocop:disable ClassLength
+class TestAaaAuthorizationService < CiscoTestCase
+  # Method to pre-configure a valid tacacs server and aaa group.  This
+  # group can be included in the testing such access to the device
+  # never is compromised.
+  def preconfig_tacacs_server_access(group_name, keep=true)
+    @device.cmd('configure terminal')
```
glennmatthews: use `config` instead:

```
def preconfig_tacacs_server_access(group_name, keep=true)
  if keep
    config('tacacs-server key testing123',
           'tacacs-server host 10.122.197.197 key testing123',
...
           'aaa authentication login ascii-authentication,
          )
  else
    config("no aaa group server tacacs+ #{group_name}")
  end
end
```

ditto below. config() saves you work because:
1) You don't have to explicitly include 'config terminal' and 'end'
2) You don't have to explicitly include 'cache_flush'

tests/test_aaa_authorization_service.rb
```diff
@@ -0,0 +1,1230 @@
...
+    service = 'default'
+    aaa_a_service = AaaAuthorizationService.new(type, service)
+
+    # Single group, with method 'unselected'
+    method = :unselected
+    groups = [group0]
+    aaa_a_service.groups_method_set(groups, method)
+    match = "#{type_str} #{service} group #{group0}"
+    line = get_match_line(match)
+    refute_nil(line,
```
glennmatthews: You should be able to use `assert_show_match` instead of this pattern.

tests/test_aaa_authorization_service.rb
```diff
@@ -0,0 +1,1230 @@
...
+  end
+
+  def test_aaaauthorizationservice_create_empty_service
+    assert_raises(ArgumentError) do
+      AaaAuthorizationService.new(:commands, '')
+    end
+  end
+
+  def test_aaaauthorizationservice_create_commands_default
+    feature_tacacs
```
glennmatthews: Consider defining setup and teardown methods to avoid repeated code in each test, i.e.:

```
def setup
  super
  feature_tacacs
end

def teardown
  feature_tacacs(false)
  super
end
```

lib/cisco_node_utils/aaa_authorization_service.rb
```diff
@@ -0,0 +1,150 @@
...
+    #   local | group <group1 [group2, ...]> [local]
+    def groups_method_set(grps, m)
+      fail TypeError unless grps.is_a? Array
+      fail TypeError unless grps.all? { |x| x.is_a? String }
+      fail TypeError unless m.is_a? Symbol
+      # only the following are supported (unselected = blank)
+      fail ArgumentError unless [:local, :unselected].include? m
+
+      # raise "type 'local' not allowed when groups are configured" if
+      #  m == :local and not grps.empty?
```
glennmatthews, line 127: What's the story on this?

lib/cisco_node_utils/command_reference_common.yaml
```diff
@@ -35,6 +35,25 @@ aaa_authentication_login:
     config_set: "%s aaa authentication login mschapv2 enable"
     default_value: false
 
+aaa_authorization_service:
+  groups:
+    config_get: "show aaa authorization all"
+    config_get_token: ["TABLE_cmd_methods", "ROW_cmd_methods"]
+    # this set is only used when there are groups to configure
+    config_set: "%s aaa authorization %s %s group %s %s"
```
glennmatthews, line 43: It'd be nice to use named parameters (https://github.com/cisco/cisco-network-node-utils/blob/develop/docs/README-develop-best-practices.md#yaml7) but that's not critical for this commit.

glennmatthews:
:+1: as work-in-progress feature code.

PR #35, opened 2015-10-22
Added initial cmd_config support

glennmatthews:
The feature is not complete yet, as a few areas were hardcoded.

A general implementation is coming next.


tests/cmd_config.yaml
```diff
@@ -2,50 +2,4 @@
...
-
-feature-int-loopback:
-  command: > 
-          interface loopback0
-            description testloopback
-
-feature-int-portchannel:
-  command: > 
-          interface port-channel100
-            description test-portchannel
```
chrisvanheuveln, line 51: Maybe just comment out those lines above instead of deleting them? I think we'll still want to use them (or a subset) for NX-side testing.

tests/test_command_config.rb
```diff
@@ -86,10 +86,10 @@ def send_device_config(config_cmd_hash)
   def build_int_scale_config(add=true)
     add ? s = '' : s = 'no '
     current_interface = 0
-    num_interfaces = 1024
```
chrisvanheuveln, line 89: Comment this one too. I don't want to lose these after the pending refactoring.

chrisvanheuveln, line 89: :+1:   (please review my comments; you may want to get travis passing too)

PR #36, opened 2015-10-22
change maxas-limit default value to 0

andy91315:
1.The default value is out of the valid range just use it to disable the function.
reference to ebgp-multihop

2.pass the minitest


chrisvanheuveln:
:+1:

PR #38, opened 2015-10-23
Fixes and extensions bgp_af

mikewiebe:
Fixes and extensions needed to support dampening_igp_metric, dampening and networks properties.

Summary
1) Fixed bug in dampen_igp_metric and refactored tests.
2) Added additional helper methods for dampen.
3) Fixed bug for networks causing a failure when is_list is empty.
4) Fixed default value for client-to-client
5) Fixed bugs in next_hop_route_map and additional_paths_selection
6) Refactored to use new streamlined networks code
- rubocop passes
- all minitests pass

### DIFF
--- a/lib/cisco_node_utils/bgp_af.rb
+++ b/lib/cisco_node_utils/bgp_af.rb
@@ -133,7 +133,12 @@ module Cisco
       route_map.strip!
       if route_map.empty?
         state = 'no'
-        route_map = next_hop_route_map
+        # Dummy routemap required if not configured.
+        if next_hop_route_map.empty?
+          route_map = 'dummy_routemap'
+        else
+          route_map = next_hop_route_map
+        end
       end
       set_args_keys(state: state, route_map: route_map)
       config_set('bgp_af', 'next_hop_route_map', @set_args)
@@ -206,7 +211,12 @@ module Cisco
       route_map.strip!
       if route_map.empty?
         state = 'no'
-        route_map = additional_paths_selection
+        # Dummy routemap required if not configured.
+        if additional_paths_selection.empty?
+          route_map = 'dummy_routemap'
+        else
+          route_map = additional_paths_selection
+        end
       end
       set_args_keys(state: state, route_map: route_map)
       config_set('bgp_af', 'additional_paths_selection', @set_args)
--- a/lib/cisco_node_utils/bgp_af.rb
+++ b/lib/cisco_node_utils/bgp_af.rb
@@ -223,12 +223,12 @@ module Cisco
     # dampen_igp_metric
     def dampen_igp_metric
       result = config_get('bgp_af', 'dampen_igp_metric', @get_args)
-      result.nil? ? default_dampen_igp_metric : result.first.to_i
+      result ? result.first.to_i : nil
     end
 
     def dampen_igp_metric=(val)
-      set_args_keys(state: (val == default_dampen_igp_metric) ? 'no' : '',
-                    num:   (val == default_dampen_igp_metric) ? '' : val)
+      set_args_keys(state: (val.nil?) ? 'no' : '',
+                    num:   (val.nil?) ? '' : val)
       config_set('bgp_af', 'dampen_igp_metric', @set_args)
     end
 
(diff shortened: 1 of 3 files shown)


lib/cisco_node_utils/bgp_af.rb
```diff
@@ -273,6 +273,66 @@ def dampening
       val
     end
 
+    # Return true if dampening is enabled, else false.
+    def dampening_state
+      !dampening.nil?
+    end
+
+    # For all of the following dampening getters, haf_time, reuse_time,
```
glennmatthews: s/haf/half/

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -273,6 +273,66 @@ def dampening
...
+    # suppress_time, and max_suppress_time, return nil if dampening
+    # is not configured, but also return nil if a dampening routemap
+    # is configured because they are mutually exclusive.
+    def dampening_half_time
+      return nil if dampening.nil? || dampening_routemap_configured?
```
glennmatthews, line 296: Is there a case where dampening is not nil but dampening_routemap_configured is set?

In other words, if they're mutually exclusive, doesn't `dampening_routemap_configured` imply `dampening.nil?`

mikewiebe, line 296: @glennmatthews 
This one is a little tricky.  I have to two cases where the individual properties, half_time, reuse_time, suppress_time and max_suppress_time will be nil.

1) Dampening is not configured at all (no dampening)
2) Dampening routemap is set.

The getter returns nil when dampening is not configured and the name of the routemap when dampening routemap is configured.

lib/cisco_node_utils/bgp_af.rb
mikewiebe: Fixed

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -273,6 +283,66 @@ def dampening
       val
     end
 
+    # Return true if dampening is enabled, else false.
+    def dampening_state
```
chrisvanheuveln: This is a dead method.

mikewiebe:
Taking care of the rubocop failures now

glennmatthews:
:+1:

chrisvanheuveln:
Just one suggestion that's slightly out of your changeset...

I think you can change this:

```
    def dampening_routemap_configured?
      if dampening_routemap.is_a?(String) && dampening_routemap.size > 0
        true
      else
        false
      end
    end
```

...to this:

```
    def dampening_routemap_configured?
      !dampening_routemap.to_s.empty?
    end
```

...in which case you could just use `!dampening_routemap.to_s.empty?` directly instead of the method calls. We can leave this for another time if you don't want to change it right now. 

:+1:  Otherwise.

tests/test_bgp_af.rb
```diff
@@ -695,414 +732,79 @@ def test_maximum_paths_ibgp
...
+                 "#{dbg} Test 3. Restore the removed networks")
 
-    # Apply config_list
-    bgp_af.networks = config_list
+    # Test: Change route-maps on existing networks
+    should = master.map { |network, rm| [network, "#{rm}_55"] }
+    af.networks = should
+    result = af.networks
+    assert_equal(should.sort, result.sort,
+                 "#{dbg} Test 4. Chane route-map on existing networks")
```
chrisvanheuveln: Chane

tests/test_bgp_af.rb
```diff
@@ -695,414 +732,79 @@ def test_maximum_paths_ibgp
...
-    # Apply config_list
-    bgp_af.networks = bgp_af.networks_delta(should_list)
-
-    # Verify is_list on device
-    should_list.each do |network|
-      assert_includes(bgp_af.networks, network,
-                      "Error: device should contain network #{network}")
+  def network_cmd(af, dbg, afi)
+    # Initial 'should' state
+    if /ipv6/.match(afi)
```
chrisvanheuveln: Nit: you don't really need to pass afi to the method since you can grep for ipv6 in dbg:
`if dbg[/ipv6/]`

Fine if you want to keep it though.

tests/test_bgp_af.rb
mikewiebe: Fixed

tests/test_bgp_af.rb
mikewiebe: Fixed

PR #40, opened 2015-10-23
add maximum-paths to bgp_af

andy91315:
1. pass the minitest default vrf and vrf ipv4 and ipv6 unicast
2. pass static analysis

### DIFF
--- a/tests/test_bgp_af.rb
+++ b/tests/test_bgp_af.rb
@@ -613,7 +613,6 @@ class TestRouterBgpAF < CiscoTestCase
   ## maximum_paths
   ##
   def maximum_paths(asn, vrf, af)
-    /ipv4/.match(af) ? af = %w(ipv4 unicast) : af = %w(ipv6 unicast)
     bgp_af = RouterBgpAF.new(asn, vrf, af)
 
     #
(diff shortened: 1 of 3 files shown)


tests/test_bgp_af.rb
```diff
@@ -610,6 +610,102 @@ def test_dampening
...
+    asn = '55'
+    vrf = 'default'
+    af = 'ipv6 unicast'
+    maximum_paths(asn, vrf, af)
+
+    asn = '55'
+    vrf = 'red'
+    af = 'ipv6 unicast'
+    maximum_paths(asn, vrf, af)
+  end
```
chrisvanheuveln: I know you got this code pattern from other places in this file but in this case I'd prefer that you simplify it. You can replace everything in this method with something like this:

```
vrfs = ['default', 'red']
afs = [%w(ipv4 unicast), %w(ipv6 unicast)]
vrfs.each do |vrf|
  afs.each do |af|
    maximum_paths(55, vrf, af)
  end
end
```

tests/test_bgp_af.rb
```diff
@@ -610,6 +610,102 @@ def test_dampening
...
+    assert_equal(bgp_af.maximum_paths, val,
+                 'Error: maximum paths value not match to set value')
+
+    val = bgp_af.default_maximum_paths
+    bgp_af.maximum_paths = val
+    assert_equal(bgp_af.maximum_paths, val,
+                 'Error: maximum paths value not match to default value')
+  end
+
+  def test_maximum_paths
```
chrisvanheuveln, line 637: I know you got this code pattern from other places in this file but in this case I'd prefer that you simplify it. You can replace everything in this method with a nested loop like this:

```
vrfs = ['default', 'red']
afs = [%w(ipv4 unicast), %w(ipv6 unicast)]
vrfs.each do |vrf|
  afs.each do |af|
    maximum_paths(55, vrf, af)
  end
end
```

That also makes it easier to add additional test cases for other AFs.

These comments apply to the ibgp tests too.

tests/test_bgp_af.rb
```diff
@@ -610,6 +610,102 @@ def test_dampening
   # rubocop:enable Metrics/AbcSize,Metrics/MethodLength
 
   ##
+  ## maximum_paths
+  ##
+  def maximum_paths(asn, vrf, af)
+    /ipv4/.match(af) ? af = %w(ipv4 unicast) : af = %w(ipv6 unicast)
```
chrisvanheuveln: See the comment I left on line 638. If you implement that then you can delete this line on 616.

andy91315:
change the test code

PR #43, opened 2015-10-27
Add redistribute property to bgp_af

chrisvanheuveln:
Add a common Utils.delta_add_remove helper function to build
an add/remove hash for network, redistribute, etc.

Note that this command is modeled after the network property that Mike has already coded.
He or I will refactor network to follow the redistribute model.

rubocop: clean
minitest: clean

### DIFF
--- a/lib/cisco_node_utils/cisco_cmn_utils.rb
+++ b/lib/cisco_node_utils/cisco_cmn_utils.rb
@@ -106,5 +106,21 @@ module Cisco
       network = address + '/' + mask unless mask.nil?
       network
     end
+
+    # Helper to build a hash of add/remove commands for a nested array.
+    # Useful for network, redistribute, etc.
+    #   should: an array of expected cmds (manifest/recipe)
+    #  current: an array of existing cmds on the device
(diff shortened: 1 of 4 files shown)


mikewiebe:
@chrisvanheuveln  From the comment above...

Note that this command is modeled after the network property that Mike has already coded.
He or I will refactor network to follow the redistribute model.

Already refactored: https://github.com/cisco/cisco-network-node-utils/pull/38

I will update to use the library version after your PR gets merged.

lib/cisco_node_utils/cisco_cmn_utils.rb
```diff
@@ -106,5 +106,21 @@ def self.process_network_mask(network)
       network = address + '/' + mask unless mask.nil?
       network
     end
+
+    # Helper to build a hash of add/remove commands for a nested array.
+    # Useful for network, redistribute, etc.
+    #   should: an array of expected cmds (manifest/recipe)
```
mikewiebe, line 112: _nit_ Check comment indentation

PR #44, opened 2015-10-27
YAML refactor to support gRPC

glennmatthews:
Note that this is a pull request to merge code into feature/grpc - not yet to develop.

Substantial refactor of the YAML files for CommandReference. Key features:
- YAML files are autodiscovered from the `cmd_ref/` subdirectory
- Each YAML file describes a feature (and all of its platform variants) instead of each YAML file describing all features for a single platform - more maintainable.
- Token generation (`_append`, printf wildcards, key-value wildcards) is now part of CommandReference instead of being part of Node.
- New YAML constructs to handle API variants (`nxapi:`, `grpc:`) and platform variants (`/N7K/:`, `/N9K/:`, `else:`)

NXAPI minitests are passing except for test_node.rb (which needs some massaging still).


docs/README-develop-best-practices.md
```diff
@@ -43,132 +44,124 @@ This document is intended to assist in developing cisco_node_utils API's that ar
...
+# syslog_settings.yaml
+timestamp:
+  config_get: "show running-config all | include '^logging timestamp'"
+  config_get_token: '/^logging timestamp (.*)$/'
+  config_set: '<state> logging timestamp <units>'
+  default_value: 'seconds'
 ```
 
-### <a name="yaml3">Y3: Avoid nested optional matches.
+### <a name="yaml4">Y4: Avoid nested optional matches.
```
mikewiebe, line 70: Since you are already modifying this file, could you add an example for this?  I meant to, but missed it.

robert-w-gries:
Travis is currently failing to build due to the nxapi gem name change.  Can you manually run rubocop and post the results?

glennmatthews:
Rubocop is passing in full - the git hooks run it automatically for every commit I do.

docs/README-develop-best-practices.md
glennmatthews, line 70: Added.

docs/README-develop-best-practices.md
```diff
@@ -43,132 +44,145 @@ This document is intended to assist in developing cisco_node_utils API's that ar
...
+encryption_password:
+  config_get: "show run tacacs all"
+  config_get_token: '/^tacacs-server key (\d+)\s+(\S+)/'
+  default_value: ""
 ```
 
-### <a name="yaml7">Y7: Use Key-value wildcards instead of Printf-style wildcards.
+### <a name="yaml8">Y8: Use Key-value wildcards instead of Printf-style wildcards.
 
 The following approach is moderately more complex to implement but is more readable in the ruby code and is flexible enough to handle significant platform differences in CLI. It is therefore the recommended approach for new development.
```
chrisvanheuveln: Consider:
"Key-value wildcards are moderately more complex to implement than Printf-style wildcards but they are more readable in the ruby code and are flexible enough to handle significant platform differences in CLI. Key-value wildcards are therefore the recommended approach for new development."

docs/README-develop-best-practices.md
```diff
@@ -239,7 +253,7 @@ Having this logic defined in the common object lets the minitest easily check th
 
 Without this equality operator `==` only passes if they are the same instance object.  With this equality operator `==` passes if they are different objects referring to the same configuration on the node.
```
chrisvanheuveln: "The built-in equality operator `==` returns true only if they are the same instance object.  The `==`method below is used to override the built-in equality operator, which returns true even when they are different objects referring to the same configuration on the node."

lib/cisco_node_utils/cmd_ref/README_YAML.md
```diff
@@ -0,0 +1,458 @@
...
+# vtp.yaml
+domain:
+  config_get: "show vtp status"
+  config_get_token: "domain_name"
+  config_set: "vtp domain %s"
+
+filename:
+  config_get: "show running vtp"
+  config_get_token: '/vtp file (\S+)/'
+  config_set: "%s vtp password %s"
```
mikewiebe: Given that we are recommending wildcard substitution, I would prefer we use it in this and other examples.  Except the example showing Printf-style wildcards of course :)

lib/cisco_node_utils/cmd_ref/README_YAML.md
```diff
@@ -0,0 +1,458 @@
...
+  config_get: 'show running-config interface all'
+  config_get_token: ['/^interface <name>$/i', '/^description (.*)$/']
+  config_set: ['interface <name>', 'description <desc>']
+
+...
+```
+
+### API variants
+
+Any of the attribute properties can be subdivided by client API using the
```
mikewiebe, line 168: I feel like this needs more context.  Can we add a reference pointer that describes the client API options in more detail?

lib/cisco_node_utils/cmd_ref/README_YAML.md
```diff
@@ -0,0 +1,458 @@
...
+
+### Platform variants
+
+Any of the attribute properties can be subdivided by platform product ID string
+using a regexp against the product ID as a key. When one or more regexp keys
+are defined thus, you can also use the special key `else` to provide values
+for all products that do not match any of the given regexps:
+
+```yaml
+# show_version.yaml
```
mikewiebe, line 191: I like this structure much better then our old structure!  Nice improvement.

chrisvanheuveln, line 191: Nice work. Aside from the nits it's good to go.

tests/test_command_reference.rb
```diff
@@ -42,44 +43,40 @@ def write(string)
     @input_file.flush
   end
 
+  def test_data_sanity
+    # Make sure the actual YAML in our library loads for various platforms
+    CommandReference.new('', '')
```
robert-w-gries, line 48: What is this testcase testing, and what are the expected results?

lib/cisco_node_utils/cmd_ref/README_YAML.md
glennmatthews, line 168: Will flesh this out a bit more in the next diff - not sure what kind of reference pointer you are looking for but hopefully the extra detail renders that moot.

tests/test_command_reference.rb
glennmatthews, line 48: This is testing that we can create a CommandReference instance (using the 'live' YAML files) that doesn't refer to any specific API or platform, without any errors being thrown.
The following lines are similarly creating CommandReference instances for various known APIs and platforms.

Basically, it's a sanity check for our YAML files, as if there's any syntax error in any of them, this should catch it.

glennmatthews:
New diffs up to address Mike's and Chris's comments - thanks!

glennmatthews:
Just wanted to note - since we aren't supporting Ruby < 2.0.0 any more, we are now guaranteed to have the new, improved 'Psych' YAML library instead of the older 'Syck' library that older releases used. This let me cut out some of the alternate code paths to handle Syck, and let me re-enable the logic that lets us check whether the YAML entries are alphabetized and complain if they aren't.

lib/cisco_node_utils/cmd_ref/README_YAML.md
```diff
@@ -0,0 +1,457 @@
...
+  config_get: 'show running-config interface all'
+  config_get_token: ['/^interface <name>$/i', '/^description (.*)$/']
+  config_set: ['interface <name>', 'description <desc>']
+
+...
+```
+
+### API variants
+
+Clients for different Cisco platforms may use different APIs. Currently these include NXAPI (used for Cisco Nexus platforms) and gRPC (used for Cisco platforms running IOS XR). Often the CLI or other input/output formats needed will vary between APIs, so the YAML must be able to accomodate this.
```
mikewiebe, line 166: Much better.

lib/cisco_node_utils/cmd_ref/README_YAML.md
```diff
@@ -0,0 +1,457 @@
...
+  config_get: 'show running interface all'
+```
+
+### `config_get_token`
+
+`config_get_token` can be a single string, a single regex, an array of strings,
+or an array of regexs.
+
+If this value is a string or array of strings, then the `config_get` command
+will be executed to produce _structured_ output and the string(s) will be
```
mikewiebe, line 263: We should probably add a word of caution to make sure that _structured_ output works across all target platforms.

lib/cisco_node_utils/cmd_ref/README_YAML.md
```diff
@@ -0,0 +1,457 @@
...
+irb(main):017:0> ref.config_set(name: 'red', state: nil, router_id: '1.1.1.1')
+=> ["router ospf red", " router-id 1.1.1.1"]
+irb(main):019:0> ref.config_set(name: 'red', vrf: 'blue',
+                                state: 'no', router_id: '1.1.1.1')
+=> ["router ospf red", "vrf blue", "no router-id 1.1.1.1"]
+```
+
+### `default_value`
+
+If there is a default value for this attribute when not otherwise specified by
```
mikewiebe: More of a question really, but do we need to add guidance about when to use ~ as a default?

glennmatthews: Yeah seems like a good suggestion - got any verbage to suggest? :)

mikewiebe: Maybe....

For properties that have no default value set the default value to `default_value: ~` which indicates NULL (no value) in YAML.

lib/cisco_node_utils/cmd_ref/interface.yaml
```diff
@@ -0,0 +1,236 @@
...
+    config_get: ~
+    config_set: ~
+    config_get_token: ~
+    default_value: false
+  else:
+    config_get_token_append: '/^negotiate auto$/'
+    config_set_append: "%s negotiate auto"
+    default_value: true
+  test_config_get_regex: [
+  !ruby/regexp '/^\s+no negotiate auto/',
```
mikewiebe: Is the !ruby/regexp still required or can we remove those?

lib/cisco_node_utils/command_reference.rb
```diff
@@ -150,7 +180,53 @@ def self.debug=(value)
       @@debug = value # rubocop:disable Style/ClassVars
     end
 
-    attr_reader :files, :product_id
+    KNOWN_APIS = %w(nxapi grpc)
+
+    def self.hash_merge(input_hash, api, product_id, base_hash=nil)
```
mikewiebe: I think this method could use some documentation.  Lot happening here that is not completely obvious.

lib/cisco_node_utils/cmd_ref/interface.yaml
glennmatthews: It was still required for test_config_get_regex (only) but I've fixed that - new diffs incoming.

lib/cisco_node_utils/cmd_ref/README_YAML.md
glennmatthews, line 263: Added:

**WARNING: structured output, although elegant, may not be supported for all commands or all platforms. Use with caution.**

lib/cisco_node_utils/cmd_ref/README_YAML.md
glennmatthews: Rewrote this section:

### `default_value`

If there is a default value for this attribute when not otherwise specified by
the user, the `default_value` parameter describes it. This can be a string,
boolean, integer, array, or nil.

``` yaml
description:
  default_value: ''

hello_interval:
  default_value: 10

auto_cost:
  default_value: [40, 'Gbps']

ipv4_address:
  # YAML represents nil as ~
  default_value: ~
```

By convention, a `default_value` of `''` (empty string) represents a configurable property that defaults to absent, while a default of `nil` (Ruby) or `~` (YAML) represents a property that has no meaningful default at all.

lib/cisco_node_utils/command_reference.rb
glennmatthews: Done. New diffs are up.

chrisvanheuveln:
:+1:

mikewiebe:
:+1:

PR #46, opened 2015-10-27
change maxas-limit default value to false

andy91315:
1. pass the minitest
2. change the default value to false

### DIFF
--- a/lib/cisco_node_utils/command_reference_common_bgp.yaml
+++ b/lib/cisco_node_utils/command_reference_common_bgp.yaml
@@ -108,7 +108,7 @@ bgp:
   maxas_limit:
     config_get_token_append: '/^maxas-limit (\d+)$/'
     config_set_append: '<state> maxas-limit <limit>'
-    default_value: 0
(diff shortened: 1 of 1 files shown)


lib/cisco_node_utils/command_reference_common_bgp.yaml
```diff
@@ -108,7 +108,7 @@ bgp:
   maxas_limit:
     config_get_token_append: '/^maxas-limit (\d+)$/'
     config_set_append: '<state> maxas-limit <limit>'
-    default_value: 0
+    default_value: false
```
mikewiebe, line 111: Did you update the minitest to test for this value?

andy91315:
in minitest use default_maxas_limit

lib/cisco_node_utils/command_reference_common_bgp.yaml
chrisvanheuveln, line 111: Mike, I reviewed his changes with him prior to PR. Minitest is good to go.

PR #47, opened 2015-10-29
added speed and duplex to interface with test

HelenCampbell:


glennmatthews:
We'd previously excluded these attributes because they were proving problematic to account for differences in platform behavior. @robert-w-gries can you provide specifics of what might be needed here?

robert-w-gries:
@glennmatthews I believe the issue came up when testing the speed attribute across different platforms, particularly the 31xx.

I'll need to test on a machine then provide update on the issue.

glennmatthews:
Nexus 3100 series unusual behavior of 'speed' command:

http://www.cisco.com/c/en/us/td/docs/switches/datacenter/nexus3000/sw/interfaces/6_x/b_Cisco_n3k_Interfaces_Configuration_Guide_602_U11/b_Cisco_n3k_Interfaces_Configuration_Guide_602_U11_chapter_010.html#concept_CBAF604FC1F940919A89BC8A9F287505

robert-w-gries:
Ah now I recall the problem.

This particular section is the issue:

> The process of changing the configuration from 40-Gigabit Ethernet to 10-Gigabit Ethernet is called breakout and the process of changing the configuration from 10-Gigabit Ethernet to Gigabit Ethernet is called breakin. When you break out a 40-Gigabit Ethernet port into 10-Gigabit Ethernet ports, the resulting ports are numbered using a 3-tuple naming convention. For example, the break-out ports of the second 40-Gigabit Ethernet port are numbered as 1/2/1, 1/2/2, 1/2/3, 1/2/4.

After Eth1/2 is renamed to 1/2/1, 1/2/2, 1/2/3, and 1/2/4, our provider is unable to find the eth1/2 interface and errors out.

glennmatthews:
It seems like a easy (but a bit hackish) way to avoid this issue would be:

``` diff
   def speed=(val)
+    if node.product_id =~ /C31\d\d/
+      fail 'Changing interface speed is not permitted on this platform'
+    end
     config_set('interface', 'speed', @name, val)
   rescue Cisco::CliError => e
     raise "[#{@name}] '#{e.command}' : #{e.clierror}"
   end
```

robert-w-gries:
@HelenCampbell please commit the following changes:

```
diff --git a/lib/cisco_node_utils/interface.rb b/lib/cisco_node_utils/interface.rb
index 280efa1..6bdf49c 100644
--- a/lib/cisco_node_utils/interface.rb
+++ b/lib/cisco_node_utils/interface.rb
@@ -267,6 +267,9 @@ module Cisco
     end

     def speed=(val)
+      if node.product_id =~ /C31\d\d/
+        fail 'Changing interface speed is not permitted on this platform'
+      end
       config_set('interface', 'speed', @name, val)
     rescue Cisco::CliError => e
       raise "[#{@name}] '#{e.command}' : #{e.clierror}"
@@ -283,6 +286,9 @@ module Cisco
     end

     def duplex=(val)
+      if node.product_id =~ /C31\d\d/
+        fail 'Changing interface duplex is not permitted on this platform'
+      end
       config_set('interface', 'duplex', @name, val)
     rescue Cisco::CliError => e
       raise "[#{@name}] '#{e.command}' : #{e.clierror}"
```

tests/test_interface.rb
```diff
@@ -467,6 +467,50 @@ def test_interface_mtu_valid
...
+    interface.duplex = 'full'
+    assert_equal('full', interface.duplex)
+    interface.duplex = 'auto'
+    assert_equal('auto', interface.duplex)
+    interface_ethernet_default(interfaces_id[0])
+  end
+
+  def test_interface_duplex_invalid
+    interface = Interface.new(interfaces[0])
+    assert_raises(RuntimeError) { interface.duplex = 'hello' }
```
robert-w-gries, line 504: For a more thorough duplex test, please add the following:

```
    interface = Interface.new(interfaces[0])
+   interface.speed = 1000
    assert_raises(RuntimeError) { interface.duplex = 'hello' }
+   interface_ethernet_default(interfaces_id[0])
```

HelenCampbell:
@glennmatthews @robert-w-gries Thank you both for your input, I've updated with your changes.

robert-w-gries:
:+1: looks good!  We are currently in the middle of a release, so I will merge once that is completed.

lib/cisco_node_utils/interface.rb
```diff
@@ -260,6 +260,44 @@ def default_mtu
...
+      raise "[#{@name}] '#{e.command}' : #{e.clierror}"
+    end
+
+    def default_speed
+      config_get_default('interface', 'speed')
+    end
+
+    def duplex
+      duplex = config_get('interface', 'duplex', @name)
+      return default_speed if speed.nil?
```
tphoney: @HelenCampbell this needs to check default_duplex not default_speed

PR #50, opened 2015-10-29
add advertise l2vpn evpn to bgp af

andy91315:
1. add vxlan feature in the minitest
2. pass the minitest
3. pass static analysis


glennmatthews:
Will not merge until we get the current release out the door.

PR #59, opened 2015-11-03
Support for cisco_os_shim refactor

glennmatthews:
Companion to cisco/cisco-nxapi/pull/3.
- Use `cli_nexus` and `cli_ios_xr` as YAML keywords instead of `nxapi` and `grpc`
- Platform variants now check `platform` (alias of `node.client.platform`) for `:nexus` or `:ios_xr` instead of checking `node.client.api` for `'gRPC'` or `'NXAPI'`

Ran test_command_reference, test_interface, test_command_config against both NXOS and IOS XR and they are passing.

### DIFF
--- a/lib/cisco_node_utils/command_reference.rb
+++ b/lib/cisco_node_utils/command_reference.rb
@@ -189,7 +189,7 @@ module Cisco
       @@debug = value # rubocop:disable Style/ClassVars
     end
 
-    attr_reader :api, :files, :product_id
+    attr_reader :cli, :files, :platform, :product_id
 
     # Constructor.
     # Normal usage is to pass product_id only, in which case all usual YAML
@@ -197,9 +197,13 @@ module Cisco
     # matching the given product_id.
     # For testing purposes (only!) you can pass an explicit list of files to
     # load instead. This list will NOT be filtered further by product_id.
-    def initialize(api, product_id, files=nil)
-      @api = api.downcase
-      @product_id = product_id
+    def initialize(product:  nil,
+                   platform: nil,
+                   cli:      false,
+                   files:    nil)
+      @product_id = product
+      @platform = platform
+      @cli = cli
       @hash = {}
       if files
         @files = files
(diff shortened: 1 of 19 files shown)


tests/test_command_reference.rb
```diff
@@ -45,13 +45,15 @@ def write(string)
...
     # Make sure the actual YAML in our library loads for various platforms
-    CommandReference.new('', '')
-    CommandReference.new('NXAPI', '')
-    CommandReference.new('NXAPI', 'N9K-C9396PX')
-    CommandReference.new('NXAPI', 'N7K-C7010')
-    CommandReference.new('NXAPI', 'N3K-C3064PQ-10GE')
-    CommandReference.new('gRPC', '')
-    CommandReference.new('gRPC', 'R-IOSXRV9000-CH')
+    CommandReference.new
+    CommandReference.new(platform: :nexus, cli: true)
```
mikewiebe, line 49: I think I would prefer we default cli: to true instead of false.  Future cases that don't support cli: can do the explicit set.

glennmatthews, line 49: My reason for defaulting cli (and any future API types) to false is that this makes it easier to add additional API types with minimal impact. Additionally it forces us to explicitly specify which APIs are supported by a given client - think:

``` ruby
CommandReference.new(platform: :ios_xe, relax_ng: true)
CommandReference.new(platform: :ios_xr, yang: true, cli: true)
CommandReference.new(platform: :nexus, cli: true)
```

vs:

``` ruby
CommandReference.new(platform: :ios_xe, relax_ng: true, cli:false, yang: false)
CommandReference.new(platform: :ios_xr, yang: true, relax_ng: false) # cli: true is default so we omit it
CommandReference.new(platform: :nexus, cli: true, yang: false, relax_ng: false)
```

mikewiebe, line 49: Fair enough.

mikewiebe:
:+1:

lib/cisco_node_utils/command_reference.rb
```diff
@@ -261,17 +263,17 @@ def debug(text)
...
     # Given a Hash of command reference data (as read from YAML), does:
-    # - Filter any API-specific data according to the given api string
-    # - Filter any platform-specific data according to the given product_id
+    # - Filter any API-specific data
+    # - Filter any platform-specific data
     # - Inherit data from the given base_hash (if any) and extend/override it
     #   with the given input data.
     # - Append 'config_set_append' data to any existing 'config_set' data
     # - Append 'config_get_token_append' data to 'config_get_token', ditto
-    def self.hash_merge(input_hash, api, product_id, base_hash=nil)
```
robert-w-gries, line 273: Why are you changing hash_merge and value_append from class to instance methods?

robert-w-gries:
:+1:

lib/cisco_node_utils/command_reference.rb
glennmatthews, line 273: Did it so that I don't have to explicitly pass instance.api and instance.product_id as arguments.

However note that in the YAML refactor branch this is different again - this work should be considered as a work-in-progress to get unblocked with but I'll have to sort things out once the YAML enhancements are merged.

PR #61, opened 2015-11-03
Feature/grpc bgp demo

RichWellum:
Minitest happy:

54 runs, 141 assertions, 0 failures, 0 errors, 18 skips
Coverage report generated for MiniTest to /ws/rwellum-rtp/github/cisco-network-node-utils/tests/coverage. 0.0 / 0.0 LOC (100.0%) covered.

Rubocop clean on modified files.

-_\- mode: compilation; default-directory: "/ws/rwellum-rtp/github/cisco-network-node-utils/lib/cisco_node_utils/" -_-
Compilation started at Tue Nov  3 15:25:11

rubocop --format emacs /ws/rwellum-rtp/github/cisco-network-node-utils/

Compilation finished at Tue Nov  3 15:25:39

### DIFF
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -84,7 +84,9 @@ module Cisco
     end
 
     def self.enable(state='')
-      config_set('bgp', 'feature', state: state)
+      # IOS XR has no '[no] feature xyz'.
+      config_set('bgp', 'feature', state: state) if
+        node.client.api == 'NXAPI' || state == 'no'
     end
 
     # Convert BGP ASN ASDOT+ to ASPLAIN
@@ -139,6 +141,7 @@ module Cisco
     def destroy
       vrf_ids = config_get('bgp', 'vrf', asnum: @asnum)
       vrf_ids = ['default'] if vrf_ids.nil?
+      router_bgp(asnum, @vrf, 'no')
       if vrf_ids.size == 1 || @vrf == 'default'
         RouterBgp.enable('no')
       else
--- a/lib/cisco_node_utils/cmd_ref/bgp.yaml
+++ b/lib/cisco_node_utils/cmd_ref/bgp.yaml
@@ -1,109 +1,197 @@
 # bgp.yaml
 ---
 _template:
-  config_get: "show running bgp all"
-  config_get_token: '/^router bgp <asnum>$/'
-  config_get_token_append:
-    - '/^vrf <vrf>$/'
-  config_set: "router bgp <asnum>"
-  config_set_append:
-    - "vrf <vrf>"
+  grpc:
+    config_get: "show running-config router bgp"
+    config_get_token: '/^router bgp <asnum>$/'
+    config_get_token_append:
+      - '/^vrf <vrf>$/'
+    config_set: "router bgp <asnum>"
+    # Note the space before 'vrf' - seems like grpc
+    # is currently indentation sensitive.
+    config_set_append:
+      - " vrf <vrf>"
+  nxapi:
+    config_get: "show running bgp all"
+    config_get_token: '/^router bgp <asnum>$/'
+    config_get_token_append:
+      - '/^vrf <vrf>$/'
+    config_set: "router bgp <asnum>"
+    config_set_append:
+      - "vrf <vrf>"
 
 address_family:
   config_set_append: '<state> address-family <afi> <safi>'
 
 bestpath_always_compare_med:
-  config_get_token_append: '/^bestpath always-compare-med$/'
-  config_set_append: '<state> bestpath always-compare-med'
-  default_value: false
+  grpc:
+    config_get_token_append: '/^bgp bestpath med always$/'
+    config_set_append: ' <state> bgp bestpath med always'
+    default_value: false
+  nxapi:
+    config_get_token_append: '/^bestpath always-compare-med$/'
+    config_set_append: '<state> bestpath always-compare-med'
+    default_value: false
 
 bestpath_aspath_multipath_relax:
-  config_get_token_append: '/^bestpath as-path multipath-relax$/'
-  config_set_append: '<state> bestpath as-path multipath-relax'
-  default_value: false
+  grpc:
+    config_get_token_append: '/^bgp bestpath as-path multipath-relax$/'
+    config_set_append: ' <state> bgp bestpath as-path multipath-relax'
+    default_value: false
+  nxapi:
+    config_get_token_append: '/^bestpath as-path multipath-relax$/'
+    config_set_append: '<state> bestpath as-path multipath-relax'
+    default_value: false
 
 bestpath_compare_routerid:
-  config_get_token_append: '/^bestpath compare-routerid$/'
-  config_set_append: '<state> bestpath compare-routerid'
-  default_value: false
+  grpc:
+    config_get_token_append: '/^bgp bestpath compare-routerid$/'
+    config_set_append: ' <state> bgp bestpath compare-routerid'
+    default_value: false
+  nxapi:
+    config_get_token_append: '/^bestpath compare-routerid$/'
+    config_set_append: '<state> bestpath compare-routerid'
+    default_value: false
 
 bestpath_cost_community_ignore:
-  config_get_token_append: '/^bestpath cost-community ignore$/'
-  config_set_append: '<state> bestpath cost-community ignore'
-  default_value: false
+  grpc:
+    config_get_token_append: '/^bgp bestpath cost-community ignore$/'
+    config_set_append: ' <state> bgp bestpath cost-community ignore'
+    default_value: false
+  nxapi:
+    config_get_token_append: '/^bestpath cost-community ignore$/'
+    config_set_append: '<state> bestpath cost-community ignore'
+    default_value: false
 
 bestpath_med_confed:
-  config_get_token_append: '/^bestpath med confed$/'
-  config_set_append: '<state> bestpath med confed'
-  default_value: false
+  grpc:
+    config_get_token_append: '/^bgp bestpath med confed$/'
+    config_set_append: ' <state> bgp bestpath med confed'
+    default_value: false
+  nxapi:
+    config_get_token_append: '/^bestpath med confed$/'
+    config_set_append: '<state> bestpath med confed'
+    default_value: false
 
 bestpath_med_missing_as_worst:
-  config_get_token_append: '/^bestpath med missing-as-worst$/'
-  config_set_append: '<state> bestpath med missing-as-worst'
-  default_value: false
+  grpc:
+    config_get_token_append: '/^bgp bestpath med missing-as-worst$/'
+    config_set_append: ' <state> bgp bestpath med missing-as-worst'
+    default_value: false
+  nxapi:
+    config_get_token_append: '/^bestpath med missing-as-worst$/'
+    config_set_append: '<state> bestpath med missing-as-worst'
+    default_value: false
 
 bestpath_med_non_deterministic:
-  config_get_token_append: '/^bestpath med non-deterministic$/'
-  config_set_append: '<state> bestpath med non-deterministic'
-  default_value: false
+  nxapi:
+    config_get_token_append: '/^bestpath med non-deterministic$/'
+    config_set_append: '<state> bestpath med non-deterministic'
+    default_value: false
 
 cluster_id:
-  config_get_token_append: '/^cluster-id (\S+)$/'
-  config_set_append: '<state> cluster-id <id>'
-  default_value: ""
+  grpc:
+    config_get_token_append: '/^bgp cluster-id (\S+)$/'
+    config_set_append: ' <state> bgp cluster-id <id>'
+    default_value: ""
+  nxapi:
+    config_get_token_append: '/^cluster-id (\S+)$/'
+    config_set_append: '<state> cluster-id <id>'
+    default_value: ""
 
 confederation_id:
-  config_get_token_append: '/^confederation identifier (\d+.?\d+?)$/'
-  config_set_append: '<state> confederation identifier <id>'
-  default_value: ""
+  grpc:
+    config_get_token_append: '/^bgp confederation identifier (\d+.?\d+?)$/'
+    config_set_append: ' <state> bgp confederation identifier <id>'
+    default_value: ""
+  nxapi:
+    config_get_token_append: '/^confederation identifier (\d+.?\d+?)$/'
+    config_set_append: '<state> confederation identifier <id>'
+    default_value: ""
 
+#TODO regexp is broken
 confederation_peers:
-  config_get_token_append: '/^confederation peers (.*)$/'
-  config_set_append: '<state> confederation peers <peer_list>'
-  default_value: ""
+  grpc:
+    config_get_token_append: '/^bgp confederation peers (.*)$/'
+    config_set_append: ' <state> bgp confederation peers <peer_list>'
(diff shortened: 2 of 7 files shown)


lib/cisco_node_utils/bgp.rb
```diff
@@ -84,7 +84,9 @@ def self.enabled
     end
 
     def self.enable(state='')
-      config_set('bgp', 'feature', state: state)
+      # IOS XR has no '[no] feature xyz'.
+      config_set('bgp', 'feature', state: state) if
+        node.client.api == 'NXAPI' || state == 'no'
```
glennmatthews, line 89: seems to me like the `state == 'no'` check is incorrect - you should _only_ try to invoke the 'feature bgp' command on NXAPI, regardless of state

lib/cisco_node_utils/cmd_ref/bgp.yaml
```diff
@@ -1,109 +1,197 @@
...
+    config_set_append:
+      - " vrf <vrf>"
+  nxapi:
+    config_get: "show running bgp all"
+    config_get_token: '/^router bgp <asnum>$/'
+    config_get_token_append:
+      - '/^vrf <vrf>$/'
+    config_set: "router bgp <asnum>"
+    config_set_append:
+      - "vrf <vrf>"
```
glennmatthews, line 17: Values that are the same for both can be shared between grpc and nxapi:

``` yaml
_template:
  config_get_token: '/^router bgp <asnum>$/'
  config_get_token_append:
    - '/^vrf <vrf>$/'
  config_set: "router bgp <asnum>"
  grpc:
    config_get: "show running-config router bgp"
    # Note the space before 'vrf' - seems like grpc
    # is currently indentation sensitive.
    config_set_append:
      - " vrf <vrf>"
  nxapi:
    config_get: "show running bgp all"
    config_set_append:
      - "vrf <vrf>"
```

lib/cisco_node_utils/cmd_ref/bgp.yaml
```diff
@@ -1,109 +1,197 @@
...
-  config_set_append: '<state> bestpath always-compare-med'
-  default_value: false
+  grpc:
+    config_get_token_append: '/^bgp bestpath med always$/'
+    config_set_append: ' <state> bgp bestpath med always'
+    default_value: false
+  nxapi:
+    config_get_token_append: '/^bestpath always-compare-med$/'
+    config_set_append: '<state> bestpath always-compare-med'
+    default_value: false
```
glennmatthews: ditto - `default_value` is `false` for both so it doesn't need to be repeated. Same suggestion throughout below.

lib/cisco_node_utils/cmd_ref/bgp.yaml
```diff
@@ -1,109 +1,197 @@
...
+  grpc:
+    config_get_token_append: '/^bgp confederation identifier (\d+.?\d+?)$/'
+    config_set_append: ' <state> bgp confederation identifier <id>'
+    default_value: ""
+  nxapi:
+    config_get_token_append: '/^confederation identifier (\d+.?\d+?)$/'
+    config_set_append: '<state> confederation identifier <id>'
+    default_value: ""
 
+#TODO regexp is broken
```
glennmatthews: can you revert this part or comment out the broken part?

lib/cisco_node_utils/cmd_ref/bgp.yaml
```diff
@@ -1,109 +1,197 @@
...
+    config_set_append: '<state> enforce-first-as'
+    default_value: true
 
 feature:
-  config_get: "show running bgp"
-  config_get_token: '/^feature bgp$/'
-  config_set: "<state> feature bgp"
+  grpc:
+    config_get: "show running-config router bgp"
+    config_get_token: '/^router bgp$/'
```
glennmatthews, line 123: I thought we'd discussed that the right thing to do for 'feature' was to make it a no-op on XR, rather than making it a clone of 'router'?

lib/cisco_node_utils/cmd_ref/bgp.yaml
```diff
@@ -1,109 +1,197 @@
...
+  grpc:
+    config_get_token_append: '/^bgp graceful-restart$/'
+    config_set_append: ' <state> bgp graceful-restart'
+    default_value: true
+  nxapi:
+    config_get_token_append: '/^graceful-restart$/'
+    config_set_append: '<state> graceful-restart'
+    default_value: true
 
+#Does not exist in IOS XR
```
glennmatthews: Can you mark this as a 'TODO:'? This is one that will need to be updated once we have proper `exclude_platform` capability in the YAML.

lib/cisco_node_utils/cmd_ref/bgp.yaml
mikewiebe: Yeah.. I think the regexp should be this instead.

(\d+|\d+.\d+)

lib/cisco_node_utils/cmd_ref/bgp.yaml
```diff
@@ -111,30 +199,57 @@ neighbor_fib_down_accelerate:
...
-  config_get_token: '/^router bgp (\d+)$/'
-  config_set: "<state> router bgp <asnum>"
+  grpc:
+    config_get: "show running router bgp"
+    config_get_token: '/^router bgp (\d+)$/'
+    config_set: "<state> router bgp <asnum>"
+  nxapi:
+    config_get: "show running bgp"
+    config_get_token: '/^router bgp (\d+)$/'
+    config_set: "<state> router bgp <asnum>"
```
glennmatthews: shared information:

```
router:
  config_get_token: '/^router bgp (\d+)$/'
  config_set: "<state> router bgp <asnum>"
  grpc:
    config_get: "show running router bgp"
  nxapi:
    config_get: "show running bgp"
```

lib/cisco_node_utils/cmd_ref/bgp.yaml
```diff
@@ -111,30 +199,57 @@ neighbor_fib_down_accelerate:
...
-  default_value: false
+  grpc:
+    config_get: "show running router bgp"
+    config_get_token: ['/^router bgp %s$/i', '/^shutdown$/']
+    config_set: ["router bgp <asnum>", "<state> shutdown"]
+    default_value: false
+  nxapi:
+    config_get: "show running bgp"
+    config_get_token: ['/^router bgp %s$/i', '/^shutdown$/']
+    config_set: ["router bgp <asnum>", "<state> shutdown"]
```
glennmatthews, line 215: shared values yadda yadda

lib/cisco_node_utils/cmd_ref/bgp.yaml
```diff
@@ -153,9 +268,14 @@ timer_bgp_keepalive:
   default_value: 60
 
 timer_bgp_keepalive_hold:
-  config_get_token_append: '/^timers bgp (\d+) (\d+)$/'
-  config_set_append: '<state> timers bgp <keepalive> <hold>'
+  #Looks the same
```
glennmatthews: please fix

tests/test_router_bgp.rb
```diff
@@ -20,17 +20,38 @@
 require_relative 'ciscotest'
 require_relative '../lib/cisco_node_utils/bgp'
 
+XR_NOT_SUPPORTED     = 'Not supported in IOS XR'
+XR_NO_VRF_SUPPORT    = 'Feature not supported in a vrf in IOS XR'
+XR_SUPPORTED_BROKEN  = 'Supported in IOS XR - needs further work'
+
+def test_add_routerid_rdauto(asnum, vrf)
```
glennmatthews: note that any function starting with `test_` is assumed to be a test case by Minitest - if this is just a helper method for other test cases, please change its name.

tests/test_router_bgp.rb
```diff
@@ -130,9 +154,16 @@ def test_routerbgp_create_valid_no_feature
     refute_nil(line, "Error: 'router bgp #{asnum}' not configured")
     bgp.destroy
 
-    s = @device.cmd('show run all | no-more')
-    line = /"feature bgp"/.match(s)
-    assert_nil(line, "Error: 'feature bgp' still configured")
+    if node.client.api == 'gRPC'
+      config('term len 0')
```
glennmatthews: basetest.rb already does this when it connects, so it shouldn't be needed here.

tests/test_router_bgp.rb
```diff
@@ -192,15 +224,19 @@ def test_routerbgp_set_get_bestpath
       assert(bgp.bestpath_cost_community_ignore, "vrf #{vrf}: "\
              'bgp bestpath_cost_community_ignore should be enabled')
       bgp.bestpath_med_confed = true
-      assert(bgp.bestpath_med_confed, "vrf #{vrf}: "\
-             'bgp bestpath_med_confed should be enabled')
+      if node.client.api == 'NXAPI' ||
```
glennmatthews, line 226: These checks are added because only some of the bestpath options are working yet in the node utils for XR, not because of any inherent NX-exclusivity of these options, right? If so, can you please mark these with 'TODO:'?

tests/test_router_bgp.rb
```diff
@@ -282,9 +325,11 @@ def test_routerbgp_set_get_cluster_id
         vrf = 'default'
         bgp = RouterBgp.new(asnum)
       else
+        skip(XR_NO_VRF_SUPPORT) if node.client.api == 'gRPC'
         asnum = 99
         vrf = 'yamllll'
         bgp = RouterBgp.new(asnum, vrf)
+        test_add_routerid_rdauto(asnum, vrf)
```
glennmatthews: shouldn't be needed since you're skipping this test case on XR, right?

Same comment throughout this file.

tests/test_router_bgp.rb
```diff
@@ -282,9 +325,11 @@ def test_routerbgp_set_get_cluster_id
         vrf = 'default'
         bgp = RouterBgp.new(asnum)
       else
+        skip(XR_NO_VRF_SUPPORT) if node.client.api == 'gRPC'
```
glennmatthews: Note that this will cause the entire test to report as skipped (including the default VRF part). If this isn't supported on XR as an OS limitation, the better thing to do would be either to refactor the test into two test cases (and skip only the non-default-VRF test case) or simply do `next` here instead of `skip(reason)` to bypass this section of the loop only.

tests/test_router_bgp.rb
```diff
@@ -342,9 +388,11 @@ def test_routerbgp_set_get_graceful_restart
...
         asnum = 99
         vrf = 'yamllll'
         bgp = RouterBgp.new(asnum, vrf)
+        test_add_routerid_rdauto(asnum, vrf)
```
glennmatthews: ditto unnecessary since skipping already on XR

tests/test_router_bgp.rb
```diff
@@ -389,7 +440,9 @@ def test_routerbgp_default_graceful_restart
     assert_equal(300, bgp.default_graceful_restart_timers_stalepath_time,
                  "bgp graceful restart default timer value should be '300'")
     refute(bgp.default_graceful_restart_helper,
-           'graceful restart helper default value should be enabled = false')
+           'graceful restart helper default value ' \
+           'should be enabled = false') if
+      node.client.api == 'NXAPI'
```
glennmatthews: Ick, that's hard to parse. Should this be marked as a TODO or is this something that just plain doesn't work on XR?

```
if node.client.api == 'NXAPI'
  refute(bgp.default_graceful_restart_helper, ....
end
```

tests/test_router_bgp.rb
```diff
@@ -20,17 +20,38 @@
 require_relative 'ciscotest'
 require_relative '../lib/cisco_node_utils/bgp'
 
+XR_NOT_SUPPORTED     = 'Not supported in IOS XR'
```
glennmatthews, line 23: Are there really so many commands that work in NXOS but not in IOS XR? I though XR was supposed to be the gold standard for BGP.

lib/cisco_node_utils/bgp.rb
RichWellum, line 89: This hits the 'feature' block of yaml amd in the XR no case only, I do a 'no router bgp'. So I am using it in a way. Otherwise changing the ASN testcases don't work well.

lib/cisco_node_utils/cmd_ref/bgp.yaml
RichWellum, line 17: I'll clean this up for the whole file.

lib/cisco_node_utils/cmd_ref/bgp.yaml
RichWellum: Ack.

lib/cisco_node_utils/cmd_ref/bgp.yaml
RichWellum: It's referring to confederation peers below. Will put a better comment in.

lib/cisco_node_utils/cmd_ref/bgp.yaml
RichWellum, line 123: I'm not sure - it seemed reasonable to equate 'no feature bgp' with 'no router bgp' to me in the end.

lib/cisco_node_utils/cmd_ref/bgp.yaml
RichWellum: Ack.

lib/cisco_node_utils/cmd_ref/bgp.yaml
RichWellum: Ack.

lib/cisco_node_utils/cmd_ref/bgp.yaml
RichWellum, line 215: Ack.

lib/cisco_node_utils/cmd_ref/bgp.yaml
RichWellum: Ack.

tests/test_router_bgp.rb
RichWellum, line 23: I think the skip count is high because of what you pointed out before - I am skipping whole test cases. But I'll do an assessment and get back to you once I've changed those tcs.

tests/test_router_bgp.rb

tests/test_router_bgp.rb

tests/test_router_bgp.rb
RichWellum, line 226: No in this case and a few others they worked in IOS XR at the global level (vrf = default) but not under a vrf. I'll add an appropriate comment to these.

tests/test_router_bgp.rb

tests/test_router_bgp.rb
RichWellum: Ack. Applied to whole file.

tests/test_router_bgp.rb

tests/test_router_bgp.rb
RichWellum: Plain doesn't work on XR. Reason I did it like the above is because rubocop complained about usign a guard clause.

lib/cisco_node_utils/bgp.rb
glennmatthews, line 89: OK - not blocking but maybe change the above comment:

```
# IOS XR has no '[no] feature bgp' command so:
#   enable == false: do 'no router bgp' instead
#   enable == true:  no-op
```

glennmatthews:
:+1:

PR #67, opened 2015-11-06
Feature/yaml enhancements

glennmatthews:
YAML usability enhancements (based on work in feature/grpc branch, but separate from the gRPC work now):
- Split monolithic YAML files (command_reference_common.yaml) into per-feature files (cmd_ref/bgp.yaml, cmd_ref/bgp_af.yaml, etc.). Per-platform logic for a feature is now defined in the same feature YAML file, rather than having separate YAML per platform.
  - `cli_nexus` tag is added to identify YAML entries that are exclusive to NXOS CLI clients like NXAPI. Currently this is the only supported API but this is the groundwork needed to let us add `cli_ios_xr` as part of the gRPC work.
  - Properties can branch by product-id regexp, including an `else` clause:
    
    ``` yaml
    negotiate_auto_ethernet:
      kind: boolean
      cli_nexus:
        /(N7K|C3064)/:
          config_get: ~
          config_set: ~
          config_get_token: ~
          default_value: false
        else:
          config_get_token_append: '/^negotiate auto$/'
          config_set_append: "%s negotiate auto"
          default_value: true
    ```
  - Features and properties can be marked as not applicable to a platform or product-id by using the new `_exclude` tag: `_exclude: [/N7K/, /N5K/]`.
- By default, feature attributes are assumed to be single in quantity (and will raise an error if config_get finds more than one match), and will return a single value rather than an array, reducing the need to do `value = value[0]`, `value = value.first`, `value.shift`, etc. in every.single.getter.function. For the small number of attributes that really do want to get multiple matches (`all_interfaces`, `vrfs`, etc.), a new YAML tag `multiple: true` is introduced.
- By default, `config_get()` will now return the defined `default_value` (if any) rather than `nil` if no match is found. This reduces the need to do things like `value = value.nil? config_get_default(...) : value` all over the place. For the small number of attributes that don't want this behavior (and want to get an explicit `nil` if no match is found), a new YAML tag `auto_default: false` is introduced.
- Added an optional `kind` YAML tag (recognized values: `boolean`, `int`, `string`) which will automatically convert the `config_get()` result into the desired data type, avoiding the need to do `value = value.to_i`, `value ? true : false`, and `value = value.strip` all over the place. If this tag is not specified, the previous behavior (return a string and let the caller sort it out) is still supported.

Minitest and rubocop are passing.


RichWellum:
Looks good Glenn!

glennmatthews:
@mikewiebe @chrisvanheuveln @robert-w-gries Anyone else have feedback before I merge this into develop?

lib/cisco_node_utils/bgp.rb
```diff
@@ -368,28 +353,19 @@ def default_confederation_peers
 
     # Graceful Restart Getters
     def graceful_restart
-      match = config_get('bgp', 'graceful_restart', @get_args)
-      match.nil? ? false : true
+      config_get('bgp', 'graceful_restart', @get_args)
```
robert-w-gries, line 356: Note:  It looks like this change fixes a bug where matching 'no graceful-restart' returned true instead of false.  We should merge develop into feature/grpc after this PR is merged.

Yaml for reference:

```
graceful_restart:
  kind: boolean
  config_get_token_append: '/^(no )?graceful-restart$/'
  config_set_append: '<state> graceful-restart'
  default_value: true
```

lib/cisco_node_utils/bgp.rb
```diff
@@ -582,7 +551,7 @@ def default_suppress_fib_pending
     # BGP Timers Getters
     def timer_bgp_keepalive_hold
       match = config_get('bgp', 'timer_bgp_keepalive_hold', @get_args)
-      match.nil? ? default_timer_bgp_keepalive_hold : match.first
+      match.nil? ? default_timer_bgp_keepalive_hold : match
```
robert-w-gries, line 554: The config_get token of this property returns two values.  Shouldn't we return match.last?

```
timer_bgp_keepalive_hold:
  config_get_token_append: '/^timers bgp (\d+) (\d+)$/'
  config_set_append: '<state> timers bgp <keepalive> <hold>'
```

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -264,8 +258,7 @@ def default_advertise_l2vpn_evpn
 
     # dampen_igp_metric
     def dampen_igp_metric
-      result = config_get('bgp_af', 'dampen_igp_metric', @get_args)
-      result ? result.first.to_i : nil
+      config_get('bgp_af', 'dampen_igp_metric', @get_args)
```
robert-w-gries, line 261: This property has the auto_default tag.  From my understanding, we should keep the old code since we can't automatically handle the default value?

```
dampen_igp_metric:
  auto_default: false
  kind: int
  config_get_token_append: '/^dampen-igp-metric (\d+)$/'
  config_set_append: '<state> dampen-igp-metric <num>'
  default_value: 600
```

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -500,8 +491,7 @@ def default_maximum_paths_ibgp
 
     # Build an array of all network commands currently on the device
     def networks
-      cmds = config_get('bgp_af', 'network', @get_args)
-      cmds.nil? ? default_networks : cmds.each(&:compact!)
+      config_get('bgp_af', 'network', @get_args).each(&:compact!)
```
robert-w-gries, line 494: [Per the Ruby documentation](http://ruby-doc.org/core-2.2.0/Array.html#method-i-compact-21), the compact method `Returns nil if no changes were made, otherwise returns the array.`

To avoid returning nil:

```
      cmds = config_get('bgp_af', 'network', @get_args)
-     cmds.nil? ? default_networks : cmds.each(&:compact!)
+     cmds.empty? ? cmds : cmds.each(&:compact!)
```

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -532,8 +522,7 @@ def default_networks
 
     # Build an array of all redistribute commands currently on the device
     def redistribute
-      cmds = config_get('bgp_af', 'redistribute', @get_args)
-      cmds.nil? ? default_redistribute : cmds.each(&:compact!)
+      config_get('bgp_af', 'redistribute', @get_args).each(&:compact!)
```
robert-w-gries, line 525: See comment above.

lib/cisco_node_utils/bgp.rb
chrisvanheuveln, line 554: no, this is a combo command (hence both the keepalive and hold props are in the name) and are always returned together like `[["60", "180"]]` so match.first is really just returning the inner array entry.

This is how they are used:

```
    def timer_bgp_keepalive
      keepalive, _hold = timer_bgp_keepalive_hold
...
    def timer_bgp_holdtime
      _keepalive, hold = timer_bgp_keepalive_hold
```

README.md
```diff
@@ -97,7 +97,7 @@ network node. It provides the base APIs `config_set`, `config_get`, and
 
 ### CommandReference
 
-The `CommandReference` module provides for the abstraction of NX-OS CLI,
+The `CommandReference` class provides for the abstraction of NX-OS CLI,
```
chrisvanheuveln: Should we keep these 'NX-OS' references ? There are several throughout the code that seem out of place now.

lib/cisco_node_utils/bgp_af.rb
chrisvanheuveln, line 261: This raises a related question for me: why the need to specify `auto_default: false` at all? Why not use the presence of an explicit `default_value:` entry to know when to override `auto_default:`?

robert-w-gries, line 261: Ah, I was mistaken.  auto_default causes `config_get` to return nil instead of the default, which works for this property.

README.md
glennmatthews: Good thought - more of them will probably be touched by the gRPC work, but this one makes sense to change now. How's the following sound?

> The `CommandReference` class abstracts away the differences between various supported Node types, be that API differences (CLI vs. YANG), platform differences (NX-OS vs. IOS XR), or hardware differences (Nexus 9xxx vs. Nexus 3xxx). A series of YAML files describe various `feature` groupings, with each file describing a set of `attributes` and the specifics of how to inspect and manage that attribute for any supported Node types. When a `Node` is ...

lib/cisco_node_utils/bgp.rb
glennmatthews, line 356: Yup. In general, if we have a command that NVGENs its 'no' form (because true, not false, is default), then having the token match the 'no' form is key to getting the correct automatic behavior.

lib/cisco_node_utils/bgp_af.rb
chrisvanheuveln, line 494: You're mostly correct but in this case compact returns nil within the scope of the iterator; the end result is that `cmds` is  still returned as an array, changed or otherwise.

lib/cisco_node_utils/bgp_af.rb
glennmatthews, line 261: (edit) Yup, Rob got it.

`auto_default` is used specifically for cases where we have a defined `default_value` but do NOT want the getter to automatically return the `default_value` in the case where the `config_get_token` fails to find a match (and would otherwise return `nil`). 

In this case, from the baseline code, the `default_value` is 600 seconds, but if the regexp fails to match, we want to return `nil` (dampen-igp-metric disabled, I guess?) rather than `600`. I'm not enough of an expert in this feature to understand why this asymmetry.

lib/cisco_node_utils/bgp_af.rb
glennmatthews, line 494: @mikewiebe you wrote this getter method originally I think - can you comment on `compact!` behavior here?

My goal in this PR was to not change any existing behavior, but if this is a bug I can arrange to fix it.

tests/test_vlan.rb
```diff
@@ -205,7 +205,7 @@ def test_vlan_add_remove_interface_valid
     interfaces_added_to_vlan = []
     count = 3
     interfaces.each do |name, interface|
-      next unless interface.name.match(%r{ethernet[0-9/]}) && count > 0
+      next unless interface.name.match(%r{ethernet[0-9/]+$}) && count > 0
```
chrisvanheuveln, line 208: Ugh. Just realized this is a problem for sub-interfaces. We could do:

```
     next unless interface.name.match(%r{ethernet[0-9/\.]+$}) && count > 0
```

lib/cisco_node_utils/bgp_af.rb
robert-w-gries, line 494: > You're mostly correct but in this case compact returns nil within the scope of the iterator; the end result is that cmds is still returned as an array, changed or otherwise.

Ack

tests/test_vlan.rb
glennmatthews, line 208: Actually I made this change specifically to exclude sub-interfaces - we can't set `switchport mode access` on sub-interfaces and I was encountering an error when running this test with an unclean base configuration that had sub-interfaces defined.

lib/cisco_node_utils/bgp_af.rb
chrisvanheuveln, line 261: Ah, okay. Thanks for the explanation.

lib/cisco_node_utils/bgp_af.rb
chrisvanheuveln, line 494: I own the compact change - I refactored the network code when I wrote the redistribute prop.

README.md
chrisvanheuveln: ship it

tests/test_vlan.rb
chrisvanheuveln, line 208: ack.  

:+1: No more comments

lib/cisco_node_utils/bgp_neighbor_af.rb
```diff
@@ -308,9 +303,7 @@ def default_disable_peer_as_check
     # -----------------------
     # <state> filter-list <str> in
     def filter_list_in
-      str = config_get('bgp_neighbor_af', 'filter_list_in', @get_args)
-      return default_filter_list_in if str.nil?
-      str.shift.strip
+      config_get('bgp_neighbor_af', 'filter_list_in', @get_args)
```
robert-w-gries, line 306: Is the `strip` method not needed?  Comment applies to the other string properties.

glennmatthews, line 306: See https://github.com/cisco/cisco-network-node-utils/pull/67/files#diff-73ac5adb236895c412f98beb3210f365R119 - attributes that are explicitly tagged as `kind: string` are stripped automatically by Node.

robert-w-gries:
:+1:

PR #69, opened 2015-11-06
Feature/grpc bgp rd

RichWellum:
IOS XR requires an additional 'rd id' for any created VRF. As that property hasn't been written yet and there is a 'rd auto', change the bgp node utils to add in the 'rd auto' when a VRF is created.

Also lots of fixup to be compatible to the new node utils structure.

Fixes to the mini test to be better about adding router-ids where necessary.

minitest all pass (or skip)
Puppet looks good too.
rubocop all happy except for the early return in shutdown.

### DIFF
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -26,10 +26,15 @@ module Cisco
       fail ArgumentError unless vrf.length > 0
       @asnum = RouterBgp.process_asnum(asnum)
       @vrf = vrf
+      @rd = 'auto'
       if @vrf == 'default'
         @get_args = @set_args = { asnum: @asnum }
       else
-        @get_args = @set_args = { asnum: @asnum, vrf: @vrf }
+        if node.client.api == 'gRPC'
+          @get_args = @set_args = { asnum: @asnum, vrf: @vrf, rd: @rd}
+        else
+          @get_args = @set_args = { asnum: @asnum, vrf: @vrf}
+        end
       end
       create if instantiate
     end
@@ -554,6 +559,8 @@ module Cisco
 
     # Shutdown (Getter/Setter/Default)
     def shutdown
+      # TODO: remove this hack when exclude is available
+      return
       match = config_get('bgp', 'shutdown', @asnum)
       match.nil? ? default_shutdown : true
     end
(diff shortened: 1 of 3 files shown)


lib/cisco_node_utils/bgp.rb
```diff
@@ -26,10 +26,15 @@ def initialize(asnum, vrf='default', instantiate=true)
       fail ArgumentError unless vrf.length > 0
       @asnum = RouterBgp.process_asnum(asnum)
       @vrf = vrf
+      @rd = 'auto'
```
glennmatthews: this should only apply to XR right? Or does NX have auto-rd as well?

lib/cisco_node_utils/cmd_ref/bgp.yaml
```diff
@@ -252,3 +254,5 @@ timer_bgp_keepalive_hold:
 vrf:
   config_get_token_append: '/^vrf\s+(\S+)$/'
   config_set: ["router bgp <asnum>", "<state> vrf <vrf>"]
+  config_set_append:
+    - "  rd <rd>"
```
glennmatthews: XR only?

tests/test_router_bgp.rb
```diff
@@ -71,11 +82,25 @@ def test_routerbgp_collection_empty
...
 
   def test_routerbgp_collection_not_empty
-    config('feature bgp') if platform == :nexus
-    config('router bgp 55',
-           'vrf blue',
-           'vrf red',
-           'vrf white')
+    if platform == :nexus
+      config('feature bgp')
+      config('router bgp 55',
```
glennmatthews: you can merge these two back together now.

```
config('feature bgp',
       'router bgp 55',
```

lib/cisco_node_utils/bgp.rb
RichWellum: XR only - I'll move it down into the XR check below.

lib/cisco_node_utils/cmd_ref/bgp.yaml
RichWellum: yup!

tests/test_router_bgp.rb

glennmatthews:
:+1:

robert-w-gries:
:+1:

PR #72, opened 2015-11-10
Minitest fixes and cleanup for rubocop 0.35.0

chrisvanheuveln:
minitest runs clean now except for radius_server (Alex is fixing).

repo is clean for rubocop 0.35.0.

### DIFF
--- a/lib/cisco_node_utils/platform.rb
+++ b/lib/cisco_node_utils/platform.rb
@@ -164,7 +164,7 @@ module Cisco
       # convert to expected format
       virts_hsh = {}
       virts.each do |serv|
-        # rubocop:disable Style/AlignHash, Style/ExtraSpacing
+        # rubocop:disable Style/AlignHash
         virts_hsh[serv['name']] = {
           'package_info' => { 'name'     => serv['package_name'],
                               'path'     => serv['ova_path'],
@@ -184,7 +184,7 @@ module Cisco
                               'cpu'      => serv['cpu_reservation'],
           },
         }
-        # rubocop:enable Style/AlignHash, Style/ExtraSpacing
+        # rubocop:enable Style/AlignHash
       end
       virts_hsh
     end
--- a/tests/test_interface.rb
+++ b/tests/test_interface.rb
@@ -75,6 +75,14 @@ class TestInterface < CiscoTestCase
     arr.count
   end
 
+  # Helper to check for misc speed change disallowed error messages.
+  def speed_change_disallowed(message)
+    pattern = Regexp.new('(port doesn t support this speed|' \
+                         'Changing interface speed is not permitted)')
+    skip('Skip test: Interface type does not allow config change') if
+         message[pattern]
+  end
+
   def create_interface(ifname=interfaces[0])
     @default_show_command = show_cmd(ifname)
     Interface.new(ifname)
@@ -464,8 +472,8 @@ class TestInterface < CiscoTestCase
       interface.speed = 1000
       assert_equal('1000', interface.speed)
       interface_ethernet_default(interfaces_id[0])
-    rescue RuntimeError => e
-      assert_match(/port doesn t support this speed/, e.message)
+    rescue => e
+      speed_change_disallowed(e.message)
     end
   end
 
@@ -479,6 +487,8 @@ class TestInterface < CiscoTestCase
     interface.speed = 1000
     assert_equal('1000', interface.speed)
     interface_ethernet_default(interfaces_id[0])
+  rescue => e
+    speed_change_disallowed(e.message)
   end
 
   def test_interface_duplex_change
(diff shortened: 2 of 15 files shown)


tests/test_interface.rb
```diff
@@ -75,6 +75,14 @@ def interface_count
     arr.count
   end
 
+  # Helper to check for misc speed change disallowed error messages.
+  def speed_change_disallowed(message)
+    pattern = Regexp.new('(port doesn t support this speed|' \
```
robert-w-gries, line 80: Just to make sure, the error message you're matching against is missing an apostrophe in `doesn't`?

tests/test_interface.rb
```diff
@@ -165,6 +173,9 @@ def validate_get_switchport(inttype_h)
                    "Error: #{interface.name}, switchport mode, default, " \
                    'not correct')
     end
+  rescue => e
```
robert-w-gries, line 176: If possible, can you explicitly rescue a specific Error

robert-w-gries:
Looks like rubocop 0.34.2 will fail after these changes.  We'll need to change the gemspec:

```
- spec.add_development_dependency 'rubocop', '= 0.34.2'
+ spec.add_development_dependency 'rubocop', '~> 0.35'
```

lib/cisco_node_utils/platform.rb
```diff
@@ -164,7 +164,7 @@ def self.virtual_services
       # convert to expected format
       virts_hsh = {}
       virts.each do |serv|
-        # rubocop:disable Style/AlignHash, Style/ExtraSpacing
+        # rubocop:disable Style/AlignHash
```
glennmatthews, line 167: There was a bug previously in rubocop that required us to disable both rules (otherwise Rubocop would get into a recursive loop of trying to satisfy two mutually conflicting rules). That bug should be fixed in 0.35 - if you enable both rules, does this pass now?

tests/test_interface.rb
glennmatthews, line 176: Agreed. Rescuing all exceptions is usually bad practice.

tests/test_interface.rb
```diff
@@ -165,6 +173,9 @@ def validate_get_switchport(inttype_h)
...
+    skip('NX-OS defect: system default switchport nvgens twice') if
+      e.message[/Expected zero.one value/]
```
glennmatthews, line 178: I really thought I'd fixed this issue (by explicitly allowing the command to nvgen twice). Not sure why this is failing now.

tests/test_interface.rb
```diff
@@ -464,8 +475,8 @@ def test_interface_speed_change
       interface.speed = 1000
       assert_equal('1000', interface.speed)
       interface_ethernet_default(interfaces_id[0])
-    rescue RuntimeError => e
-      assert_match(/port doesn t support this speed/, e.message)
+    rescue => e
+      speed_change_disallowed(e.message)
```
glennmatthews, line 479: This doesn't seem like the right approach - so now you rescue ANY exception, and if it matches the expected error message, you report a skip, and otherwise (exception does not match) you just plain pass the test?

tests/test_interface.rb
chrisvanheuveln, line 176: Sure - I interpreted the spec as saying that the default exception is RuntimeError but that doesn't appear to be the case so I'll fix it.

tests/test_platform.rb
```diff
@@ -181,7 +181,7 @@ def test_virtual_services
     vir_hsh_hsh = {}
     unless vir_arr.nil?
       vir_arr.each do |serv|
-        # rubocop:disable Style/AlignHash, Style/ExtraSpacing
+        # rubocop:disable Style/AlignHash
```
glennmatthews, line 184: ditto comment about enabling both rules and seeing what's needed.

tests/test_snmpserver.rb
```diff
@@ -314,7 +314,7 @@ def test_snmpserver_global_enforce_priv_get_enabled
     if device_enabled
       refute_show_match(pattern: /no snmp-server globalEnforcePriv/)
     else
-      assert_show_match(pattern: /no snmp-server globalEnforcePriv/)
+      assert_show_match(pattern: /snmp-server globalEnforcePriv/)
```
glennmatthews, line 317: This doesn't look right - if it's enabled, we expect to NOT see 'no', but if it's disabled, we expect TO see 'yes'? Doesn't seem symmetric.

tests/test_interface.rb
chrisvanheuveln, line 80: It's correct. NX-OS spits it out that way.

lib/cisco_node_utils/platform.rb
chrisvanheuveln, line 167: I was trying to make this a pre-emptive fix for 0.35 but several of these changes are not compatible with 34.2 so I'm reverting the rubocop changes I made for this diff. I'll resubmit with just the minitest fixes.

tests/test_interface.rb
chrisvanheuveln, line 178: I see it semi-consistently on .103 and .109.

tests/test_interface.rb
chrisvanheuveln, line 479: It's not - 1) see prev comment on RuntimeError as default (wrong) ; 2) I need to add a fail/raise following the disallow check.

PR #73, opened 2015-11-10
Add Cisco::UnsupportedError class

glennmatthews:
Features and attributes explicitly excluded by a platform will now
raise a Cisco::UnsupportedError instead of an IndexError.
(IndexError is still raised for flat-out missing data)

Added 'default_only' YAML flag to mark an attribute that has a
default_value but no config_get_token or config_set (such as
several 'interface' attributes). config_get() automatically returns
the default_value for a flagged default_only attribute, and
config_set() will raise a Cisco::UnsupportedError.

Added tests to test_command_reference.rb to cover these new features.
Refactored interface.yaml, interface.rb, and test_interface.rb to take advantage of `default_only`.

test_command_reference and test_interface are passing.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/README_YAML.md
+++ b/lib/cisco_node_utils/cmd_ref/README_YAML.md
@@ -474,17 +474,17 @@ By convention, a `default_value` of `''` (empty string) represents a configurabl
 
 ### `default_only`
 
-Some attributes may be hard-coded on certain platforms in such a way that they have a meaningful `default_value` but no relevant `config_get_token` or `config_set` behavior. On such platforms, the key `default_only: true` can be set, which causes the `config_get()` API to always return the default value and `config_set()` to raise a `Cisco::UnsupportedError`.
+Some attributes may be hard-coded in such a way that they have a meaningful default value but no relevant `config_get_token` or `config_set` behavior. For such attributes, the key `default_only` should be used as an alternative to `default_value`. The benefit of using this key is that it causes the `config_get()` API to always return the default value and `config_set()` to raise a `Cisco::UnsupportedError`.
 
 ```yaml
 negotiate_auto_ethernet:
   kind: boolean
   cli_nexus:
     /(N7K|C3064)/:
-      default_value: false
-      default_only: true
+      # this feature is always off on these platforms and cannot be changed
+      default_only: false
     else:
-      config_get_token_append: '/^negotiate auto$/'
+      config_get_token_append: '/^(no )?negotiate auto$/'
       config_set_append: "%s negotiate auto"
       default_value: true
 ```
@@ -521,13 +521,13 @@ feature_lacp:
 
 ### `multiple`
 
-By default, `config_get_token` should uniquely identify a single configuration entry, and `config_get()` will raise an error if more than one match is found. For a small number of attributes, it may be desirable to permit multiple matches (in particular, '`all_*`' attributes that are used up to look up all interfaces, all VRFs, etc.). For such attributes, you must specifyg `multiple: true`. When this value is `true`, `config_get()` will permit multiple matches and will return an array of matches (even if there is only a single match).
+By default, `config_get_token` should uniquely identify a single configuration entry, and `config_get()` will raise an error if more than one match is found. For a small number of attributes, it may be desirable to permit multiple matches (in particular, '`all_*`' attributes that are used up to look up all interfaces, all VRFs, etc.). For such attributes, you must specify the key `multiple:`. When this key is present, `config_get()` will permit multiple matches and will return an array of matches (even if there is only a single match).
 
 ```yaml
 # interface.yaml
 ---
 all_interfaces:
-  multiple: true
+  multiple:
   config_get_token: '/^interface (.*)/'
 ```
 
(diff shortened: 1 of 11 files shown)


lib/cisco_node_utils/cmd_ref/interface.yaml
```diff
@@ -130,12 +123,8 @@ negotiate_auto_ethernet:
 
 negotiate_auto_other_interfaces:
   kind: boolean
-  config_get: ~
-  config_set: ~
   default_value: false
-  test_config_result:
-    false: RuntimeError
-    true: RuntimeError
+  default_only: yes
```
robert-w-gries: s/yes/true

lib/cisco_node_utils/cmd_ref/README_YAML.md
```diff
@@ -469,6 +472,23 @@ By convention, a `default_value` of `''` (empty string) represents a configurabl
...
+
+```yaml
+negotiate_auto_ethernet:
+  kind: boolean
+  cli_nexus:
+    /(N7K|C3064)/:
+      default_value: false
+      default_only: true
+    else:
+      config_get_token_append: '/^negotiate auto$/'
```
chrisvanheuveln: nit: I just changed this particular regexp to `'/^(no )?negotiate auto$/'`

lib/cisco_node_utils/cmd_ref/README_YAML.md
```diff
@@ -469,6 +472,23 @@ By convention, a `default_value` of `''` (empty string) represents a configurabl
 
 `config_get()` will return the defined `default_value` if the defined `config_get_token` does not match anything on the node. Normally this is desirable behavior, but you can use [`auto_default`](#auto_default) to change this behavior if needed.
 
+### `default_only`
+
+Some attributes may be hard-coded on certain platforms in such a way that they have a meaningful `default_value` but no relevant `config_get_token` or `config_set` behavior. On such platforms, the key `default_only: true` can be set, which causes the `config_get()` API to always return the default value and `config_set()` to raise a `Cisco::UnsupportedError`.
```
chrisvanheuveln: Would it make sense to just use a simple label here or do you have to supply a value? 

I don't see a use case for `default_only: false` so would `default_only:` be sufficient?

robert-w-gries:
Is adding a `default_only` keyword necessary?  It seems to me that we can pre-process the keys of a property and set a flag (ex: `@default_only = true`) if `default_value` is the only key specified for a property.

lib/cisco_node_utils/cmd_ref/interface.yaml
glennmatthews: I think YAML actually supports 'yes' and 'no' and even 'on' and 'off', which is why this didn't fail for me. But I agree this needs to be changed - I'll also add this as a best practice in the README.

lib/cisco_node_utils/cmd_ref/README_YAML.md
glennmatthews: Ack. Will fix.

glennmatthews:
@robert-w-gries The issue I hit was with using templates - since the template provides a baseline `config_get` value (and possibly others) the only ways I can force the other values to nil are either:

1) explicitly set them to nil in the YAML (`config_get: ~`)
2) add a flag that can be used to specify my intent.

If you look at the diffs, the previous interface.yaml was using option 1 but then you run into the issue of it being harder to determine algorithmically whether a value is missing by accident (and raise IndexError) or intentionally excluded (and raise UnsupportedError).

lib/cisco_node_utils/cmd_ref/README_YAML.md
glennmatthews: Good thought. I can _maybe_ see a case where setting it explicitly to `false` is desirable (say it's default_only for all platforms with _one_ exception?) and I think it'd be confusing to have `default_only: false` be ignored and have the flag still set to true.

However, I can certainly change this one so that `default_only:` implies `default_only: true` - and make the same change to allow `multiple:` while I'm at it. Sound reasonable?

robert-w-gries: If I understand correctly, you're saying the following example YAML is possible:

```
feature_foo:
  default_only:
  default_value: 10
  /PLATFORM/:
    default_only: false
    [snip]
```

Sounds reasonable to me.

chrisvanheuveln: Yes, I like the implied true better since it still leaves us open for an unlikely false case. 

I'm okay with the original code too - I was mainly just bouncing the idea off you.

glennmatthews: Another crazy idea I just had - roll the default value itself into the default_only entry. So instead of:

``` yaml
feature_foo:
  default_value: 'x'
  default_only:
```

we could just do:

``` yaml
feature_foo:
  default_only: 'x'
```

Any thoughts on this?

(EDIT) This would make `default_value` and `default_only` mutually exclusive at any given level - so Rob's above special-case exception would be written as:

``` yaml
feature_foo:
  default_only: 10
  /PLATFORM/:
    default_value: 10
    config_get: ...
```

Is this cool, or too clever for its own good?

robert-w-gries: I prefer that idea as a solution, given that templates force the creation of a new keyword.

glennmatthews:
New diffs are up, with the discussed changes including now using default_only to set the default value.

glennmatthews:
Figured out how to make Minitest automatically convert an UnsupportedError into a 'Skip' result. :-)

chrisvanheuveln:
:+1: :+1:

PR #74, opened 2015-11-10
Minitest fixes

chrisvanheuveln:
Tests in diff now run clean on n30k,n31k,n9k

### DIFF
--- a/tests/test_interface.rb
+++ b/tests/test_interface.rb
@@ -81,7 +81,7 @@ class TestInterface < CiscoTestCase
                          'Changing interface speed is not permitted)')
     skip('Skip test: Interface type does not allow config change') if
          message[pattern]
-    fail
+    flunk(message)
   end
 
   def create_interface(ifname=interfaces[0])
@@ -177,7 +177,7 @@ class TestInterface < CiscoTestCase
   rescue RuntimeError => e
     skip('NX-OS defect: system default switchport nvgens twice') if
       e.message[/Expected zero.one value/]
-    raise
+    flunk(e.message)
   end
 
   def validate_description(inttype_h)
--- a/tests/test_interface.rb
+++ b/tests/test_interface.rb
@@ -75,6 +75,15 @@ class TestInterface < CiscoTestCase
     arr.count
   end
 
+  # Helper to check for misc speed change disallowed error messages.
+  def speed_change_disallowed(message)
+    pattern = Regexp.new('(port doesn t support this speed|' \
+                         'Changing interface speed is not permitted)')
+    skip('Skip test: Interface type does not allow config change') if
+         message[pattern]
+    fail
+  end
+
   def create_interface(ifname=interfaces[0])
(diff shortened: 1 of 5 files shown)


tests/test_interface.rb
```diff
@@ -75,6 +75,15 @@ def interface_count
     arr.count
   end
 
+  # Helper to check for misc speed change disallowed error messages.
+  def speed_change_disallowed(message)
+    pattern = Regexp.new('(port doesn t support this speed|' \
+                         'Changing interface speed is not permitted)')
+    skip('Skip test: Interface type does not allow config change') if
+         message[pattern]
+    fail
```
glennmatthews: better: `flunk(message)`

tests/test_snmpserver.rb
```diff
@@ -312,9 +312,9 @@ def test_snmpserver_global_enforce_priv_get_enabled
     snmpserver.global_enforce_priv = true
     device_enabled = snmpserver.global_enforce_priv?
     if device_enabled
-      refute_show_match(pattern: /no snmp-server globalEnforcePriv/)
+      assert_show_match(pattern: /snmp-server globalEnforcePriv/)
     else
-      assert_show_match(pattern: /no snmp-server globalEnforcePriv/)
+      refute_show_match(pattern: /no snmp-server globalEnforcePriv/)
```
glennmatthews: still confused here - both branches have the same expectation (device enabled --> expect on ; device not enabled --> do not expect off)

glennmatthews: For that matter, what's with this test? We just explicitly set it to true at line 312, so why do we suddenly have doubts about what the current state is?

chrisvanheuveln: Good points. It looks like this test was a cut&paste of the test before it (a default test) which also has the dubious logic. I think the intent they were going for in the first test was to set it to default (from the yaml instead of an explicit boolean) and then be able to test the pattern in either case. This test just copied the logic. I'll just rewrite these two tests.

glennmatthews:
:+1:

tests/test_interface.rb
```diff
@@ -682,9 +712,8 @@ def test_negotiate_auto_ethernet
     # Some platforms/interfaces/versions do not support negotiation changes
     begin
       interface.negotiate_auto = false
-    rescue => e
-      skip('Skip test: Interface type does not allow config change') if
-        e.message[/requested config change not allowed/]
+    rescue RuntimeError => e
+      speed_change_disallowed(e.message)
```
glennmatthews, line 716: @chrisvanheuveln This seems to have been an incorrect fix - speed_change_disallowed doesn't match the `/requested config change not allowed/` pattern so this test case is now failing for me.

chrisvanheuveln, line 716: Hmmm. You're right. I tested on n30,n31,n9k. Maybe you have a different linecard. Which testbed? i'll fix it.

PR #76, opened 2015-11-11
Feature/vxlan

alok-aggarwal:
YAML refactoring

### DIFF
--- a/lib/cisco_node_utils/DELETED.command_reference_common.yaml
+++ b/lib/cisco_node_utils/DELETED.command_reference_common.yaml
@@ -1054,38 +1054,38 @@ yum:
   remove:
     config_set: "install deactivate %s"
 
-vxlan_global:
-  feature:
-    config_get: "show running | i ^feature"
-    config_get_token: '/^feature fabric forwarding$/'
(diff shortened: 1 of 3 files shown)


lib/cisco_node_utils/DELETED.command_reference_common.yaml
```diff
@@ -1053,39 +1053,3 @@ yum:
     config_get_token: '/^%s\S*\s+(\S+)\s+(?:installed|@\S+)/'
   remove:
     config_set: "install deactivate %s"
```
glennmatthews, line 1055: Is there more work needed before this file can be deleted altogether?

mikewiebe:
:+1:  @alok-aggarwal I will merge once you address @glennmatthews comment.

lib/cisco_node_utils/DELETED.command_reference_common.yaml
alok-aggarwal, line 1055: I believe so. The definitions for vxlan_vtep, vxlan, pim etc exist there and they don't have their own yaml files yet. The node utils for these needs to be developed.

PR #77, opened 2015-11-12
route distinguisher to bgp and route target to bgp_af

andy91315:
1. pass minitest
   route distinguisher:
   TestRouterBgp#test_routerbgp_set_get_route_distinguisher = 11.09 s = .
   TestRouterBgp#test_routerbgp_get_route_distinguisher_not_configured = 7.66 s =
   TestRouterBgp#test_routerbgp_default_route_distinguisher = 7.38 s =

route target:
TestRouterBgpAF#test_route_target = 20.72 s = .
Finished in 20.723943s, 0.0483 runs/s, 2.1231 assertions/s.
1 runs, 44 assertions, 0 failures, 0 errors, 0 skips
1. pass static analysis


andy91315:
No idea how to fix the recursion with two returns

lib/cisco_node_utils/cisco_cmn_utils.rb
```diff
@@ -111,14 +111,24 @@ def self.process_network_mask(network)
     # Useful for network, redistribute, etc.
     #   should: an array of expected cmds (manifest/recipe)
     #  current: an array of existing cmds on the device
+    def self.depth(a)
+      return 0 unless a.is_a?(Array)
+      return 1 + depth(a[0])
```
chrisvanheuveln: Just delete the second 'return':

```
    def self.depth(a)
      return 0 unless a.is_a?(Array)
      1 + depth(a[0])
    end
```

tests/test_bgp_af.rb
```diff
@@ -959,4 +959,89 @@ def test_utils_delta_add_remove
...
+      bgp_af.send("route_target_#{opt}_evpn=", should)
+    end
+
+    opts.each do |opt|
+      result = bgp_af.send("route_target_#{opt}_evpn")
+      assert_equal(should, result,
+                   "#{test_id} : #{af} : route_target_#{opt}_evpn")
+    end
+  end
 end
```
chrisvanheuveln, line 1084: We'll need some additional tests for the delta_add_remove method since it's been modified to handle depth == 1.

Take a look at test_utils_delta_add_remove and replicate it to test depth 1, e.g.:

```
def   def test_utils_delta_add_remove_depth_1
  ...
    should = ['1:1', '2:2', '3:3', '4:4', '5:5', '6:6']

   current = []
    expected = { add: should, remove: [] }
    result = Utils.delta_add_remove(should, current)
    assert_equal(expected, result, 'Test 1. delta mismatch')
  ...
  ... more tests
```

lib/cisco_node_utils/bgp.rb
```diff
@@ -494,6 +507,35 @@ def default_reconnect_interval
...
+    # Route Distinguisher (Getter/Setter/Default)
+    # Configure in vrf context
+    def route_distinguisher
+      return false unless RouterBgp.feature_nv_overlay_evpn_enabled
+      config_get('bgp', 'route_distinguisher', @get_args)
+    end
+
+    def route_distinguisher=(rd)
+      # In order to remove a bgp route-distinguisher you cannot simply issue
+      # 'no rd'. Dummy-id specified to work around this.
```
chrisvanheuveln: Are you sure about this comment? I can delete it without an rd value:

```
# sh run | section 'vrf context red' 
vrf context red
  rd 4:4

# no rd
# sh run | section 'vrf context red' 
vrf context red
```

lib/cisco_node_utils/bgp.rb
```diff
@@ -494,6 +507,35 @@ def default_reconnect_interval
...
+      RouterBgp.feature_nv_overlay_evpn_enable if
+        rd != default_route_distinguisher &&
```
chrisvanheuveln: Why do you need to test rd here? Just do this:

```
      RouterBgpAF.feature_nv_overlay_evpn_enable if
        !RouterBgpAF.feature_nv_overlay_evpn_enabled
```

Note that I'm using RouterBgpAF here instead since this method is already defined there. Which means to use it here you will also need to add a require to the top of this file for bgp_af.rb:

```
require_relative 'bgp_af'
```

lib/cisco_node_utils/bgp.rb
```diff
@@ -494,6 +507,35 @@ def default_reconnect_interval
...
+        !RouterBgp.feature_nv_overlay_evpn_enabled
+      dummy_rd = '1:1'
+      if rd == default_route_distinguisher
+        @set_args[:state] = 'no'
+        @set_args[:rd] = dummy_rd
```
chrisvanheuveln: If you actually need to specify rd (see above comment) then don't use a dummy_rd, use the getter method:

```
   @set_args[:rd] = route_distinguisher
```

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -547,5 +547,183 @@ def redistribute=(should)
...
+      config_get_default('bgp_af', 'route_target_both_auto_evpn')
+    end
+
+    # route target export
+    def route_target_export
+      cmds = config_get('bgp_af', 'route_target_export', @get_args)
+      # config_get returns an array, so cmds == ['1:1','2:2']
+
+      # normalize to a string
+      cmds.empty? ? default_route_target_export : cmds.join(' ')
```
chrisvanheuveln: I'd rather see this as:

```
cmds.sort.join(' ') unless cmds.empty?
```

Also, remove the `# normalize to a string` comment

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -547,5 +547,183 @@ def redistribute=(should)
...
+    end
+
+    def route_target_export=(should)
+      # 'should'comes in as a string, normalize into array
```
chrisvanheuveln: `# 'should' is a string, normalize to an array`

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -547,5 +547,183 @@ def redistribute=(should)
...
+      should = should.nil? ? [] : should.split(' ')
```
chrisvanheuveln: `should = should.empty? ? [] : should.split(' ')`

same change for the `is.nil?` below

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -547,5 +547,183 @@ def redistribute=(should)
...
+
+      # 'is' is a string, normalize into array
+      is = route_target_export
+      is = is.nil? ? [] : is.split(' ')
+
+      delta_hash = Utils.delta_add_remove(should, is)
+      puts "delta hash=#{delta_hash}"
```
chrisvanheuveln: remove `puts`

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -547,5 +547,183 @@ def redistribute=(should)
...
+      return if delta_hash.values.flatten.empty?
+      [:add, :remove].each do |action|
+        CiscoLogger.debug('route_target_export delta ' \
+          "#{@get_args}\n #{action}: #{delta_hash[action]}")
+        delta_hash[action].each do |target|
+          state = (action == :add) ? '' : 'no'
+          @set_args[:state] = state
+          @set_args[:community] = target
```
chrisvanheuveln: Keep the var name the same as the hash key:

```
          @set_args[:community] = community
```

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -547,5 +547,183 @@ def redistribute=(should)
...
+        CiscoLogger.debug('route_target_export_evpn delta ' \
+          "#{@get_args}\n #{action}: #{delta_hash[action]}")
+        delta_hash[action].each do |target|
+          state = (action == :add) ? '' : 'no'
+          @set_args[:state] = state
+          @set_args[:community] = target
+
+          config_set('bgp_af', 'route_target_export_evpn', @set_args)
+        end
+      end
```
chrisvanheuveln: There's a lot of duplication here with the other route_target setters. I think we could just make a helper function instead, e.g.:

1) replace your setter code with:

```
    def route_target_export_evpn=(should)
      # 'should' is a string, normalize to an array
      should = should.empty? ? [] : should.split(' ')

      # 'is' is a string, normalize into array
      is = route_target_export_evpn
      is = is.empty? ? [] : is.split(' ')

      route_target_delta(should, is, 'route_target_export_evpn')
    end
```

2) Create helper method:

```
def route_target_delta(should, is, prop)
  delta_hash = Utils.delta_add_remove(should, is)
  return if delta_hash.values.flatten.empty?
  [:add, :remove].each do |action|
    CiscoLogger.debug(#{prop} #{@get_args}\n #{action}: #{delta_hash[action]}")
    delta_hash[action].each do |community|
      state = (action == :add) ? '' : 'no'
      @set_args[:state] = state
      @set_args[:community] = community
       config_set('bgp_af', prop, @set_args)
    end
  end
end
```

lib/cisco_node_utils/cisco_cmn_utils.rb
```diff
@@ -111,14 +111,24 @@ def self.process_network_mask(network)
     # Useful for network, redistribute, etc.
     #   should: an array of expected cmds (manifest/recipe)
     #  current: an array of existing cmds on the device
```
chrisvanheuveln, line 113: These comments should be at the top of the delta_add_remove method instead of here. Or just move the depth method to follow delta_add_remove.

lib/cisco_node_utils/cmd_ref/bgp.yaml
```diff
@@ -84,6 +84,13 @@ feature:
   config_get_token: '/^feature bgp$/'
   config_set: "<state> feature bgp"
 
+# This is a multi-provider, enable-only feature. Do not disable from provider.
+feature_nv_overlay_evpn:
```
chrisvanheuveln: We already have this defined in bgp_af.yaml so no need to duplicate it here. This is related to my comment for line 520 in bgp.rb.

tests/test_bgp_af.rb
```diff
@@ -959,4 +959,89 @@ def test_utils_delta_add_remove
...
+    refute(bgp_af.route_target_both_auto_evpn, "vrf context #{vrf} af #{af}: "\
+           'bgp_af route-target both auto evpn should be disabled')
+
+    opts = [:import, :export]
+
+    # Master list of communities to test against
+    master = '1:1 2:2 3:3 4:5'
+
+    # Test 1: both/import/export when no commands are present. Each target
+    # option will be tested with and without evpn (6 separate types)
```
chrisvanheuveln, line 1046:
```
    # Test 1:import/export when no commands are present. Each target
    # option will be tested with and without evpn (4 separate types)
```

tests/test_bgp_af.rb
```diff
@@ -959,4 +959,89 @@ def test_utils_delta_add_remove
...
+    # Now check the results
+    opts.each do |opt|
+      # non-evpn
+      result = bgp_af.send("route_target_#{opt}")
+      assert_equal(should, result,
+                   "#{test_id} : #{af} : route_target_#{opt}")
+    end
+
+    opts.each do |opt|
+      bgp_af.send("route_target_#{opt}_evpn=", should)
```
chrisvanheuveln: This should be part of the earlier setter loop. Remove this loop (lines 1037-1039) and move this setter call into the loop at 1024-1027 so that it looks like this:

```
    opts.each do |opt|
      # non-evpn
      bgp_af.send("route_target_#{opt}=", should)
      # evpn
      bgp_af.send("route_target_#{opt}_evpn=", should)
    end
```

tests/test_bgp_af.rb
```diff
@@ -959,4 +959,89 @@ def test_utils_delta_add_remove
...
+    end
+
+    opts.each do |opt|
+      result = bgp_af.send("route_target_#{opt}_evpn")
```
chrisvanheuveln, line 1079: likewise, remove this loop and add the getter/asserts to the above loop at 1030-1035:

```
    # Now check the results
    opts.each do |opt|
      # non-evpn
      result = bgp_af.send("route_target_#{opt}")
      assert_equal(should, result,
                   "#{test_id} : #{af} : route_target_#{opt}")

      # evpn
      result = bgp_af.send("route_target_#{opt}_evpn")
      assert_equal(should, result,
                   "#{test_id} : #{af} : route_target_#{opt}_evpn")
    end

```

tests/test_router_bgp.rb
```diff
@@ -624,6 +624,40 @@ def test_routerbgp_get_reconnect_interval_default
     bgp.destroy
   end
 
+  def test_routerbgp_set_get_route_distinguisher
+    asnum = 55
+    vrf = 'blue'
+    bgp = RouterBgp.new(asnum, vrf)
```
chrisvanheuveln: You're not using asnum or vrf anywhere else in this method so just get rid of the variables and put the values explicitly in the call:

`bgp = RouterBgp.new(55, 'blue')`

tests/test_router_bgp.rb
```diff
@@ -624,6 +624,40 @@ def test_routerbgp_get_reconnect_interval_default
...
+    bgp.route_distinguisher = '1:1'
+    assert_equal('1:1', bgp.route_distinguisher,
+                 "bgp route_distinguisher should be set to '1:1'")
+    bgp.route_distinguisher = ''
+    assert_empty(bgp.route_distinguisher,
+                 'bgp route_distinguisher should *NOT* be configured')
+    bgp.destroy
+  end
+
+  def test_routerbgp_get_route_distinguisher_not_configured
```
chrisvanheuveln: Delete this test method and the next one since you're already doing this test at line 637. 
Change line 637 to be this instead:

`bgp.route_distinguisher = bgp.default_route_distinguisher`

lib/cisco_node_utils/bgp.rb
```diff
@@ -154,6 +154,19 @@ def set_args_keys_default
       end
     end
 
+    # This is an enable-only method for enabling the 'nv overlay evpn' feature
```
chrisvanheuveln: You don't need these feature methods. Please remove them. See comment on line 520.

lib/cisco_node_utils/bgp.rb
```diff
@@ -494,6 +495,32 @@ def default_reconnect_interval
...
 
+    # Route Distinguisher (Getter/Setter/Default)
+    # Configure in vrf context
+    def route_distinguisher
+      return false unless RouterBgpAF.feature_nv_overlay_evpn_enabled
+      config_get('bgp', 'route_distinguisher', @get_args)
+    end
+
+    def route_distinguisher=(rd)
+      # In order to remove a bgp route-distinguisher you cannot simply issue
```
chrisvanheuveln: remove comment

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -547,5 +547,109 @@ def redistribute=(should)
...
+    end
+
+    def default_route_target_both_auto_evpn
+      config_get_default('bgp_af', 'route_target_both_auto_evpn')
+    end
+
+    # route target export
+    def route_target_export
+      cmds = config_get('bgp_af', 'route_target_export', @get_args)
+      cmds.empty? ? [] : cmds.sort
```
chrisvanheuveln: Glenn pointed out that these can be simplified even more to just cmds.sort since you'll get [] from that as well:

```
-      cmds.empty? ? [] : cmds.sort
+      cmds.sort
```

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -547,5 +547,109 @@ def redistribute=(should)
...
+      route_target_delta(should, route_target_import_evpn,
+                         'route_target_import_evpn')
+    end
+
+    def default_route_target_import_evpn
+      config_get_default('bgp_af', 'route_target_import_evpn')
+    end
+
+    def route_target_delta(should, is, prop)
+      puts "delta #{should} is #{is}"
```
chrisvanheuveln: Remove the puts line

chrisvanheuveln: Mostly just nits on this last round of diffs. Address those and I think we're ready to merge.

andy91315:
make the fix

PR #78, opened 2015-11-12
Feature/grpc confed

RichWellum:
Note: there is an associated puppet change that is needed to run this manifest and that's going to be a separate review.

Changes to make bgp confederation peers work on XR.

Change the setter to use the delta functionality introduced recently.

Mini test on both platforms:

Xrv9k:

```
Finished in 478.095544s, 0.1129 runs/s, 0.2886 assertions/s.
54 runs, 138 assertions, 0 failures, 0 errors, 24 skips
```

Nexus:

```
Finished in 595.645666s, 0.0907 runs/s, 0.3693 assertions/s.
54 runs, 220 assertions, 0 failures, 0 errors, 0 skips
```

Adding confed peers to my bgp manifest:

root@rtp-puppetserver228:~/cisco-network-puppet-module# cat /etc/puppetlabs/code/environments/production/manifests/site.pp

``` puppet
class ciscopuppet::demo_bgp {
  cisco_bgp { 'default_vrf_global':
    ensure                                 => present,
    asn                                    => 55.77,
    vrf                                    => 'default',
    router_id                              => '1.1.1.1',

    # Graceful Restart Properties
    graceful_restart                       => true,

    # Inherited Properties - applied to VRF but configured globally
    confederation_id                       => '33',
    confederation_peers                    => '999',
    cluster_id                             => '55',
    graceful_restart_timers_restart        => '131',
    graceful_restart_timers_stalepath_time => '311',
    bestpath_med_confed                    => true,
  } 
}
```

Results in:

```
Info: Loading facts
Info: Caching catalog for sunstoneabc.cisco.com
Info: Applying configuration version '1447352626'
Notice: Applied catalog in 2.18 seconds

RP/0/RP0/CPU0:ios#sh run router bgp
Thu Nov 12 18:35:16.666 UTC
router bgp 3604557
 bgp confederation peers
  999
 !
 bgp confederation identifier 33
 bgp router-id 1.1.1.1
 bgp cluster-id 55
 bgp graceful-restart restart-time 131
 bgp graceful-restart stalepath-time 311
 bgp graceful-restart
 bgp bestpath med confed
 vrf blue
  rd auto
  timers bgp 46 111
  bgp router-id 192.168.0.66
  bgp bestpath cost-community ignore
  bgp bestpath med always
  bgp bestpath compare-routerid
  bgp bestpath as-path multipath-relax
 !
!
```

Changing the peer to a list and expecting a CLI error:

``` puppet
    confederation_peers                    => '999 44 55.77 22 1001 1003 88',
```

```
Notice: /Stage[main]/Ciscopuppet::Demo_bgp/Cisco_bgp[default_vrf_global]/confederation_peers: confederation_peers changed '999' to '999 44 55.77 22 1001 1003 88'
Error: /Stage[main]/Ciscopuppet::Demo_bgp/Cisco_bgp[default_vrf_global]: Could not evaluate: CliError: 'input TODO' rejected with message:
'["'BGP' detected the 'warning' condition 'Local member-AS not allowed in confed peer list'"]'
Notice: Applied catalog in 3.93 seconds
```

Change it to a working range of peers:

``` puppet
    confederation_peers                    => '999 44 55.77 22 1001 1003 88',
```

```
Notice: /Stage[main]/Ciscopuppet::Demo_bgp/Cisco_bgp[default_vrf_global]/confederation_peers: confederation_peers changed '1001 1003 22 44 88 999' to '999 44 22 1001 1003 88'
```

```
RP/0/RP0/CPU0:ios#sh run router bgp 3604557 bgp confederation peers
Thu Nov 12 19:15:18.091 UTC
router bgp 3604557
 bgp confederation peers
  22
  44
  88
  999
  1001
  1003
 !
!
```

### DIFF
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+require_relative 'cisco_cmn_utils'
 require_relative 'node_util'
 
 module Cisco
@@ -338,23 +339,38 @@ module Cisco
 
     # Confederation Peers (Getter/Setter/Default)
     def confederation_peers
-      config_get('bgp', 'confederation_peers', @get_args)
+      cmds = config_get('bgp', 'confederation_peers', @get_args)
+      # For XR convert array of peers to a string
+      cmds.to_s.empty? ? default_confederation_peers : cmds.join(' ')
     end
 
-    def confederation_peers_set(peers)
-      # The confederation peers command is additive so we first need to
-      # remove any existing peers.
-      unless confederation_peers.empty?
-        @set_args[:state] = 'no'
-        @set_args[:peer_list] = confederation_peers
-        config_set('bgp', 'confederation_peers', @set_args)
+    def confederation_peers=(should)
+      # confederation peers are additive. So create a delta hash of entries
+      # and only apply the changes
+      if should.instance_of? String
+        should = should.nil? ? [] : should.split(' ')
+      else
+        should = Array(should)
       end
-      unless peers == default_confederation_peers
-        @set_args[:state] = ''
-        @set_args[:peer_list] = peers
-        config_set('bgp', 'confederation_peers', @set_args)
+      is = confederation_peers
+      is = is.nil? ? [] : is.split(' ')
+
+      puts "is: #{is.class}, should: #{should.class}"
+      delta_hash = Utils.delta_add_remove(should, is)
+      puts "delta hash=#{delta_hash}"
+      return if delta_hash.values.flatten.empty?
+      [:add, :remove].each do |action|
+        CiscoLogger.debug("confederation_peers delta " \
+                          "#{@get_args}\n #{action}: " \
+                          "#{delta_hash[action]}")
+        delta_hash[action].each do |peer|
+          state = (action == :add) ? '' : 'no'
+          @set_args[:state] = state
+          @set_args[:peer] = peer
+
+          config_set('bgp', 'confederation_peers', @set_args)
+        end
       end
-      set_args_keys_default
     end
 
     def default_confederation_peers
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -343,10 +343,9 @@ module Cisco
       # XR: retrieved as an array.
       # So make it an array in the NX case.
       # Sort the end results to make it consistent across both platforms.
-      # And if peers is an empty list - make it an empty array.
       peers = config_get('bgp', 'confederation_peers', @get_args)
       peers = peers.split(' ') if peers.is_a?(String)
-      peers.empty? ? [] : peers.sort
+      peers.sort
     end
 
     def confederation_peers=(should)
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -62,7 +62,7 @@ module Cisco
       hash_final = {}
       asnum = asnum.to_i unless /\d+.\d+/.match(asnum)
       hash_tmp = { asnum =>
-        { 'default' => RouterBgp.new(asnum, 'default', false) } }
+                   { 'default' => RouterBgp.new(asnum, 'default', false) } }
       vrf_ids = config_get('bgp', 'vrf', asnum: asnum)
       unless vrf_ids.nil?
         vrf_ids.each do |vrf|
@@ -358,7 +358,7 @@ module Cisco
       delta_hash = Utils.delta_add_remove(should, is)
       return if delta_hash.values.flatten.empty?
       [:add, :remove].each do |action|
-        CiscoLogger.debug("confederation_peers delta " \
+        CiscoLogger.debug('confederation_peers delta ' \
                           "#{@get_args}\n #{action}: " \
                           "#{delta_hash[action]}")
         delta_hash[action].each do |peer|
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -61,8 +61,9 @@ module Cisco
 
       hash_final = {}
       asnum = asnum.to_i unless /\d+.\d+/.match(asnum)
-      hash_tmp = { asnum =>
-                   { 'default' => RouterBgp.new(asnum, 'default', false) } }
+      hash_tmp = {
+        asnum => { 'default' => RouterBgp.new(asnum, 'default', false) }
+      }
       vrf_ids = config_get('bgp', 'vrf', asnum: asnum)
       unless vrf_ids.nil?
         vrf_ids.each do |vrf|
@@ -340,8 +341,9 @@ module Cisco
     # Confederation Peers (Getter/Setter/Default)
     def confederation_peers
       cmds = config_get('bgp', 'confederation_peers', @get_args)
-      # For XR convert array of peers to a string
-      cmds.to_s.empty? ? default_confederation_peers : cmds.join(' ')
+      return default_confederation_peers if cmds.to_s.empty?
+      cmds = cmds.split(' ') if cmds.instance_of? String
+      cmds.sort.join(' ')
     end
 
     def confederation_peers=(should)
@@ -349,9 +351,9 @@ module Cisco
       # and only apply the changes
       # Convert strings (1 2 3) to an array
       # Or FixedNum (1) to a string then an array
-      should = should.to_s.nil? ? [] : should.to_s.split(' ')
+      should = should.to_s.empty? ? [] : should.to_s.split(' ')
       is = confederation_peers
-      is = is.nil? ? [] : is.split(' ')
+      is = is.empty? ? [] : is.split(' ')
 
       delta_hash = Utils.delta_add_remove(should, is)
       return if delta_hash.values.flatten.empty?
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -61,9 +61,8 @@ module Cisco
 
       hash_final = {}
       asnum = asnum.to_i unless /\d+.\d+/.match(asnum)
-      hash_tmp = {
-        asnum => { 'default' => RouterBgp.new(asnum, 'default', false) }
-      }
+      hash_tmp = { asnum =>
+        { 'default' => RouterBgp.new(asnum, 'default', false) } }
       vrf_ids = config_get('bgp', 'vrf', asnum: asnum)
       unless vrf_ids.nil?
         vrf_ids.each do |vrf|
@@ -340,20 +339,20 @@ module Cisco
 
     # Confederation Peers (Getter/Setter/Default)
     def confederation_peers
-      cmds = config_get('bgp', 'confederation_peers', @get_args)
-      return default_confederation_peers if cmds.to_s.empty?
-      cmds = cmds.split(' ') if cmds.instance_of? String
-      cmds.sort.join(' ')
+      # NX: confederation peers is retrieved as a string '1 2 3'
+      # XR: retrieved as an array.
+      # So make it an array in the NX case.
+      # Sort the end results to make it consistent across both platforms.
+      # And if peers is an empty list - make it an empty array.
+      peers = config_get('bgp', 'confederation_peers', @get_args)
+      peers = peers.split(' ') if peers.is_a?(String)
+      peers.empty? ? [] : peers.sort
     end
 
     def confederation_peers=(should)
       # confederation peers are additive. So create a delta hash of entries
       # and only apply the changes
-      # Convert strings (1 2 3) to an array
-      # Or FixedNum (1) to a string then an array
-      should = should.to_s.empty? ? [] : should.to_s.split(' ')
       is = confederation_peers
-      is = is.empty? ? [] : is.split(' ')
 
       delta_hash = Utils.delta_add_remove(should, is)
       return if delta_hash.values.flatten.empty?
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -355,9 +355,7 @@ module Cisco
       is = confederation_peers
       is = is.nil? ? [] : is.split(' ')
 
-      puts "is: #{is.class}, should: #{should.class}"
       delta_hash = Utils.delta_add_remove(should, is)
-      puts "delta hash=#{delta_hash}"
       return if delta_hash.values.flatten.empty?
       [:add, :remove].each do |action|
         CiscoLogger.debug("confederation_peers delta " \
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -347,11 +347,9 @@ module Cisco
     def confederation_peers=(should)
       # confederation peers are additive. So create a delta hash of entries
       # and only apply the changes
-      if should.instance_of? String
-        should = should.nil? ? [] : should.split(' ')
-      else
-        should = Array(should)
-      end
+      # Convert strings (1 2 3) to an array
+      # Or FixedNum (1) to a string then an array
+      should = should.to_s.nil? ? [] : should.to_s.split(' ')
       is = confederation_peers
       is = is.nil? ? [] : is.split(' ')
 
--- a/lib/cisco_node_utils/cisco_cmn_utils.rb
+++ b/lib/cisco_node_utils/cisco_cmn_utils.rb
@@ -121,8 +121,8 @@ module Cisco
       should.each(&:compact!) unless should.empty? if depth(should) > 1
       delta = { add: should - current, remove: current - should }
 
-      # To allow this method to handle comparing nested and unnested arrays
-      # the depth checks differentiate between the two cases.
+      # Differentiate between comparing nested and unnested arrays by
+      # checking the depth of the array.
       if depth(should) == 1
         # Delete entries from :remove if f1 is an update to an existing command
         delta[:add].each do |id|
--- a/lib/cisco_node_utils/cisco_cmn_utils.rb
+++ b/lib/cisco_node_utils/cisco_cmn_utils.rb
@@ -121,6 +121,8 @@ module Cisco
       should.each(&:compact!) unless should.empty? if depth(should) > 1
       delta = { add: should - current, remove: current - should }
 
+      # To allow this method to handle comparing nested and unnested arrays
+      # the depth checks differentiate between the two cases.
       if depth(should) == 1
         # Delete entries from :remove if f1 is an update to an existing command
         delta[:add].each do |id|
--- a/lib/cisco_node_utils/cisco_cmn_utils.rb
+++ b/lib/cisco_node_utils/cisco_cmn_utils.rb
@@ -107,18 +107,32 @@ module Cisco
       network
     end
 
+    def self.depth (a)
+      return 0 unless a.is_a?(Array)
+      return 1 + depth(a[0])
+    end
+
     # Helper to build a hash of add/remove commands for a nested array.
     # Useful for network, redistribute, etc.
     #   should: an array of expected cmds (manifest/recipe)
     #  current: an array of existing cmds on the device
     def self.delta_add_remove(should, current=[])
       # Remove nil entries from array
-      should.each(&:compact!) unless should.empty?
+      if depth(should) > 1
+        should.each(&:compact!) unless should.empty?
+      end
       delta = { add: should - current, remove: current - should }
 
-      # Delete entries from :remove if f1 is an update to an existing command
-      delta[:add].each do |id, _|
-        delta[:remove].delete_if { |f1, f2| [f1, f2] if f1.to_s == id.to_s }
+      if depth(should) == 1
+        # Delete entries from :remove if f1 is an update to an existing command
+        delta[:add].each do |id, _| #todo remove _
+          delta[:remove].delete_if { |f1| [f1] if f1.to_s == id.to_s }
+        end
+      else
+        # Delete entries from :remove if f1 is an update to an existing command
+        delta[:add].each do |id, _|
+          delta[:remove].delete_if { |f1, f2| [f1, f2] if f1.to_s == id.to_s }
+        end
       end
       delta
     end # delta_add_remove
--- a/lib/cisco_node_utils/cisco_cmn_utils.rb
+++ b/lib/cisco_node_utils/cisco_cmn_utils.rb
@@ -125,7 +125,7 @@ module Cisco
 
       if depth(should) == 1
         # Delete entries from :remove if f1 is an update to an existing command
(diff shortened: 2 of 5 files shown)


lib/cisco_node_utils/bgp.rb
```diff
@@ -338,23 +340,35 @@ def default_enforce_first_as
...
-        config_set('bgp', 'confederation_peers', @set_args)
-      end
-      unless peers == default_confederation_peers
-        @set_args[:state] = ''
-        @set_args[:peer_list] = peers
-        config_set('bgp', 'confederation_peers', @set_args)
+      cmds = config_get('bgp', 'confederation_peers', @get_args)
+      return default_confederation_peers if cmds.to_s.empty?
+      cmds = cmds.split(' ') if cmds.instance_of? String
+      cmds.sort.join(' ')
```
chrisvanheuveln: Some comments would be helpful here since it's not obvious why you're splitting on one line and joining on the next.

Also you could combine those two lines into:
```
cmds.is_a?(String) ? cmds.split(' ') : cmds.sort.join(' ')
```

lib/cisco_node_utils/bgp.rb
```diff
@@ -338,23 +340,35 @@ def default_enforce_first_as
...
-        @set_args[:peer_list] = confederation_peers
-        config_set('bgp', 'confederation_peers', @set_args)
-      end
-      unless peers == default_confederation_peers
-        @set_args[:state] = ''
-        @set_args[:peer_list] = peers
-        config_set('bgp', 'confederation_peers', @set_args)
+      cmds = config_get('bgp', 'confederation_peers', @get_args)
+      return default_confederation_peers if cmds.to_s.empty?
+      cmds = cmds.split(' ') if cmds.instance_of? String
```
robert-w-gries: @glennmatthews Wondering if we should update the YAML entry for this property after your enhancements.  Also, should we remove `return default_confederation_peers if cmds.to_s.empty?` since config_get should do that by default?

Edit: moved yaml comment to the appropiate location

lib/cisco_node_utils/cmd_ref/bgp.yaml
```diff
@@ -109,14 +109,14 @@ confederation_id:
     config_set_append: '<state> confederation identifier <id>'
 
 confederation_peers:
-  # TODO: Supported in XR but broken as identifier is nvgened on a seperate line.
   default_value: ""
```
robert-w-gries: @glennmatthews Thoughts on adding `kind: string`?  Also, I'm not sure whether we need `auto_default: false` or not

lib/cisco_node_utils/node.rb
```diff
@@ -82,8 +82,8 @@ def config_get(feature, name, *args)
         # Just get the whole output
         return massage(show(ref.config_get, :structured), ref)
       elsif token[0].kind_of?(Regexp)
-        return massage(find_ascii(show(ref.config_get, :ascii),
-                                  token[-1], *token[0..-2]), ref)
+        return massage(Cisco.find_ascii(show(ref.config_get, :ascii),
+                                        token[-1], *token[0..-2]), ref)
```
robert-w-gries, line 86: Context:  This code was originally added in [this commit](https://github.com/cisco/cisco-network-node-utils/commit/dc9e1141d1cb2ec9974adc7625c02da46db7134e).  It was removed in [this commit](https://github.com/cisco/cisco-network-node-utils/commit/b31d506c84068f716b17c1c17fc81ec4f82bd12a#diff-73ac5adb236895c412f98beb3210f365L103).

lib/cisco_node_utils/bgp.rb
```diff
@@ -60,8 +61,9 @@ def self.routers
 
       hash_final = {}
       asnum = asnum.to_i unless /\d+.\d+/.match(asnum)
-      hash_tmp = { asnum =>
-        { 'default' => RouterBgp.new(asnum, 'default', false) } }
+      hash_tmp = {
+        asnum => { 'default' => RouterBgp.new(asnum, 'default', false) }
+      }
```
glennmatthews: This is a rubocop 0.34 vs 0.35 change right? Might want to hold off on this since develop is still using 0.34.2

lib/cisco_node_utils/cisco_cmn_utils.rb
```diff
@@ -107,18 +107,30 @@ def self.process_network_mask(network)
...
+    end
+
     # Helper to build a hash of add/remove commands for a nested array.
     # Useful for network, redistribute, etc.
     #   should: an array of expected cmds (manifest/recipe)
     #  current: an array of existing cmds on the device
     def self.delta_add_remove(should, current=[])
       # Remove nil entries from array
-      should.each(&:compact!) unless should.empty?
+      should.each(&:compact!) unless should.empty? if depth(should) > 1
```
glennmatthews, line 121: Please add more comments - it's not at all obvious to me why we're checking the 'depth' of an array.

lib/cisco_node_utils/cmd_ref/bgp.yaml
glennmatthews: The issue as I understand it is that in XR this is:

```
bgp confederation peers
  1
  2
  3
```

while in NX it is:

```
bgp confederation peers 1 2 3
```

(hint hint it sure would be nice to document this somewhere).

So for NX config_get() returns a single value '1 2 3' but for XR it returns the array ['1', '2', '3'](hence the weird logic in the node_utils class).

~~Assuming I'm correct in my understanding, I would think that default_value should be `""` for NX but `[]` for XR. `kind` can be `string` for NX but not for XR (maybe?)~~

Actually it'd be better to treat this like the array it really represents, so `default_value: []` (for both NX and XR).

lib/cisco_node_utils/bgp.rb
RichWellum: Ack.

lib/cisco_node_utils/bgp.rb
glennmatthews: Instead of returning a space-separated string, how about changing this API to treat the peer-ids as what they really are (an array of ASNs) - i.e.,

```
peers = config_get('bgp', 'confederation_peers', @get_args)
peers = peers.split(' ') if peers.is_a?(String)
peers
```

lib/cisco_node_utils/bgp.rb
glennmatthews: Yeah, if you set the default_value properly it will be automatically returned by config_get - see below for my suggested approach here.

tests/test_router_bgp.rb
```diff
@@ -525,13 +523,13 @@ def test_routerbgp_default_confederation_id
...
 
   def test_routerbgp_set_get_confederation_peers
-    skip(XR_SUPPORTED_BROKEN) if platform == :ios_xr
     %w(test_default test_vrf).each do |t|
       if t == 'test_default'
         asnum = 55
         vrf = 'default'
         bgp = RouterBgp.new(asnum)
       else
+        next if platform == :ios_xr
```
glennmatthews: A possible alternative to get more detailed reporting:

``` diff
- def test_routerbgp_set_get_confederation_peers
-   %w(test_default test_vrf).each do |t|
-        if t == 'test_default'
+ def test_confederation_peers_default
     asnum = 55
     vrf = 'default'
     bgp = RouterBgp.new(asnum)
+    confed_peers_test(bgp)
+ end
+
-        else
+ def test_confederation_peers_vrf
+   skip(XR_NO_VRF_SUPPORT) if platform == :ios_xr
     asnum = 99
     vrf = 'yamllll'
     bgp = create_bgp_vrf(asnum, vrf)
+    confed_peers_test(bgp)
end
+
+ def confed_peers_test(bgp)
       # Confederation peer configuration requires that a
       # confederation id be configured first so the expectation
       # in the next test is an empty peer list
       bgp.confederation_id = 55
```

lib/cisco_node_utils/bgp.rb
RichWellum: Ack to that.

tests/test_router_bgp.rb
```diff
@@ -540,31 +538,32 @@ def test_routerbgp_set_get_confederation_peers
...
                    "vrf #{vrf}: bgp confederation_peers list should be '16'")
-      bgp.confederation_peers_set(55.77)
+      bgp.confederation_peers = (55.77)
       assert_equal('55.77', bgp.confederation_peers,
-                   "vrf #{vrf}: bgp confederation_peers list should be" \
+                   "vrf #{vrf}: bgp confederation_peers list should be " \
                    "'55.77'")
-      bgp.confederation_peers_set('15 16 55.77 18 555 299')
-      assert_equal('15 16 55.77 18 555 299',
+      bgp.confederation_peers = ('15 16 55.77 18 555 299')
```
glennmatthews: if you go by my above suggestion about changing this parameter to a proper array, this would be:
`bgp.confederation_peers = ['15', '16', '55.77', '18', '555', '299']`

lib/cisco_node_utils/cmd_ref/bgp.yaml
robert-w-gries: This looks good to me:

``` diff
 confederation_peers:
-  default_value: ""
   cli_ios_xr:
+    default_value: []
     multiple: true
     config_get_token_append: ['/^bgp confederation peers$/', '/(\S+)/']
     config_set_append: ' <state> bgp confederation peers <peer>'
   cli_nexus:
#+    kind: string
#+    default_value: ""
+    default_value: []
     config_get_token_append: '/^confederation peers (.*)$/'
     config_set_append: '<state> confederation peers <peer>'
```

Edit:  Updated to address Glenn's new suggestion

lib/cisco_node_utils/bgp.rb
RichWellum: I like this but it means you set with a String and receive an array:

  1) Failure:
TestRouterBgp#test_set_get_confederation_peers [test_router_bgp.rb:545]:
vrf default: bgp confederation_peers list should be '15'.
Expected: "15"
  Actual: ["15"]

Is this ok (I can change the minitest to be happy with this).

lib/cisco_node_utils/cisco_cmn_utils.rb
RichWellum, line 121: This was added by Chris originally to handle the difference between two nested arrays of data. I needed it to handle two arrays but not nested... I'll add a comment.

RichWellum, line 121: Added.

lib/cisco_node_utils/cmd_ref/bgp.yaml
RichWellum: Got it. Note  Rob - I can move the default to the common area of the yaml block instead of doing it 2x.

confederation_id:
    default_value: []
  cli_ios_xr:
    # Note: Only applies to global BGP; not VRF applicable
    config_get_token_append: '/^bgp confederation identifier (\d+|\d+.\d+)$/'
    config_set_append: ' <state> bgp confederation identifier <id>'
  cli_nexus:
    config_get_token_append: '/^confederation identifier (\d+|\d+.\d+)$/'
    config_set_append: '<state> confederation identifier <id>'

robert-w-gries: Good call!

tests/test_router_bgp.rb
RichWellum: But what about: bgp.confederation_peers_set(15) should this be: bgp.confederation_peers_set([15])?

And if we are moving to setting with an array - doesn't that affect the manifest? So instead of:

confed_peers => '1 2 3'

now:

confed_peers => "['1', '2', '3']"

Sort of the whole point was: input=string, convert to array so it can be sorted etc, output=string

lib/cisco_node_utils/cmd_ref/bgp.yaml
RichWellum: Not sure about the auto_default call?

robert-w-gries: From Glenn's readme update:

> Normally, if config_get_token produces no match, config_get() will return the defined default_value for this attribute. For some attributes, this may not be desirable. Setting auto_default: false will force config_get() to return nil in the non-matching case instead.

In this case, you want to config_get to return `default_confederation_peers` automatically, so you do not need to specify `auto_default: false`.

lib/cisco_node_utils/bgp.rb
RichWellum: Chatted to Glenn and we feel that the peers should be represented by an array - both for NX and XR. This is also consistent with the YANG model. So there'll be more puppet changes.

lib/cisco_node_utils/cmd_ref/bgp.yaml
RichWellum: Thanks Rob.

tests/test_router_bgp.rb
RichWellum: Ack done.

There's a lot of these - will refactor.

lib/cisco_node_utils/bgp.rb
```diff
@@ -338,23 +339,35 @@ def default_enforce_first_as
...
+      peers = peers.split(' ') if peers.is_a?(String)
+      peers.empty? ? [] : peers.sort
+    end
+
+    def confederation_peers=(should)
+      # confederation peers are additive. So create a delta hash of entries
+      # and only apply the changes
+      is = confederation_peers
+
+      delta_hash = Utils.delta_add_remove(should, is)
```
chrisvanheuveln, line 356: no need for 'is' -- just use confederation_peers directly here

lib/cisco_node_utils/cisco_cmn_utils.rb
```diff
@@ -107,18 +107,32 @@ def self.process_network_mask(network)
...
       delta = { add: should - current, remove: current - should }
 
-      # Delete entries from :remove if f1 is an update to an existing command
-      delta[:add].each do |id, _|
-        delta[:remove].delete_if { |f1, f2| [f1, f2] if f1.to_s == id.to_s }
+      # To allow this method to handle comparing nested and unnested arrays
+      # the depth checks differentiate between the two cases.
```
chrisvanheuveln: Sounds like Yoda - but I get it :-)

tests/test_router_bgp.rb
```diff
@@ -44,6 +43,18 @@ def create_bgp_vrf(asnum, vrf)
   bgp
 end
 
+def setup_default
+  asnum = 55
+  @vrf = 'default'
+  RouterBgp.new(asnum)
```
chrisvanheuveln: Nit: Might as well just use 55 here instead of the variable.

tests/test_router_bgp.rb
```diff
@@ -223,106 +234,100 @@ def test_routerbgp_destroy
...
+           'bgp bestpath_aspath_multipath_relax should be enabled')
+    bgp.bestpath_compare_routerid = true
+    assert(bgp.bestpath_compare_routerid, "vrf #{@vrf}: "\
+           'bgp bestpath_compare_routerid should be enabled')
+    bgp.bestpath_cost_community_ignore = true
+    assert(bgp.bestpath_cost_community_ignore, "vrf #{@vrf}: "\
+           'bgp bestpath_cost_community_ignore should be enabled')
+    bgp.bestpath_med_confed = true
+    if platform == :nexus ||
+       (platform == :ios_xr && @vrf == 'default')
```
chrisvanheuveln: Nit: Is the :nexus check needed? Could just do:

```
unless platform == :ios_xr && !@vrf[/default/]
```

But you can keep it as is if you prefer.

tests/test_router_bgp.rb
```diff
@@ -223,106 +234,100 @@ def test_routerbgp_destroy
...
+      # TODO: This property only works on IOS XR at the global level.
+      assert(bgp.bestpath_med_confed, "vrf #{@vrf}: "\
+             'bgp bestpath_med_confed should be enabled')
     end
+    bgp.bestpath_med_missing_as_worst = true
+    assert(bgp.bestpath_med_missing_as_worst, "vrf #{@vrf}: "\
+           'bgp bestpath_med_missing_as_worst should be enabled')
+    if platform == :nexus
+      # TODO: only applies to :nexus
```
chrisvanheuveln: still need this comment? Just wondering what it refers to? You one of these after every :nexus check?

lib/cisco_node_utils/bgp.rb
RichWellum, line 356: Will keep as is, because the 'should, is' is becoming standard nomenclature.

lib/cisco_node_utils/cisco_cmn_utils.rb
RichWellum:
```
  # Differentiate between comparing nested and unnested arrays by
  # checking the depth of the array.
```

tests/test_router_bgp.rb

tests/test_router_bgp.rb
RichWellum: Removed all of them.

lib/cisco_node_utils/bgp.rb
```diff
@@ -338,23 +339,35 @@ def default_enforce_first_as
...
-        @set_args[:peer_list] = peers
-        config_set('bgp', 'confederation_peers', @set_args)
+      # NX: confederation peers is retrieved as a string '1 2 3'
+      # XR: retrieved as an array.
+      # So make it an array in the NX case.
+      # Sort the end results to make it consistent across both platforms.
+      # And if peers is an empty list - make it an empty array.
+      peers = config_get('bgp', 'confederation_peers', @get_args)
+      peers = peers.split(' ') if peers.is_a?(String)
+      peers.empty? ? [] : peers.sort
```
glennmatthews: Is the empty check needed, or does `peers.sort` work just fine even if the array is empty?

lib/cisco_node_utils/bgp.rb
```diff
@@ -338,23 +339,35 @@ def default_enforce_first_as
...
-      end
-      unless peers == default_confederation_peers
-        @set_args[:state] = ''
-        @set_args[:peer_list] = peers
-        config_set('bgp', 'confederation_peers', @set_args)
+      # NX: confederation peers is retrieved as a string '1 2 3'
+      # XR: retrieved as an array.
+      # So make it an array in the NX case.
+      # Sort the end results to make it consistent across both platforms.
+      # And if peers is an empty list - make it an empty array.
```
glennmatthews: A list is an array?

lib/cisco_node_utils/bgp.rb
RichWellum: Both code and comment removed as not needed.

tests/test_router_bgp.rb
```diff
@@ -223,106 +234,100 @@ def test_routerbgp_destroy
...
+    assert(bgp.bestpath_always_compare_med, "vrf #{@vrf}: "\
+           'bgp bestpath_always_compare_med should be enabled')
+    bgp.bestpath_aspath_multipath_relax = true
+    assert(bgp.bestpath_aspath_multipath_relax, "vrf #{@vrf}: "\
+           'bgp bestpath_aspath_multipath_relax should be enabled')
+    bgp.bestpath_compare_routerid = true
+    assert(bgp.bestpath_compare_routerid, "vrf #{@vrf}: "\
+           'bgp bestpath_compare_routerid should be enabled')
+    bgp.bestpath_cost_community_ignore = true
+    assert(bgp.bestpath_cost_community_ignore, "vrf #{@vrf}: "\
```
glennmatthews: Could possibly remove the `"vrf #{@vrf}: "` from all of these messages - as now it will be clear from the separate test cases whether this failure is in default or in a vrf.

RichWellum: I like that - then I don't need to make the vrf a global either.

tests/test_router_bgp.rb
```diff
@@ -223,106 +232,95 @@ def test_routerbgp_destroy
...
+           'bgp bestpath_cost_community_ignore should be enabled')
+    bgp.bestpath_med_confed = true
+    unless platform == :ios_xr && !@vrf == 'default'
+      assert(bgp.bestpath_med_confed, "vrf #{@vrf}: "\
+             'bgp bestpath_med_confed should be enabled')
+    end
+    bgp.bestpath_med_missing_as_worst = true
+    assert(bgp.bestpath_med_missing_as_worst, "vrf #{@vrf}: "\
+           'bgp bestpath_med_missing_as_worst should be enabled')
+    if platform == :nexus
```
glennmatthews, line 264: If these properties are unsupported on XR, I'd like to see that tested as well, i.e.:

``` ruby
if platform == :nexus
  ...
else
  assert_raises(Cisco::UnsupportedError) { bgp.bestpath_med_non_deterministic = true }
```

RichWellum, line 264: Wow nice idea. Will do.

 else
        bgp.bestpath_med_non_deterministic = true
      end

tests/test_router_bgp.rb
```diff
@@ -338,39 +348,39 @@ def test_routerbgp_default_bestpath
     refute(bgp.default_bestpath_med_missing_as_worst,
            'default value for bestpath_med_missing_as_worst should be false')
     if platform == :nexus
-      # TODO: Only applies to :nexus
       refute(bgp.default_bestpath_med_non_deterministic,
              'default value for bestpath_med_non_deterministic should be false')
+    else
+      assert_raises(Cisco::UnsupportedError) do
+        bgp.bestpath_med_non_deterministic
```
glennmatthews: should this be `bgp.default_bestpath_med_non_deterministic`?

tests/test_router_bgp.rb
```diff
@@ -524,47 +535,45 @@ def test_routerbgp_default_confederation_id
...
+                 "bgp confederation_peers list should be ['15']")
+    bgp.confederation_peers = [16]
+    assert_equal(['16'], bgp.confederation_peers,
+                 "bgp confederation_peers list should be ['16']")
+    bgp.confederation_peers = [55.77]
+    assert_equal(['55.77'], bgp.confederation_peers,
+                 'bgp confederation_peers list should be ' \
+                 "['55.77']")
+    bgp.confederation_peers = ['15', '55.77', '16', '18', '555', '299']
+    assert_equal(['15', '16', '18', '299', '55.77', '555'],
```
glennmatthews, line 566: I wonder - is this the 'correct' sorting order, or should '55.77' come after '555' (since it's actually a 32-bit number). Don't know if this actually matters in any case, just wondering.

glennmatthews:
:+1:

tests/test_router_bgp.rb
RichWellum: Yup changed.

tests/test_router_bgp.rb
chrisvanheuveln, line 566: lexical vs numeric

Doesn't matter in this case as long as the peer sort is consistent.

chrisvanheuveln:
:+1:

PR #81, opened 2015-11-13
replace tacacs group with modified compatible version from AAA

hunsbea:
Pulling these changes off of develop instead of AAA now, since these changes can't wait for AAA to collapse. This replaces the tacacs group node utils class and test compatible with both Cisco and NetDev puppet providers. The diffs show as file changes, but it's just 2 file replacements.

### DIFF
--- a/lib/cisco_node_utils/tacacs_server_group.rb
+++ b/lib/cisco_node_utils/tacacs_server_group.rb
@@ -32,16 +32,15 @@ module Cisco
       return unless create
 
       TacacsServer.new.enable unless TacacsServer.enabled
-      config_set('aaa_server_group', 'tacacs_group', '', name)
+      config_set('tacacs_server_group', 'group', state: '', name: name)
     end
 
     def destroy
-      config_set('aaa_server_group', 'tacacs_group', 'no', @name)
+      config_set('tacacs_server_group', 'group', state: 'no', name: @name)
     end
 
     def servers
-      tacservers = config_get('aaa_server_group',
-                              'tacacs_servers', @name)
+      tacservers = config_get('tacacs_server_group', 'servers', @name)
       servs = {}
       unless tacservers.nil?
         tacservers.each { |s| servs[s] = TacacsServerHost.new(s, false) }
@@ -56,17 +55,17 @@ module Cisco
       new_servs.each do |s|
         # add any servers not yet configured
         next if current_servs.include? s
-        config_set('aaa_server_group', 'tacacs_server', @name, '', s)
+        config_set('tacacs_server_group', 'servers', name: @name, state: '', server: s)
       end
       current_servs.each do |s|
         # remove any undesired existing servers
         next if new_servs.include? s
-        config_set('aaa_server_group', 'tacacs_server', @name, 'no', s)
+        config_set('tacacs_server_group', 'servers', name: @name, state: 'no', server: s)
       end
     end
 
     def default_servers
-      config_get_default('aaa_server_group', 'servers')
+      config_get_default('tacacs_server_group', 'servers')
     end
 
     def ==(other)
@@ -80,7 +79,7 @@ module Cisco
 
     def self.groups
       grps = {}
-      tacgroups = config_get('aaa_server_group', 'tacacs_groups') if
+      tacgroups = config_get('tacacs_server_group', 'group') if
         TacacsServer.enabled
       unless tacgroups.nil?
         tacgroups.each { |s| grps[s] = TacacsServerGroup.new(s, false) }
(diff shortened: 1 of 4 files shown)


lib/cisco_node_utils/tacacs_server_group.rb
```diff
@@ -17,91 +18,134 @@
...
+      config_set('tacacs_server_group', 'group', state: 'no', name: @name)
+    end
 
-      group_list.each do |id|
-        hash[id] = TacacsServerGroup.new(id, false)
+    def servers
+      tacservers = config_get('tacacs_server_group', 'servers', @name)
+      servs = {}
+      unless tacservers.nil?
+        tacservers.each { |s| servs[s] = TacacsServerHost.new(s, false) }
```
chrisvanheuveln: Confused... how does this work? 
You get a list of groups and then get a list of servers for each group; but it looks like the calls to `TacacsServerHost.new(s, false)` should just return right away because create is false...?

tests/test_tacacs_server_group.rb
```diff
@@ -17,98 +14,447 @@
 
 require_relative 'ciscotest'
 require_relative '../lib/cisco_node_utils/tacacs_server_group'
+require_relative '../lib/cisco_node_utils/tacacs_server_host'
 
-# TestTacacsServerGroup - Minitest for TacacsServerGroup node utility.
+DEFAULT_AAA_SERVER_GROUP_VRF = 'default'
+DEFAULT_AAA_SERVER_GROUP_DEADTIME = 0
+DEFAULT_AAA_SERVER_GROUP_SOURCE_INTERFACE = ''
```
chrisvanheuveln: Don't hard-code these, use the yaml defs:

```
DEFAULT_AAA_SERVER_GROUP_VRF = cmd_ref.lookup('tacacs_server_group', 'vrf')
DEFAULT_AAA_SERVER_GROUP_DEADTIME = cmd_ref.lookup('tacacs_server_group', 'deadtime')
DEFAULT_AAA_SERVER_GROUP_SOURCE_INTERFACE = cmd_ref.lookup('tacacs_server_group', 'source_interface')
```

tests/test_tacacs_server_group.rb
```diff
@@ -17,98 +14,447 @@
...
+    assert_raises(TypeError) do
+      TacacsServerGroup.new(nil)
+    end
+  end
+
+  def test_create_valid_tacacs
+    group_name = 'Group1'
+    aaa_group = TacacsServerGroup.new(group_name)
+    assert_show_match(command: 'show run tacacs+ all | no-more',
+                      pattern: /Group1/)
```
chrisvanheuveln: nit: could do

```
pattern: /#{group_name}/
```

tests/test_tacacs_server_group.rb
```diff
@@ -17,98 +14,447 @@
...
+
+  def test_collection_single_tacacs
+    clean_tacacs_config
+    group_name1 = 'Group1'
+    group_name2 = 'Group2'
+    group_name3 = 'Group3'
+    aaa_group1 = TacacsServerGroup.new(group_name1)
+
+    groups = TacacsServerGroup.groups
+    refute_empty(groups, 'Error: TacacsServerGroup collection is not filled')
```
chrisvanheuveln: Do you really need both of these? ie. checking both empty AND for size 1

tests/test_tacacs_server_group.rb
```diff
@@ -17,98 +14,447 @@
...
+    assert(groups.key?(group_name2),
+           "Error: TacacsServerGroup collection does contain #{group_name2}")
+    assert(groups.key?(group_name3),
+           "Error: TacacsServerGroup collection does contain #{group_name3}")
+
+    destroy_aaa_group(group_name2, 'tacacs+')
+    destroy_aaa_group(group_name3, 'tacacs+')
+  end
+
+  def test_collection_multi_tacacs
```
chrisvanheuveln: Is there something subtle here? What does this test offer that isn't tested in the previous test?

tests/test_tacacs_server_group.rb
```diff
@@ -17,98 +14,447 @@
...
+    clean_tacacs_config
+    server_name1 = 'server1'
+    server_name2 = 'server2'
+    server1 = create_tacacsserverhost(server_name1)
+    server2 = create_tacacsserverhost(server_name2)
+
+    aaa_group = TacacsServerGroup.new('Group1')
+    # aaa_group.servers = [server_name1, server_name2, server_name2]
+    # this behavior is different on n9k vs n3k, so comment out for now
+    # n3k throws a CLIError and n9k silently ignores
```
chrisvanheuveln: If this isn't a valid test then get rid of the comment (my vote).
If it is valid then it sounds like it "works" for n9k -- if it configs name1 & name2 ;
for n3k you can rescue it and validate/skip the result.

tests/test_tacacs_server_group.rb
```diff
@@ -17,98 +14,447 @@
...
+    servers = aaa_group.servers
+    assert_equal(2, servers.size,
+                 'Error: Collection is not two servers')
+
+    # Now remove them and then check again
+    aaa_group.servers = [server_name2]
+    refute_show_match(command: 'show run tacacs+ all | no-more',
+                      pattern: /server server1/)
+
+    # Now remove server 1 again
```
chrisvanheuveln: server 1? You're removing server 2...

tests/test_tacacs_server_group.rb
```diff
@@ -17,98 +14,447 @@
...
+
+  def test_set_vrf_tacacs
+    vrf = 'management-123'
+    aaa_group = TacacsServerGroup.new('Group1')
+    aaa_group.vrf = vrf
+    assert_show_match(command: 'show run tacacs+ all | no-more',
+                      pattern: /use-vrf management-123/)
+
+    # Invalid case
+    vrf = 2450
```
chrisvanheuveln: why is '2450' an invalid vrf? Also, no need to use a variable.

hunsbea: It's just because vrf= expects a string and not an int

PR #82, opened 2015-11-13
Default_value is now nil for unsupported nodes

glennmatthews:
As discussed by email, this changes the behavior for excluded attributes such that unsupported attributes are equivalent to an explicitly specified `default_only: ~` - they now return `nil` for `config_get_default` and `config_get`, but still raise `Cisco::UnsupportedError` on `config_set`.

``` bash
ruby tests/test_command_reference.rb 
Run options: --seed 6709

# Running:

........................

Finished in 0.872312s, 27.5131 runs/s, 88.2712 assertions/s.

24 runs, 77 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for MiniTest to /home/glmatthe/cisco-network-node-utils/coverage. 256 / 266 LOC (96.24%) covered.
```

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/README_YAML.md
+++ b/lib/cisco_node_utils/cmd_ref/README_YAML.md
@@ -238,7 +238,7 @@ attribute:
   config_set: 'attribute'
 ```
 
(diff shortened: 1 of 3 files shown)


RichWellum:
Looks good Glenn.

PR #86, opened 2015-11-16
Minor change to flatten confed peers

RichWellum:
While working on the provider and type for confederation_peers, discovered a nested array of peers was breaking the setter. Small fix to flatten the array.

### DIFF
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -351,6 +351,7 @@ module Cisco
     def confederation_peers=(should)
       # confederation peers are additive. So create a delta hash of entries
       # and only apply the changes
+      should = should.flatten
       is = confederation_peers
 
       delta_hash = Utils.delta_add_remove(should, is)


lib/cisco_node_utils/bgp.rb
```diff
@@ -351,6 +351,7 @@ def confederation_peers
     def confederation_peers=(should)
       # confederation peers are additive. So create a delta hash of entries
       # and only apply the changes
+      should = should.flatten
```
glennmatthews, line 354: Is there any check that you should do here to ensure the flatten is appropriate? Something like:

``` ruby
should = should.flatten if should[0].is_a?(Array)
```

RichWellum, line 354: Doesn't seem to hurt to flatten a non-nested array:

irb(main):001:0> [["1", "2", "3"]]
=> [["1", "2", "3"]]
irb(main):002:0> [["1", "2", "3"]].flatten
=> ["1", "2", "3"]
irb(main):003:0> ["1", "2", "3"].flatten
=> ["1", "2", "3"]

I've thrown a array (pun intended) of different inputs/strings/arrays at this - seems to be safe.

Will do something like the depth check if you insist?

PR #87, opened 2015-11-17
Add support for ios_xr platform in bgp_af

gvlaemyn:


lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -1,15 +1,24 @@
...
+      - '/^vrf <vrf>$/'
+      - '/^address-family <afi> <safi>$/'
+    config_set_append:
+      - 'vrf <vrf>'
+      - 'address-family <afi> <safi>'
+  cli_ios_xr:
+    config_get: 'show running-config router bgp'
+    config_get_token_append:
+      - '/^vrf <vrf>$/'
+      - '/^address-family <afi> <safi>$/'
```
glennmatthews: You should be able to share the config_get_token_append between both platforms since it's the same.

``` yaml
_template:
  config_get_token: '/^router bgp <asnum>$/'
  config_get_token_append:
      - '/^vrf <vrf>$/'
      - '/^address-family <afi> <safi>$/'
  cli_nexus:
  ...
```

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -46,9 +65,19 @@ all_afs:
...
-  default_value: true
+  cli_nexus:
+    config_get_token_append: '/^client-to-client reflection$/'
+    config_set_append: '<state> client-to-client reflection'
+    default_value: true
+  cli_ios_xr:
+    config_set: 'router bgp <asnum>'
+    config_set_append:
+      - ' address-family <afi> <safi>'
+      - ' <state> bgp client-to-client reflection disable'
```
glennmatthews, line 64: You're doing this to exclude VRF support for this property? Not sure this is the right way to address that - perhaps @RichWellum can comment?

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -46,9 +65,19 @@ all_afs:
...
+    config_get: 'show running-config router bgp | in bgp client-to-client'
+    config_get_token: '/^router bgp <asnum>$/'
+    config_get_token_append: '/^bgp client-to-client reflection disable$/'
```
glennmatthews: This looks wrong as well - your config_get now ignores both address-family and vrf?

RichWellum:
Can you add your mini test results (XR and NX) to the review please?

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -59,8 +88,16 @@ dampen_igp_metric:
...
-  config_set_append: '<state> dampening <route_map> <decay> <reuse> <suppress> <suppress_max>'
+  cli_nexus:
+    config_get_token_append: '/^dampening(?: (?:(\d+) (\d+) (\d+) (\d+)|route-map (.*)))?$/'
+    config_set_append: '<state> dampening <route_map> <decay> <reuse> <suppress> <suppress_max>'
+  cli_ios_xr:
+    config_set: 'router bgp <asnum>'
+    config_set_append:
+      - ' address-family <afi> <safi>'
+      - ' <state> bgp dampening <decay> <reuse> <suppress> <suppress_max>'
+    config_get: 'show running-config router bgp | in bgp dampening'
```
glennmatthews: Why is this needed? Can't you just use the template config_get of 'show running-config router bgp'?

tests/test_bgp_af.rb
```diff
@@ -106,6 +118,38 @@ def test_collection_not_empty
     end
   end
 
+  def config_ios_xr_dependencies(asn)
+    # These dependencies are required on ios xr
```
glennmatthews, line 114: Do we need to add any of this configuration to the BgpAf class itself? I know Rich had to add a few things to the Bgp class along these lines.

tests/test_bgp_af.rb
```diff
@@ -126,9 +172,13 @@ def test_default_information_originate
...
 
-    pattern = /^ *default-information originate$/
-    af_string = get_bgp_af_cfg(asn, vrf, af)
-
+    if platform == :nexus
+      pattern = /^ *default-information originate$/
+      af_string = get_bgp_af_cfg(asn, vrf, af)
+    elsif platform == :ios_xr
+      pattern = /^ default-information originate$/
+      af_string = @device.cmd("show run router bgp | in default-information")
```
glennmatthews: Why does `get_bgp_af_cfg` not work for XR?

tests/test_bgp_af.rb
```diff
@@ -152,51 +206,77 @@ def test_default_information_originate
...
+      refute_match(pattern, af_string, "bgp client-to-client disable is configured but should not be")
+      # Set and verify
+      bgp_af.client_to_client = true
+      af_string = @device.cmd("show run router bgp | in bgp client-to-client")
+      assert_match(pattern, af_string, "bgp client-to-client disable is configured and should be")
+      # Unset and verify
+      bgp_af.client_to_client = false
+      af_string = @device.cmd("show run router bgp | in bgp client-to-client")
+      refute_match(pattern, af_string, "bgp client-to-client disable is configured but should not be")
+    elsif platform == :nexus
```
glennmatthews, line 245: This seems wrong - I would expect you'd be able to share most of the test logic between the platforms. I'm also concerned that on Nexus you're testing the getter (`assert(bgp_af.client_to_client)`) but on XR it doesn't look like you are?

tests/test_bgp_af.rb
```diff
@@ -228,6 +308,39 @@ def test_next_hop_route_map
                    'configured and should not be')
   end
 
+  def test_next_hop_route_policy
```
glennmatthews: Are next-hop route-map and next-hop route-policy two different platforms' way of configuring the same behavior? If so I'd like to see a single test case for both rather than mutually exclusive test cases.

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
gvlaemyn, line 64: The commands for nexus and iOS-xr are different and appear at different levels.
Even the defaults are different for client-to-client reflection.

tests/test_bgp_af.rb
```diff
@@ -282,9 +408,9 @@ def test_additional_paths
     # Test getter
     #
 
-    assert(bgp_af.additional_paths_send)
-    assert(bgp_af.additional_paths_receive)
-    assert(bgp_af.additional_paths_install)
+    assert(bgp_af.additional_paths_send, "No additional-paths send configured")
+    assert(bgp_af.additional_paths_receive, "No additional-paths receive configured")
+    assert(bgp_af.additional_paths_install) if platform == :nexus
```
glennmatthews, line 462: Is this property unsupported on XR? If so then I'd like to see something like this in the test:

```
if platform == :nexus
  assert(bgp_af.additional_paths_install)
elsif platform == :ios_xr
  assert_nil(bgp_af.additional_paths_install)
end
```

tests/test_bgp_af.rb
```diff
@@ -305,7 +435,7 @@ def test_additional_paths
 
   ##
   ## additional_paths_selection route-map
-  ##
+  ##
```
glennmatthews: git and rubocop should have complained about trailing whitespace here - are you running rubocop? do you have the Git hooks installed?

tests/test_bgp_af.rb
RichWellum, line 114: I added rd-auto as we had no property for that. router-id is supported so it's fine to add this here.

Regarding rd-auto - it was added recently, but I am informed that it wasn't tested on XR. So three choices:
1. Add it here (but note when you come to puppet it will be broken)
2. Use the rd-auto property but make it work for XR (I have a US open for that). The we can use the setter here - and add it to the manifest in puppet.
3. Add it to the class as per my global bgp commit.

tests/test_bgp_af.rb
```diff
@@ -376,6 +520,8 @@ def advertise_l2vpn_evpn(asn, vrf, af)
   end
 
   def test_advertise_l2vpn_evpn
+    # no equivalent for IOS XR
+    return if platform != :nexus
```
glennmatthews: Instead, please do:

`skip('L2VPN EVPN is unsupported on IOS XR') if platform == :ios_xr`

tests/test_bgp_af.rb
```diff
@@ -456,23 +604,38 @@ def test_dampening
     ############################################
     # Set and verify 'dampening' with defaults #
     ############################################
-    vrf = 'orange'
+    if platform == :nexus
+      vrf = 'orange'
+    elsif platform == :ios_xr
+      vrf = 'default'
```
glennmatthews, line 678: Why not a vrf?

tests/test_bgp_af.rb
```diff
@@ -828,7 +1001,14 @@ def network_cmd(af, dbg)
                  "#{dbg} Test 3. Restore the removed networks")
 
     # Test: Change route-maps on existing networks
-    should = master.map { |network, rm| [network, "#{rm}_55"] }
+    if platform == :ios_xr
+      ['rtmap1_55', 'rtmap2_55', 'rtmap3_55', 'rtmap5_55', 'rtmap6_55', 'rtmap7_55'].each do |policy|
+        config("route-policy #{policy}", 'end')
```
glennmatthews, line 1126: Should you be adding a similar 'no route-policy...' to clean up at the end of the test?

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
glennmatthews, line 64: Different defaults are easy to handle in the YAML. By different levels do you mean that they're VRF-aware on NXOS but not VRF-aware on XR?

tests/test_bgp_af.rb
gvlaemyn, line 678: The command is not available under a named vrf in XR, only under default.  I found several commands like this.

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
gvlaemyn, line 64: VRF-aware on NXOS and a different command under address-family on XR

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
gvlaemyn: It can only appear in one place, so I just gripped for bgp client-to-client directly

tests/test_bgp_af.rb
gvlaemyn: get_bgp_af_cfg does "show run bgp all" which is not supported on XR

glennmatthews: Ok, but can't it be updated to do 'show run router bgp' on XR?

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
glennmatthews, line 64: My concern is that bgp_af is VRF-aware at an object level. By masking away this VRF on XR, you allow objects to conflict:

``` ruby
af1 = RouterBgpAF(100, 'red', 'ipv4 unicast')
af2 = RouterBgpAF(100, 'default', 'ipv4 unicast')
af1.client_to_client = true
# because client_to_client is not VRF-aware, now af2.client_to_client is also true, unexpectedly?
```

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
glennmatthews: Oh - so it's not even address-family aware? Then it definitely shouldn't be supported in `RouterBgpAF` on XR.

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
glennmatthews: The advantage of using the same config_get across all properties is that we can cache the CLI output and only call the device once. If you have a different config_get for each property, that's a different show command needed to call for each property.

tests/test_bgp_af.rb
gvlaemyn: It would involve more work because of the unsupported "sec" modifiers on XR:
    @device.cmd("show run bgp all | sec 'bgp #{asn}' |  sec 'vrf #{vrf}' | " \
                  "sec 'address-family #{afi} #{safi}' | no-more")

glennmatthews: Perhaps:

`@device.cmd('show run router bgp #{asn} vrf #{vrf} address-family #{afi} #{safi}')`

RichWellum: Yes like Glenn points out you can drill down to the exact command in XR.

E.g:

show run router bgp 55 vrf blue bgp enforce_as_first

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
gvlaemyn: ruby test_bgp_af.rb -v  -- 192.168.122.245:57888 root lab  
/home/gvlaemyn/.rvm/gems/ruby-2.2.1/gems/grpc-0.11.0/lib/grpc/generic/client_stub.rb:173: warning: circular argument reference - parent
Run options: -v -- --seed 3405

# Running:

Node under test:
- name  - 
- type  - R-IOSXRV9000-RP
  TestRouterBgpAF#test_collection_empty = 3.83 s = .
  TestRouterBgpAF#test_redistribute = 78.16 s = .
  TestRouterBgpAF#test_collection_not_empty = 4.15 s = .
  TestRouterBgpAF#test_dampen_igp_metric = 2.50 s = S
  TestRouterBgpAF#test_remove_route_policies = 15.07 s = .
  TestRouterBgpAF#test_additional_paths = 8.16 s = .
  TestRouterBgpAF#test_additional_paths_selection = 8.19 s = .
  TestRouterBgpAF#test_network = 138.94 s = .
  TestRouterBgpAF#test_advertise_l2vpn_evpn = 2.34 s = S
  TestRouterBgpAF#test_next_hop_route_policy = 14.13 s = .
  TestRouterBgpAF#test_client_to_client = 6.99 s = .
  TestRouterBgpAF#test_next_hop_route_map = 2.62 s = S
  TestRouterBgpAF#test_maximum_paths_ibgp = 20.27 s = .
  TestRouterBgpAF#test_dampening = 12.65 s = .
  TestRouterBgpAF#test_default_information_originate = 7.41 s = .
  TestRouterBgpAF#test_feature_nv_overlay_evpn = 2.37 s = S
  TestRouterBgpAF#test_maximum_paths = 21.81 s = .
  TestRouterBgpAF#test_utils_delta_add_remove = 2.32 s = .

Finished in 351.919095s, 0.0511 runs/s, 0.3609 assertions/s.

  1) Skipped:
TestRouterBgpAF#test_dampen_igp_metric [test_bgp_af.rb:534]:
dampen-igp-metric unsupported on IOS XR

  2) Skipped:
TestRouterBgpAF#test_advertise_l2vpn_evpn [test_bgp_af.rb:523]:
L2VPN EVPN is unsupported on IOS XR

  3) Skipped:
TestRouterBgpAF#test_next_hop_route_map [test_bgp_af.rb:279]:
nexthop route-map drop_all unsupported on IOS XR

  4) Skipped:
TestRouterBgpAF#test_feature_nv_overlay_evpn [test_bgp_af.rb:839]:
nv overlay evpn unsupported on IOS XR

18 runs, 127 assertions, 0 failures, 0 errors, 4 skips
Coverage report generated for MiniTest to /home/gvlaemyn/cisco-network-node-utils/tests/coverage. 0.0 / 0.0 LOC (100.0%) covered.

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -398,9 +422,15 @@ def dampening=(damp_array)
...
-        route_map.strip!
-        CiscoLogger.debug("Dampening 'dampening #{route_map}'")
+        if platform == :nexus
+          route_map = "route-map #{damp_array}"
+          route_map.strip!
+          CiscoLogger.debug("Dampening 'dampening #{route_map}'")
+        elsif platform == :ios_xr
+          route_policy = "route-policy #{damp_array}"
+          route_policy.strip!
+          CiscoLogger.debug("Dampening 'dampening #{route_policy}'")
```
glennmatthews, line 446: In the XR case, do you need to unset `route_map` after setting `route_policy`?

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -20,19 +26,29 @@ additional_paths_install:
...
 additional_paths_selection:
   kind: string
-  config_get_token_append: '/^additional-paths selection route-map (.*)$/'
-  config_set_append: '<state> additional-paths selection route-map <route_map>'
+  cli_nexus:
+    config_get_token_append: '/^additional-paths selection route-map (.*)$/'
+    config_set_append: '<state> additional-paths selection route-map <route_map>'
+  cli_ios_xr:
+    config_get_token_append: '/^additional-paths selection route-policy (.*)$/'
+    config_set_append: ' <state> additional-paths selection route-policy <route_policy>'
```
glennmatthews, line 36: Since Nexus never has a route-policy and XR never has a route-map, and they are used for similar purposes, would it make more sense to have both OS share a single <route_map_policy> key?

tests/test_bgp_af.rb
glennmatthews: Were you planning to address this and make get_bgp_af_cfg work for XR?

gvlaemyn: > On 20 Nov 2015, at 16:47, Glenn Matthews notifications@github.com wrote:
> 
> In tests/test_bgp_af.rb https://github.com/cisco/cisco-network-node-utils/pull/87#discussion_r45482597:
> 
> > @@ -126,9 +172,13 @@ def test_default_information_originate
> >      assert(bgp_af.default_information_originate,
> >             'Error: default-information originate not set')
> > -    pattern = /^ *default-information originate$/
> > 
> > ## \-    af_string = get_bgp_af_cfg(asn, vrf, af)
> > -    if platform == :nexus
> > -      pattern = /^ *default-information originate$/
> > -      af_string = get_bgp_af_cfg(asn, vrf, af)
> > -    elsif platform == :ios_xr
> > -      pattern = /^ default-information originate$/
> > -      af_string = @device.cmd("show run router bgp | in default-information")
> >   Were you planning to address this and make get_bgp_af_cfg work for XR?
> 
> The above was the old code.  The new code is:

 173     elsif platform == :ios_xr
 174       af_string = @device.cmd("show run router bgp #{asn} vrf #{vrf} " \
 175                               'default-information originate’)

get_bp_af_cfg is used for multiple commands on nexus
but there’s no equivalent in iOS_xr that can be used for all
so I use the appropriate extended show run router bgp command in each case.

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
gvlaemyn, line 36: I thought it was safer to use different keys for route-map and route-policy.
I think ios classic had both route-map and route-policy in some versions.

lib/cisco_node_utils/bgp_af.rb
gvlaemyn, line 446: I’m only doing config(“no route-policy …”) for iOS_xr

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -398,9 +422,15 @@ def dampening=(damp_array)
         CiscoLogger.debug("Dampening 'dampening #{damp_array.join(' ')}''")
       elsif route_map.is_a? String
```
glennmatthews: Actually this is the part that confused me with respect to the below code - @RichWellum we're setting `route_map = ''` above and not touching it again until we get here, so this `elsif route_map.is_a? String` will always pass - right? So really this is just an `else` in effect.

tests/test_bgp_af.rb
glennmatthews: Why won't my earlier suggestion of `@device.cmd('show run router bgp #{asn} vrf #{vrf} address-family #{afi} #{safi}')` work for this purpose?

lib/cisco_node_utils/bgp_af.rb
glennmatthews, line 446: I was confused by the check against route_map above (see comment to Rich). You can disregard this. :-)

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
glennmatthews, line 36: OK, makes sense I think.

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -20,19 +26,29 @@ additional_paths_install:
 additional_paths_receive:
   kind: boolean
   config_get_token_append: '/^additional-paths receive$/'
-  config_set_append: '<state> additional-paths receive'
+  cli_nexus:
+    config_set_append: '<state> additional-paths receive'
+  cli_ios_xr:
+    config_set_append: ' <state> additional-paths receive'
```
glennmatthews: Since the leading whitespace is harmless on Nexus, consider just adding the space to the shared config_set_append entry rather than creating separate Nexus/XR entries that differ only in whitespace.

lib/cisco_node_utils/bgp_af.rb
RichWellum: Yeah bad code agreed.

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
glennmatthews, line 64: As we discussed this morning - if this command is applicable under the default vrf only, you should check in bgp_af.rb when this property is being set and handle it correctly, something like:

``` diff
    def client_to_client
+      return nil if platform == :ios_xr && @vrf != 'default'
      state = config_get('bgp_af', 'client_to_client', @get_args)
      state ? true : false
    end

    def client_to_client=(state)
+      fail Cisco::UnsupportedError, 'client-to-client is not configurable on a per-VRF basis on IOS XR' \
+        if platform == :ios_xr && @vrf != 'default'
      state = (state ? '' : 'no')
      set_args_keys(state: state)
      config_set('bgp_af', 'client_to_client', @set_args)
    end

    def default_client_to_client
+      return nil if platform == :ios_xr && @vrf != 'default'
      config_get_default('bgp_af', 'client_to_client')
    end

```

and should explicitly test for this behavior in the minitest.

tests/test_bgp_af.rb
```diff
@@ -149,83 +199,191 @@ def test_default_information_originate
...
+                   'is not configured and should be')
+    end
   end
 
   ##
-  ## next_hop route-map
+  ## Negative tests with non-default vrf on ios_xr
   ##
-  def test_next_hop_route_map
+  def next_hop_route_policy_vrf_xr_neg_test
```
RichWellum, line 308: This is going to run on a nexus too and will fail.
I think these vrf cases need to be wrapped into the existing test code and have a platform check for XR.

if xr
  #test getter and setter for a non-default vrf
  assert/raise
  next
end

tests/test_bgp_af.rb
```diff
@@ -149,83 +199,191 @@ def test_default_information_originate
...
     asn = '55'
```
RichWellum, line 309: Consider writing a helper function to do this. Look at global bgp for example:

```
def setup_default
  @asnum = 55
  @vrf = 'default'
  RouterBgp.new(@asnum)
end

def setup_vrf
  @asnum = 99
  @vrf = 'yamllll'
  create_bgp_vrf(@asnum, @vrf)
end
```

Then either:

```
def test_bestpath_vrf
    bestpath(setup_vrf)
  end

  def bestpath(bgp)
...
```

or bgp = setup_vrf

tests/test_bgp_af.rb
gvlaemyn, line 308: I thought minitest only ran methods starting with “test_” automatically.
The call to next_hop_route_policy_vrf_xr_neg_test is done only if platform == :ios_xr:

 def test_next_hop_route_map_or_policy
    asn = '55'
    vrf = 'red'
    if platform == :ios_xr
      config_ios_xr_dependencies(asn)
      next_hop_route_policy_vrf_xr_neg_test
      vrf = 'default'
    end

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -101,17 +101,22 @@ def set_args_keys(hash={}) # rubocop:disable Style/AccessorMethodName
     # Client to client (Getter/Setter/Default)
     #
     def client_to_client
+      return nil if platform == :ios_xr && @vrf != 'default'
       state = config_get('bgp_af', 'client_to_client', @get_args)
       state ? true : false
     end
 
     def client_to_client=(state)
+      fail Cisco::UnsupportedError.new('bgp_af', 'client-to-client', 'set',
```
RichWellum, line 110: Not sure the '/' concatenated code is worth it. Maybe just:
if platform == :ios_xr && @vrf != 'default'
  fail code
end

?

tests/test_bgp_af.rb
RichWellum, line 308: Correct yeah.

I still feel the layout is a bit complicated - having to reset the vrf etc.

I'm not 100% happy with my layout but I think it's quite a bit cleaner:

``` ruby
def test_cluster_id_default
    cluster_id(setup_default)
  end

  def test_cluster_id_vrf
    cluster_id(setup_vrf)
  end

  def cluster_id(bgp)
    if platform == :ios_xr && @vrf != 'default'
      # XR does not support this under a VRF, so test the getter and setter
      assert_nil(bgp.cluster_id,
                 'cluster_is should return nil on XR with non-default vrf')
        bgp.cluster_id = 34
      end
    end
    bgp.cluster_id = 34
...
```

(Glenn: added code block fencing for readability)

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -159,6 +164,34 @@ def default_next_hop_route_map
       config_get_default('bgp_af', 'next_hop_route_map')
     end
 
+    def next_hop_route_policy
+      return nil if platform == :ios_xr && @vrf != 'default'
+      config_get('bgp_af', 'next_hop_route_policy', @get_args)
+    end
+
+    def next_hop_route_policy=(route_policy)
+      fail Cisco::UnsupportedError.new('bgp_af', 'nexthop route-policy', 'set',
```
mikewiebe, line 173: Can you move this `fail..` code into a local method called `check_supported` or something along those lines?  That way it can be extended for future cases easily and avoid the repetition.

lib/cisco_node_utils/bgp_af.rb
glennmatthews, line 110: Agreed. You can add a comment to disable rubocop if it's complaining:

``` ruby
if platform == :ios_xr && @vrf != 'default' # rubocop:disable Style/IfUnlessModifier
```

(but see Mike's alternative suggestion below)

lib/cisco_node_utils/bgp_af.rb
glennmatthews, line 173: Actually that's a better suggestion - something like `vrf_unsupported_xr`:

``` ruby
def vrf_unsupported_xr
  return if platform != :ios_xr || @vrf == 'default'
  fail Cisco::UnsupportedError.new(...
end
```

lib/cisco_node_utils/bgp_af.rb
glennmatthews: Geert, would you mind changing this line to 'else' and removing the unreachable 'else fail ArgumentError' below?

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
glennmatthews, line 64: Now that you're doing the check for VRF in the class itself, can you simplify the YAML to inherit config_set/config_set from the template?

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -59,8 +76,16 @@ dampen_igp_metric:
...
 dampening:
   auto_default: false
-  config_get_token_append: '/^dampening(?: (?:(\d+) (\d+) (\d+) (\d+)|route-map (.*)))?$/'
-  config_set_append: '<state> dampening <route_map> <decay> <reuse> <suppress> <suppress_max>'
+  cli_nexus:
+    config_get_token_append: '/^dampening(?: (?:(\d+) (\d+) (\d+) (\d+)|route-map (.*)))?$/'
+    config_set_append: '<state> dampening <route_map> <decay> <reuse> <suppress> <suppress_max>'
+  cli_ios_xr:
+    config_set_append:
+      - ' address-family <afi> <safi>'
```
glennmatthews, line 84: Isn't this a duplicate of the address-family line in the template?

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -59,8 +76,16 @@ dampen_igp_metric:
...
+      - ' <state> bgp dampening <route_policy> <decay> <reuse> <suppress> <suppress_max>'
+    config_get_token_append:
+      - '/^address-family <afi> <safi>$/'
```
glennmatthews, line 87: Ditto this - seems redundant.

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -83,9 +111,21 @@ dampening_suppress_time:
...
-  config_set_append: '<state> default-information originate'
-  default_value: false
+  cli_nexus:
+    config_get_token_append: '/^default-information originate$/'
+    config_set_append: '<state> default-information originate'
+    default_value: true
+  cli_ios_xr:
+    config_set: 'router bgp <asnum>'
+    config_set_append: 
+      - ' vrf <vrf>'
```
glennmatthews, line 121: Here as above, can you inherit from the template and do any XR exclusions in the class?

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -117,12 +165,28 @@ next_hop_route_map:
   config_set_append: '<state> nexthop route-map <route_map>'
   default_value: ""
 
+next_hop_route_policy:
+  cli_ios_xr:
+    config_set_append:
+      - 'address-family <afi> <safi>'
```
glennmatthews, line 171: redundant?

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -117,12 +165,28 @@ next_hop_route_map:
...
+      - '<state> nexthop route-policy <route_policy>'
+    config_get: 'show running-config router bgp'
+    config_get_token: '/^router bgp <asnum>$/'
+    config_get_token_append: '/nexthop route-policy (.*)$/'
```
glennmatthews, line 175: seems like this should inherit from the template - or else you are missing an address-family match line?

tests/test_bgp_af.rb
glennmatthews, line 114: @RichWellum @gvlaemyn What was the conclusion here? Is this going to be covered by Rich's open user story?

tests/test_bgp_af.rb
```diff
@@ -343,10 +533,17 @@ def test_additional_paths_selection
       bgp_af.default_additional_paths_selection
 
     af_string = get_bgp_af_cfg(asn, vrf, af)
+    if platform == :nexus
+      af_string = get_bgp_af_cfg(asn, vrf, af)
+    elsif platform == :ios_xr
+      af_string = @device.cmd("show run router bgp #{asn} vrf #{vrf} " \
+                              'address-family ipv4 unicast')
```
glennmatthews: I think my comment got lost in the shuffle - I would much prefer if you refactor `get_bgp_af_cfg` to work for both nexus and ios_xr: `@device.cmd('show run router bgp #{asn} vrf #{vrf} address-family #{afi} #{safi}')`

tests/test_bgp_af.rb
```diff
@@ -519,103 +756,146 @@ def test_dampening
...
+
+      error = ("Error: 'dampening' properties are incorrect")
+
+      assert_match(pattern_params1, af_string, error)
+      assert_match(pattern_params2, af_string, error)
+      assert_match(pattern_params3, af_string, error)
+      assert_match(pattern_params4, af_string, error)
+      assert_match(pattern_params5, af_string, error)
+
+      # Check getters
```
glennmatthews, line 801: Surely these should still work and be tested on XR as well?

tests/test_bgp_af.rb
```diff
@@ -457,50 +671,73 @@ def test_dampening
...
+      assert_match(pattern_params1, af_string, error)
+      assert_match(pattern_params2, af_string, error)
+      assert_match(pattern_params3, af_string, error)
+      assert_match(pattern_params4, af_string, error)
+      assert_match(pattern_params5, af_string, error)
+      # Check getter
+      assert_empty(bgp_af.dampening, 'Error: dampening is configured and ' \
+                   'should not be')
+    end
+    # "show bgp dampening param not supported on IOS XR"
```
glennmatthews, line 727: Seems like we should be testing the getters here for both NX and XR to make sure they return the correct default values.

tests/test_bgp_af.rb
```diff
@@ -519,103 +756,146 @@ def test_dampening
...
     # Set and verify 'dampening' with route-map #
     #############################################
-    vrf = 'brown'
-    af = %w(ipv6 unicast)
+    if platform == :nexus
+      vrf = 'brown'
+      af = %w(ipv6 unicast)
+    elsif platform == :ios_xr
+      vrf = 'default'
+      af = %w(ipv4 unicast)
```
glennmatthews: does XR not support ipv6 unicast af? Or why the discrepancy?

tests/test_bgp_af.rb
```diff
@@ -519,103 +756,146 @@ def test_dampening
...
+      assert_match(pattern_params, af_string,
+                   'Error: dampening properties DropAllTraffic is not ' \
                    'configured and should be')
 
-    # Check getters
-    assert_equal(bgp_af.dampening, 'DropAllTraffic',
-                 'Error: dampening getter did not match')
-    assert_equal(bgp_af.dampening_routemap, 'DropAllTraffic',
-                 'Error: dampening getter did not match')
+      # Check getters
```
glennmatthews, line 868: Getters should work on XR too right?

tests/test_bgp_af.rb
```diff
@@ -519,103 +756,146 @@ def test_dampening
...
+    end
 
     refute_match(pattern, af_string, "Error: 'dampening' is still configured")
 
+    config('no route-policy DropAllTraffic') if platform == :ios_xr
+
     #############################################
     # Set and verify 'dampening' with default   #
     #############################################
+    return if platform == :ios_xr
```
glennmatthews: Instead of this, can you split the test case in half here and have the second test case report as skipped on XR?

tests/test_bgp_af.rb
```diff
@@ -828,7 +1117,15 @@ def network_cmd(af, dbg)
                  "#{dbg} Test 3. Restore the removed networks")
 
     # Test: Change route-maps on existing networks
-    should = master.map { |network, rm| [network, "#{rm}_55"] }
+    if platform == :ios_xr
+      %w(rtmap1_55, rtmap2_55, rtmap3_55, rtmap5_55,
+         rtmap6_55, rtmap7_55).each do |policy|
+        config("route-policy #{policy}", 'end')
+      end
+      should = master.map { |network, rm| [network, rm.nil? ? nil : "#{rm}_55"] }
```
glennmatthews: Curious why you need a nil check for XR but not for NXOS?

tests/test_bgp_af.rb
RichWellum, line 114: @glennmatthews - I committed the rd auto property to feature/grpc last week - so '2' should work now, can look at latest global bgp @gvlaemyn .

Also @gvlaemyn don't forget to look at route-target which you will also inherit from the develop merge.

tests/test_bgp_af.rb
```diff
@@ -828,7 +1117,15 @@ def network_cmd(af, dbg)
                  "#{dbg} Test 3. Restore the removed networks")
 
     # Test: Change route-maps on existing networks
-    should = master.map { |network, rm| [network, "#{rm}_55"] }
+    if platform == :ios_xr
+      %w(rtmap1_55, rtmap2_55, rtmap3_55, rtmap5_55,
```
glennmatthews: Note that commas are not needed when using the %w syntax - this would be more correctly written as

``` ruby
%w(rtmap1_55 rtmap2_55 rtmap3_55 ...
```

Same comment elsewhere in this file.

tests/test_bgp_af.rb
```diff
@@ -840,6 +1137,11 @@ def network_cmd(af, dbg)
     result = af.networks
     assert_equal(should.sort, result.sort,
                  "#{dbg} Test 5. 'Default'")
+    %w(rtmap1, rtmap2, rtmap3, rtmap5, rtmap6, rtmap7,
+       rtmap1_55, rtmap2_55, rtmap3_55,
+       rtmap5_55, rtmap6_55, rtmap7_55).each do |policy|
+      config("no route-policy #{policy}")
+    end if platform == :ios_xr
```
glennmatthews: I'm not a big fan of putting an if conditional at the end of a block like this - it's not at all obvious. Would much prefer:

```
  return unless platform == :ios_xr
  %w(rtmap1 rtmap2...
     ...).each do |policy|
    ...
  end
end
```

tests/test_bgp_af.rb
```diff
@@ -149,83 +199,191 @@ def test_default_information_originate
...
     bgp_af = RouterBgpAF.new(asn, vrf, af)
-    bgp_af.next_hop_route_map = 'drop_all'
-    assert_match(bgp_af.next_hop_route_map, 'drop_all',
-                 'Error: nexthop route-map not set')
-    pattern = /^ *nexthop route-map drop_all$/
-    af_string = get_bgp_af_cfg(asn, vrf, af)
+    if platform == :nexus
+      bgp_af.next_hop_route_map = 'drop_all'
+    elsif platform == :ios_xr
+      config('route-policy drop_all', 'end') if platform == :ios_xr
```
RichWellum: I think you need to swap 'end' with 'end-policy'.

gvlaemyn: Indeed. Thanks!

tests/test_bgp_af.rb
gvlaemyn: without the nil check test_network fails with:

D, [2015-12-02T11:00:59.117090 #5537] DEBUG -- : CLI sent to device:
D, [2015-12-02T11:00:59.117138 #5537] DEBUG -- : router bgp 1
D, [2015-12-02T11:00:59.117177 #5537] DEBUG -- :  address-family ipv4 unicast
D, [2015-12-02T11:00:59.117213 #5537] DEBUG -- :   network 192.168.8.0/24 route-policy _55
D, [2015-12-02T11:00:59.117285 #5537] DEBUG -- : Sending 'cli_config' request with args: '#IOSXRExtensibleManagabilityService::CliConfigArgs:0x00000001347dc0'
D, [2015-12-02T11:00:59.252547 #5537] DEBUG -- : Handling 'cli_config' reply:
#IOSXRExtensibleManagabilityService::CliConfigReply:0x00000001346c18
D, [2015-12-02T11:00:59.252659 #5537] DEBUG -- : Reply includes errors:
{
 "cisco-grpc:errors": {
  "error": [
   {
    "error-type": "application",
    "error-tag": "operation-failed",
    "error-severity": "error",
    "error-message": "\nUnable to process cmd, ret-val: 3, cmd:   network 192.168.8.0/24 route-policy _55"
   }
  ]
 }
}

TestRouterBgpAF#test_network = 35.67 s = E

Finished in 35.669107s, 0.0280 runs/s, 0.0841 assertions/s.

  1) Error:
TestRouterBgpAF#test_network:
Cisco::CliError: CliError: 'input TODO' rejected with message:
'["\nUnable to process cmd, ret-val: 3, cmd:   network 192.168.8.0/24 route-policy _55"]'
    /home/gvlaemyn/cisco-network-node-utils/lib/cisco_node_utils/node.rb:271:in `rescue in config'
    /home/gvlaemyn/cisco-network-node-utils/lib/cisco_node_utils/node.rb:263:in`config'
    /home/gvlaemyn/cisco-network-node-utils/lib/cisco_node_utils/node.rb:162:in `config_set'
    /home/gvlaemyn/cisco-network-node-utils/lib/cisco_node_utils/node_util.rb:66:in`config_set'
    /home/gvlaemyn/cisco-network-node-utils/lib/cisco_node_utils/bgp_af.rb:557:in `block (2 levels) in networks='
    /home/gvlaemyn/cisco-network-node-utils/lib/cisco_node_utils/bgp_af.rb:551:in`each'
    /home/gvlaemyn/cisco-network-node-utils/lib/cisco_node_utils/bgp_af.rb:551:in `block in networks='
    /home/gvlaemyn/cisco-network-node-utils/lib/cisco_node_utils/bgp_af.rb:548:in`each'
    /home/gvlaemyn/cisco-network-node-utils/lib/cisco_node_utils/bgp_af.rb:548:in `networks='
    test_bgp_af.rb:1131:in`network_cmd'
    test_bgp_af.rb:1063:in `block (2 levels) in test_network'
    test_bgp_af.rb:1059:in`each'
    test_bgp_af.rb:1059:in `block in test_network'
    test_bgp_af.rb:1058:in`each'
    test_bgp_af.rb:1058:in `test_network’

tests/test_bgp_af.rb
gvlaemyn: Agree. I changed it back to ipv6 unicast on both nexus and xr.  
Don’t remember why I used ipv4 on xr.  It also works with ipv6.

tests/test_bgp_af.rb
glennmatthews: OK, but seems like this would be an issue for NXOS too. Why make it XR-specific?

gvlaemyn: I’ve only been able to test it on XR so far.  If I can get my hands on a Nexus I’ll test it.

tests/test_bgp_af.rb
gvlaemyn: OK. Split and added test_dampening_with_default test.

brturn:
@glennmatthews @RichWellum 
Please close this pull request; replaced by pull request #136

PR #89, opened 2015-11-17
Unsupported interface properties return nil on XR

glennmatthews:
Unsupported properties return nil on XR
- switchport
- switchport_mode
- switchport_trunk_allowed_vlan
- switchport_trunk_native_vlan
- switchport_vtp

Stop wrapping Cisco::CliError into RuntimeError

Refactor interface minitests:
- for properties unsupported on XR, explicitly test such instead of
  relying on the default handling of Cisco::UnsupportedError
- for test groups unsupported on XR, explicitly skip

No more auto-skips when running any of test_interface*.rb on XR.
Code coverage same or better than grpc baseline on XR and NXOS.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/interface.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface.yaml
@@ -231,34 +231,34 @@ switchport_autostate_exclude:
   default_value: false
 
 switchport_mode_dot1q_tunnel:
+  _exclude: [cli_ios_xr]
   default_only: ""
 
 switchport_mode_ethernet:
-  cli_nexus:
-    auto_default: false
-    config_get_token_append: '/^(?:no )?switchport mode ?(.*)$/'
-    config_set_append: "<state> switchport mode <mode>"
-    default_value: "access"
-  cli_ios_xr:
-    default_only: ""
+  _exclude: [cli_ios_xr]
+  auto_default: false
+  config_get_token_append: '/^(?:no )?switchport mode ?(.*)$/'
+  config_set_append: "<state> switchport mode <mode>"
+  default_value: "access"
 
 switchport_mode_other_interfaces:
+  _exclude: [cli_ios_xr]
   default_only: ""
 
 switchport_mode_port_channel:
-  cli_nexus:
-    config_get_token_append: '/^switchport mode (.*)$/'
-    config_set_append: "<state> switchport mode <mode>"
-    default_value: ""
-  cli_ios_xr:
-    default_only: ""
+  _exclude: [cli_ios_xr]
+  config_get_token_append: '/^switchport mode (.*)$/'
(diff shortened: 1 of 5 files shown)


lib/cisco_node_utils/cmd_ref/interface.yaml
```diff
@@ -231,34 +231,34 @@ switchport_autostate_exclude:
...
 
 switchport_mode_ethernet:
-  cli_nexus:
-    auto_default: false
-    config_get_token_append: '/^(?:no )?switchport mode ?(.*)$/'
-    config_set_append: "<state> switchport mode <mode>"
-    default_value: "access"
-  cli_ios_xr:
-    default_only: ""
+  _exclude: [cli_ios_xr]
```
RichWellum, line 238: Will the return nil work if you don't do exclude but explicitly just do nexus, for example?

lib/cisco_node_utils/interface.rb
```diff
@@ -450,6 +440,7 @@ def switchport_mode=(mode_set)
     end
 
     def default_switchport_mode
+      return nil if platform == :ios_xr
```
RichWellum, line 443: Nice! :clap:

tests/test_interface_switchport.rb
```diff
@@ -675,3 +386,268 @@ def test_interface_svi_command_on_non_vlan
...
+  attr_reader :svi
+
+  def setup
+    super
+    skip('VLAN interfaces are not supported on IOS XR') if platform == :ios_xr
+    @svi = Interface.new('Vlan23')
+  end
+
+  def teardown
+    svi.destroy unless platform == :ios_xr
```
RichWellum, line 403: maybe the xr check should be in the destroy method?

lib/cisco_node_utils/cmd_ref/interface.yaml
glennmatthews, line 238: doing `cli_nexus` instead of `_exclude: [cli_ios_xr]` will result in exactly the same behavior. Not sure exactly what you're asking here.

tests/test_interface_switchport.rb
glennmatthews, line 403: No can do - if we're on XR, svi doesn't ever get created (it's nil), so simply trying to call destroy raises an error.

lib/cisco_node_utils/cmd_ref/interface.yaml
RichWellum, line 238: I haven't been using the _exclude so wanted to check you've covered this.

tests/test_interface_switchport.rb
RichWellum, line 403: Ok ty.

RichWellum, line 403: :+1:

PR #90, opened 2015-11-17
Feature/grpc bgp fin

RichWellum:
Making enforce_first_as work on XR.

Minitest pass on XR and Nexus.

Puppet test - toggled enforce_first_as off and on in the following manifest - checked that it was applied and removed from both global VRF and non-global vrf correctly.

``` puppet
node 'default' {
  cisco_bgp { 'default_vrf_global':
    ensure                                 => present,
    asn                                    => 55,
    vrf                                    => 'default',
    router_id                              => '1.1.1.1',

    confederation_id                       => '33',
    cluster_id                             => '55',
    graceful_restart_timers_restart        => '131',
    graceful_restart_timers_stalepath_time => '311',
    bestpath_med_confed                    => true,
    graceful_restart => true,
    enforce_first_as                       => true,
}

  cisco_bgp { 'vrf_blue':
    ensure                                 => present,
    asn                                    => 55,
    vrf                                    => 'blue',
    router_id                              => '192.168.0.66',

    # Best Path Properties
    bestpath_always_compare_med            => true,
    bestpath_aspath_multipath_relax        => true,
    bestpath_compare_routerid              => true,
    bestpath_cost_community_ignore         => true,

    # Timer Properties
    timer_bgp_keepalive                    => '46',
    timer_bgp_holdtime                     => '111',

    enforce_first_as                       => true,
    }
  }
```

Rubocop happy on both ruby files, no puppet changes required.

### DIFF
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -327,10 +327,17 @@ module Cisco
       config_get('bgp', 'enforce_first_as', @get_args)
     end
 
-    def enforce_first_as=(enable)
-      @set_args[:state] = (enable ? '' : 'no')
-      config_set('bgp', 'enforce_first_as', @set_args)
-      set_args_keys_default
+    #    def enforce_first_as=(enable)
+    # XR: action is disable (on by default, but command is removed)
+    # Nexus: action is enable (on by default, but no command is nvgenned)
+    def enforce_first_as=(action)
+      # if platform == :ios_xr
+    # 
+      # else
+        @set_args[:state] = (action ? '' : 'no')
+        config_set('bgp', 'enforce_first_as', @set_args)
+        set_args_keys_default
+      # end
     end
 
     def default_enforce_first_as
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -327,17 +327,10 @@ module Cisco
       config_get('bgp', 'enforce_first_as', @get_args)
     end
 
-    #    def enforce_first_as=(enable)
-    # XR: action is disable (on by default, but command is removed)
-    # Nexus: action is enable (on by default, but no command is nvgenned)
     def enforce_first_as=(action)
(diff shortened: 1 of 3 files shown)


lib/cisco_node_utils/bgp.rb
```diff
@@ -327,8 +327,8 @@ def enforce_first_as
       config_get('bgp', 'enforce_first_as', @get_args)
     end
 
-    def enforce_first_as=(enable)
-      @set_args[:state] = (enable ? '' : 'no')
+    def enforce_first_as=(action)
+      @set_args[:state] = (action ? '' : 'no')
```
RichWellum: Not a functional change, just 'enable' to 'action' change as XR has the 'disable' keyword so you'd be 'enabling disable'. Which hurts my :bulb:

glennmatthews: Why not 'state'?

tests/test_router_bgp.rb
```diff
@@ -410,11 +409,15 @@ def test_set_get_enforce_first_as
   end
 
   def test_default_enforce_first_as
-    skip(XR_SUPPORTED_BROKEN) if platform == :ios_xr
     asnum = 55
     bgp = RouterBgp.new(asnum)
-    assert(bgp.enforce_first_as,
-           'bgp enforce-first-as value should be enabled = true')
+    if platform == :ios_xr
```
glennmatthews: I'd prefer not making this os-specific:

``` ruby
assert_equal(bgp.default_enforce_first_as,
             bgp.enforce_first_as,
             'bgp enforce-first-as default value is incorrect')
```

lib/cisco_node_utils/bgp.rb
RichWellum: Sure why not.

tests/test_router_bgp.rb
RichWellum: Yup fixed - ty.

glennmatthews:
:+1:

PR #94, opened 2015-11-18
Feature/aaa

hunsbea:
node utils classes for AAA authentication login, authorization service, authentication login service (config, exec). minitests/code has been updated to pass with all new develop changes. no rubocop failures, etc. I think it's a good idea to merge this now, as I will soon be adding the remaining puppet providers that rely on this code. It also adds visibility to work already completed to avoid any confusion/duplication.


tests/test_aaa_authentication_login.rb
```diff
@@ -0,0 +1,218 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative 'ciscotest'
+require_relative '../lib/cisco_node_utils/aaa_authentication_login'
+
+# Test class for AAA Authentication Login
+class TestAaaAuthenticationLogin < CiscoTestCase
+  DEFAULT_AAA_AUTHENTICATION_LOGIN_ASCII_AUTH = false
```
mikewiebe: Why define these again here instead of just pulling them from the yaml?  Avoid updating in two places if it changes.

lib/cisco_node_utils/aaa_authentication_login.rb
```diff
@@ -0,0 +1,96 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require_relative 'node_util'
+
+module Cisco
+  # NXAPI implementation of AAA Authentication Login class
+  class AaaAuthenticationLogin < NodeUtil
```
mikewiebe, line 25: Explore making this a singleton... change IF it makes sense :).

tests/test_aaa_authentication_login.rb
```diff
@@ -0,0 +1,218 @@
...
+    assert_equal(DEFAULT_AAA_AUTHENTICATION_LOGIN_ASCII_AUTH,
+                 aaaauthlogin.default_ascii_authentication,
+                 'Error: AAA authentication login, default ascii incorrect')
+    aaaauthenticationlogin_detach(aaaauthlogin)
+  end
+
+  def test_set_ascii_authentication
+    aaaauthlogin = AaaAuthenticationLogin
+
+    aaaauthlogin.ascii_authentication = true
```
mikewiebe, line 63: Better to make this `!DEFAULT_AAA_AUTHENTICATION_LOGIN_ASCII_AUTH`.  Philosophically of course...

hunsbea, line 63: Now that you mention it, I think I actually want to change the DEFAULT below to be false instead. Since I'm specifically checking for the presence and absence of the configuration, I don't want to be assuming the default in either case.

PR #96, opened 2015-11-19
Correctly remove/restore IP when changing VRF

glennmatthews:
When testing with Puppet I found an issue - XR doesn't let you change VRF if there's an IP address configured:

`"'RSI' detected the 'fatal' condition 'The interface's numbered and unnumbered IPv4/IPv6 addresses must be removed prior to changing or deleting the VRF'"`

Additionally, when changing the VRF on Nexus, any existing IP address configuration is lost.

Solution:
- Add minitest case for changing VRF with an IP address present
- in Interface class, get current IP address before changing VRF, remove it (if on XR), and restore it after changing VRF.

### DIFF
--- a/lib/cisco_node_utils/interface.rb
+++ b/lib/cisco_node_utils/interface.rb
@@ -578,13 +578,20 @@ module Cisco
       config_get('interface', 'vrf', name: @name)
     end
 
-    def vrf=(vrf)
-      fail TypeError unless vrf.is_a?(String)
-      if vrf.empty?
+    def vrf=(v)
+      fail TypeError unless v.is_a?(String)
+      return if v == vrf
+      # Changing the VRF can result in loss of IP address, so cache it
+      addr_mask = ipv4_addr_mask
+      addr_mask = addr_mask[0] unless addr_mask.nil?
+      # XR actually blocks you from changing the VRF if IP addr is present
+      ipv4_addr_mask_set(nil, nil) unless addr_mask.nil? || platform == :nexus
+      if v.empty?
         config_set('interface', 'vrf', name: @name, state: 'no', vrf: '')
       else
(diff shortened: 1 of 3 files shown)


lib/cisco_node_utils/interface.rb
```diff
@@ -578,13 +578,20 @@ def vrf
       config_get('interface', 'vrf', name: @name)
     end
 
-    def vrf=(vrf)
-      fail TypeError unless vrf.is_a?(String)
-      if vrf.empty?
+    def vrf=(v)
+      fail TypeError unless v.is_a?(String)
+      return if v == vrf
```
RichWellum, line 583: Previously 'vrf' was passed in - what is it now?

glennmatthews, line 583: Call to the `vrf` instance method to get the current VRF setting - so that we can correctly do a no-op.

RichWellum, line 583: Ah you explained verbally that vrf is like vrf() the getter... Thanks!

tests/test_interface.rb
```diff
@@ -1136,4 +1136,36 @@ def test_interface_vrf_valid
...
+                 'IPv4 mask wrong after changing from vrf test1 => test2')
+    assert_equal(vrf2, interface.vrf)
+
+    interface.vrf = DEFAULT_IF_VRF
+    assert_equal(address, interface.ipv4_address,
+                 'IPv4 address wrong after changing from vrf test2 => default')
+    assert_equal(length, interface.ipv4_netmask_length,
+                 'IPv4 mask wrong after changing from vrf test2 => default')
+    assert_equal(DEFAULT_IF_VRF, interface.vrf)
+  end
```
RichWellum, line 1170: :+1:

PR #97, opened 2015-11-20
acl&ace node utils, mini test and yaml cmd file

yjyongz:
[18:02 shenma-laas]($):$ ruby test_acl.rb -v -- 10.122.197.183 admin admin
Run options: -v -- --seed 11454
# Running:

Node under test:
- name  - agent-lab3-nx
- type  - N9K-NXOSV

TestAcl#test_router_stats_enable = 6.77 s = .
TestAcl#test_router_stats_enable_v6 = 4.45 s = .
TestAcl#test_router_create_acl_v4 = 4.25 s = .
TestAcl#test_router_create_acl_v6 = 4.09 s = .

Finished in 19.564593s, 0.2045 runs/s, 1.2267 assertions/s.

4 runs, 24 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for MiniTest to /auto/tftp-sjc-users3/shenma/cisco-network-node-utils/tests/coverage. 0.0 / 0.0 LOC (100.0%) covered.
# 

[02:41 shenma-laas]:$ ruby test_ace_v4.rb -v -- 10.122.197.183 admin admin
Run options: -v -- --seed 55981
# Running:

Node under test:
- name  - agent-lab3-nx
- type  - N9K-NXOSV

TestAceV4#test_router_create_one = 7.94 s = .

Finished in 7.939242s, 0.1260 runs/s, 0.7557 assertions/s.

1 runs, 6 assertions, 0 failures, 0 errors, 0 skips
# Coverage report generated for MiniTest to /auto/tftp-sjc-users3/shenma/cisco-network-node-utils/tests/coverage. 0.0 / 0.0 LOC (100.0%) covered.

[02:46 shenma-laas]:$ rubocop test_ace_v4.rb 
Inspecting 1 file
.
#1 file inspected, no offenses detected

[02:50 shenma-laas]:$ rubocop tests/test_acl.rb 
Inspecting 1 file
.
#1 file inspected, no offenses detected

[02:59 shenma-laas]:$ rubocop lib/cisco_node_utils/ace_v4.rb
Inspecting 1 file
.
#1 file inspected, no offenses detected

[03:02 shenma-laas]:$ rubocop lib/cisco_node_utils/acl.rb
Inspecting 1 file
.

1 file inspected, no offenses detected

### DIFF
--- a/lib/cisco_node_utils/ace_v4.rb
+++ b/lib/cisco_node_utils/ace_v4.rb
@@ -21,20 +21,20 @@ module Cisco
 
     # name: name of the router instance
     # instantiate: true = create router instance
-    def initialize(acl_name, seqno, afi, instantiate=true)
+    def initialize(acl_name, seqno, afi)
       @acl_name = acl_name
       @afi = afi
       @seqno = seqno
-      @set_args = @get_args = {acl_name: @acl_name, seqno: @seqno}
+      @set_args = @get_args = { acl_name: @acl_name, seqno: @seqno }
     end
 
     # Create a hash of all current router instances.
     def self.aces
-      instances = config_get('acl_v4' , 'ace')
+      instances = config_get('acl_v4', 'ace')
       return {} if instances.nil?
       hash = {}
       instances.each do |name|
-          hash[name] = Ace.new(acl_name, false)
+        hash[name] = Ace.new(acl_name, false)
       end
       return hash
     rescue Cisco::CliError => e
@@ -42,7 +42,7 @@ module Cisco
       raise unless e.clierror =~ /Syntax error/
       return {}
     end
-    
+
     # Destroy a router instance; disable feature on last instance
     def destroy
       config_ace('no')
@@ -64,7 +64,7 @@ module Cisco
     def seqno
       match = config_get('acl_v4', 'ace', @get_args)
       return if match.nil?
-      seqno = match[0]
+      match[0]
     end
 
     # setter of seqno
@@ -76,7 +76,7 @@ module Cisco
     def action
       match = config_get('acl_v4', 'ace', @get_args)
       return if match.nil?
-      action = match[1]
+      match[1]
     end
 
     # setter of action
@@ -88,9 +88,9 @@ module Cisco
     def proto
       match = config_get('acl_v4', 'ace', @get_args)
       return if match.nil?
-      proto = match[2]
+      match[2]
     end
-    
+
     # setter of proto
     def proto=(proto)
       @set_args[:proto] = proto
@@ -100,9 +100,9 @@ module Cisco
     def v4_src_addr_format
       match = config_get('acl_v4', 'ace', @get_args)
       return if match.nil?
-      src_addr = match[3]
+      match[3]
     end
-    
+
     # setter of v4_src_addr_format
     def v4_src_addr_format=(src_addr)
       @set_args[:v4_src_addr_format] = src_addr
@@ -112,9 +112,9 @@ module Cisco
     def v4_src_port_format
       match = config_get('acl_v4', 'ace', @get_args)
       return if match.nil?
-      src_port = match[5]
+      match[5]
     end
-    
+
     # setter of v4_src_port_format
     def v4_src_port_format=(src_port)
       @set_args[:v4_src_port_format] = src_port
@@ -124,9 +124,9 @@ module Cisco
     def v4_dst_addr_format
       match = config_get('acl_v4', 'ace', @get_args)
       return if match.nil?
-      dst_addr = match[6]
+      match[6]
     end
-    
+
     # setter of v4_dst_addr_format
     def v4_dst_addr_format=(dst_addr)
       @set_args[:v4_dst_addr_format] = dst_addr
@@ -136,9 +136,9 @@ module Cisco
     def v4_dst_port_format
       match = config_get('acl_v4', 'ace', @get_args)
       return if match.nil?
-      src_port = match[7]
+      match[7]
     end
-    
+
     # setter of v4_dst_port_format
     def v4_dst_port_format=(src_port)
       @set_args[:v4_dst_port_format] = src_port
@@ -148,9 +148,9 @@ module Cisco
     def option_format
       match = config_get('acl_v4', 'ace', @get_args)
       return if match.nil?
-      option_format = match[7]
+      match[8]
     end
-    
+
     # setter of option_format
     def option_format=(option)
       @set_args[:option_format] = option
--- a/lib/cisco_node_utils/ace_v4.rb
+++ b/lib/cisco_node_utils/ace_v4.rb
@@ -1,0 +1,159 @@
+# Copyright (c) 2014-2015 Cisco and/or its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative 'node_util'
+
+module Cisco
+  # RouterAcl - node utility class for RouterAcl config mgmt.
+  class Ace < NodeUtil
+    attr_reader :acl_name, :seqno, :afi
+
+    # name: name of the router instance
+    # instantiate: true = create router instance
+    def initialize(acl_name, seqno, afi, instantiate=true)
+      @acl_name = acl_name
+      @afi = afi
+      @seqno = seqno
+      @set_args = @get_args = {acl_name: @acl_name, seqno: @seqno}
+    end
+
+    # Create a hash of all current router instances.
+    def self.aces
+      instances = config_get('acl_v4' , 'ace')
+      return {} if instances.nil?
+      hash = {}
+      instances.each do |name|
+          hash[name] = Ace.new(acl_name, false)
+      end
+      return hash
+    rescue Cisco::CliError => e
+      # CLI will syntax reject when feature is not enabled
+      raise unless e.clierror =~ /Syntax error/
+      return {}
+    end
+    
+    # Destroy a router instance; disable feature on last instance
+    def destroy
+      config_ace('no')
+    rescue Cisco::CliError => e
+      # CLI will syntax reject when feature is not enabled
+      raise unless e.clierror =~ /Syntax error/
+    end
+
+    # First create acl
+    def config_ace(state='')
+      @set_args[:state] = state
+      config_set('acl_v4', 'ace', @set_args)
+    end
+
+    # ----------
+    # PROPERTIES
+    # ----------
+    # getter of seqno
+    def seqno
+      match = config_get('acl_v4', 'ace', @get_args)
+      return if match.nil?
+      seqno = match[0]
+    end
+
+    # setter of seqno
+    def seqno=(state)
+      @set_args[:state] = state ? '' : 'no'
+    end
+
+    # getter of action
+    def action
+      match = config_get('acl_v4', 'ace', @get_args)
+      return if match.nil?
+      action = match[1]
+    end
+
+    # setter of action
+    def action=(action)
+      @set_args[:action] = action
+    end
+
+    # getter of proto
+    def proto
+      match = config_get('acl_v4', 'ace', @get_args)
+      return if match.nil?
+      proto = match[2]
+    end
+    
+    # setter of proto
+    def proto=(proto)
+      @set_args[:proto] = proto
+    end
+
+    # getter of v4_src_addr_format
+    def v4_src_addr_format
+      match = config_get('acl_v4', 'ace', @get_args)
+      return if match.nil?
+      src_addr = match[3]
+    end
+    
+    # setter of v4_src_addr_format
+    def v4_src_addr_format=(src_addr)
+      @set_args[:v4_src_addr_format] = src_addr
+    end
+
+    # getter of v4_src_port_format
+    def v4_src_port_format
+      match = config_get('acl_v4', 'ace', @get_args)
+      return if match.nil?
+      src_port = match[5]
+    end
+    
+    # setter of v4_src_port_format
+    def v4_src_port_format=(src_port)
+      @set_args[:v4_src_port_format] = src_port
+    end
+
+    # getter of v4_dst_addr_format
+    def v4_dst_addr_format
+      match = config_get('acl_v4', 'ace', @get_args)
+      return if match.nil?
+      dst_addr = match[6]
+    end
+    
+    # setter of v4_dst_addr_format
+    def v4_dst_addr_format=(dst_addr)
+      @set_args[:v4_dst_addr_format] = dst_addr
+    end
+
+    # getter of v4_dst_port_format
+    def v4_dst_port_format
+      match = config_get('acl_v4', 'ace', @get_args)
+      return if match.nil?
+      src_port = match[7]
+    end
+    
+    # setter of v4_dst_port_format
+    def v4_dst_port_format=(src_port)
+      @set_args[:v4_dst_port_format] = src_port
+    end
+
+    # getter of option_format
+    def option_format
+      match = config_get('acl_v4', 'ace', @get_args)
+      return if match.nil?
+      option_format = match[7]
+    end
+    
+    # setter of option_format
+    def option_format=(option)
+      @set_args[:option_format] = option
+    end
+  end
+end
--- a/lib/cisco_node_utils/acl.rb
+++ b/lib/cisco_node_utils/acl.rb
@@ -24,13 +24,13 @@ module Cisco
     def initialize(acl_name, afi, instantiate=true)
       @acl_name = acl_name
       @afi = afi
-      @set_args = @get_args = {acl_name: @acl_name} 
+      @set_args = @get_args = { acl_name: @acl_name }
       create if instantiate
     end
 
     def self.acls
-      instances = config_get('acl_v4' , 'acl') if @afi == "v4"
-      instances = config_get('acl_v6' , 'acl') if @afi == "v6"
+      instances = config_get('acl_v4', 'acl') if @afi == 'v4'
+      instances = config_get('acl_v6', 'acl') if @afi == 'v6'
       return {} if instances.nil?
       hash = {}
       instances.each do |name|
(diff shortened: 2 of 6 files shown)


lib/cisco_node_utils/acl.rb
```diff
@@ -0,0 +1,94 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative 'node_util'
+
+module Cisco
+  # RouterAcl - node utility class for RouterAcl config mgmt.
```
mikewiebe, line 18: /RouterAcl/Acl/

lib/cisco_node_utils/acl.rb
```diff
@@ -0,0 +1,94 @@
...
+  class Acl < NodeUtil
+    attr_reader :acl_name, :afi
+
+    # name: name of the router instance
```
mikewiebe, line 22: **nit** Name of the acl instance?

lib/cisco_node_utils/acl.rb
```diff
@@ -0,0 +1,94 @@
...
+    # instantiate: true = create router instance
+    def initialize(acl_name, afi, instantiate=true)
+      @acl_name = acl_name
```
mikewiebe, line 25: Please add some error checking for `acl_name` and `afi`.  Also, I personally would enforce `ipv4` and `ipv6` as the only valid `afi` options for the constructor.

```
def initialize(acl_name, afi, instantiate=true)
  fail TypeError unless acl_name.is_a?(String) && afi.is_a?(String)
  fail ArgumentError unless afi == 'ipv4' || afi == 'ipv6'
  ...
end
```

lib/cisco_node_utils/ace_v4.rb
```diff
@@ -0,0 +1,159 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative 'node_util'
+
+module Cisco
+  # RouterAcl - node utility class for RouterAcl config mgmt.
```
chrisvanheuveln, line 18: This is the `ace_v4` file... your comment says `RouterAcl`...? You have `class Ace` below but the filename suggests v4 only? Your initialize has an `afi` parameter...? 
Please let us know if this is intended for v4 only or both v4/v6 or something else altogether.

edit: Okay I see v6 calls in here as well. This implementation needs a fair amount of rework to support both afi's. And since it supports both it should not be in a file called `ace_v4.rb`.

lib/cisco_node_utils/ace_v4.rb
```diff
@@ -0,0 +1,159 @@
...
+  class Ace < NodeUtil
+    attr_reader :acl_name, :seqno, :afi
+
+    # name: name of the router instance
+    # instantiate: true = create router instance
```
chrisvanheuveln, line 23: `%s/router instance/ace instance/`

You don't have `name` or `instantiate` variables.

lib/cisco_node_utils/ace_v4.rb
```diff
@@ -0,0 +1,159 @@
...
+    def initialize(acl_name, seqno, afi)
+      @acl_name = acl_name
+      @afi = afi
+      @seqno = seqno
+      @set_args = @get_args = { acl_name: @acl_name, seqno: @seqno }
+    end
+
+    # Create a hash of all current router instances.
+    def self.aces
+      instances = config_get('acl_v4', 'ace')
```
chrisvanheuveln, line 33: If this provider is intended to support v6 as well then `self.aces` is not going to work very well.

lib/cisco_node_utils/acl.rb
```diff
@@ -0,0 +1,94 @@
...
+      @afi = afi
+      @set_args = @get_args = { acl_name: @acl_name }
+      create if instantiate
+    end
+
+    def self.acls
+      instances = config_get('acl_v4', 'acl') if @afi == 'v4'
+      instances = config_get('acl_v6', 'acl') if @afi == 'v6'
```
chrisvanheuveln, line 33: 1. For consistency with all of our other providers it would be better if you use `ipv4` and `ipv6` for the afi strings.
2. There's a lot of redundancy here. You could simplify these if you set a class variable in intialize and then just use that for each of the getter/setters, ie.:

```
def initialize( ...
  @afi_str = "acl_v#{$1}" if afi[/ipv(\d)/]    # assumes string is 'ipv4' or 'ipv6'
   ...

def self.acls
  instances = config_get(@afi_str, 'acl')
  ...

def acl
  config_get(@afi_str, 'acl', @get_args)
end

def acl=(state)
  @set_args[:state] = (state ? '' : 'no')
  config_set(@afi_str, 'acl', @set_args)
end
```

lib/cisco_node_utils/cmd_ref/acl_v4.yaml
```diff
@@ -0,0 +1,25 @@
+# Command Reference Common ACL
```
mikewiebe, line 1: You can reduce the redundancy here and collapse `acl_v4.yaml` and `acl_v6.yaml` into a single file `acl.yaml` that handles both.  Also `acl.yaml` per our type design email thread should only have `statistics` and `fragments` properties.  The ace properties should go in a separate `ace.yaml` file.

Update:  Actually thinking about this some more, keeping all of the acl/ace properties in one file called `acl.yaml` is fine, but add the `fragments` property.

lib/cisco_node_utils/cmd_ref/acl_v4.yaml
```diff
@@ -0,0 +1,25 @@
...
+#
+# For documentation please see:
+#   - README_YAML.md
+#
+---
+_template:
+  config_get: 'show run aclmgr'
+  config_get_token: '/^ip access-list <acl_name>$/'
```
mikewiebe, line 9: Change to `config_get_token: '/^<afi> access-list <acl_name>$/'`

lib/cisco_node_utils/cmd_ref/acl_v4.yaml
```diff
@@ -0,0 +1,25 @@
...
+  config_set: 'ip access-list <acl_name>'
```
mikewiebe, line 10: Change to `config_set: '<afi> access-list <acl_name>'`

lib/cisco_node_utils/cmd_ref/acl_v4.yaml
```diff
@@ -0,0 +1,25 @@
...
+ace:
+  config_get_token_append: '/\s+(<seqno>)\s+(permit|deny)\s+(sctp|tcp|ip|udp|pim|icmp|igmp|gre|nos|pcp|ospf|eigrp|ahp|esp|\d+)\s+(any|host \S+|\S+\/\d+|\S+ [:\.0-9a-fA-F]+)\s+(eq \d+|neq \d+|lt \d+|gt \d+|range \d+ \d+)*\s*(any|host \S+|\S+\/\d+|\S+ [:\.0-9a-fA-F]+)\s*(eq \d+|neq \d+|lt \d+|gt \d+|range \d+ \d+)*( [a-zA-Z0-9\/ ]*)*$'
+  config_set_append: 
+      '<state> <seqno> <action> <proto> <v4_src_addr_format> <v4_src_port_format> <v4_dst_addr_format> <v4_dst_port_format> <option_format>'
+
+acl:
+  config_get_token: '/^ip access-list (/s+)$/'
+  config_set: '<state> ip access-list <acl_name>'
+                      
+stats_perentry:
```
chrisvanheuveln, line 21: camel_case

lib/cisco_node_utils/cmd_ref/acl_v4.yaml
```diff
@@ -0,0 +1,25 @@
...
+#   - README_YAML.md
+#
+---
+_template:
+  config_get: 'show run aclmgr'
+  config_get_token: '/^ip access-list <acl_name>$/'
+  config_set: 'ip access-list <acl_name>'
+
+ace:
+  config_get_token_append: '/\s+(<seqno>)\s+(permit|deny)\s+(sctp|tcp|ip|udp|pim|icmp|igmp|gre|nos|pcp|ospf|eigrp|ahp|esp|\d+)\s+(any|host \S+|\S+\/\d+|\S+ [:\.0-9a-fA-F]+)\s+(eq \d+|neq \d+|lt \d+|gt \d+|range \d+ \d+)*\s*(any|host \S+|\S+\/\d+|\S+ [:\.0-9a-fA-F]+)\s*(eq \d+|neq \d+|lt \d+|gt \d+|range \d+ \d+)*( [a-zA-Z0-9\/ ]*)*$'
```
mikewiebe, line 13: Some notes about this regular expression.
1) You don't need the \s+ before the (<seqno>).  You can anchor it... `'/^(seqno)...$/`
2) Instead of `(permit|deny)` `(sctp|tcp|ip|udp|pim...)` etc, just return any non-whitespace value `(\S+)`  That way if new values are supported in the future at that position we already handle it.
3) You omitted the trailing `/` to complete the regular expression.
4) Please add test cases to thoroughly test the different kinds of ace's your regular expression should match.  I did a quick test and the following does not match with this regexp.

```
2.2.1 :026 > ace = '10 permit ip 4.4.4.4 255.255.255.0 8.8.8.8 255.255.255.0 fragments'
 => "10 permit ip 4.4.4.4 255.255.255.0 8.8.8.8 255.255.255.0 fragments" 
2.2.1 :027 > /\s+(10)\s+(permit|deny)\s+(sctp|tcp|ip|udp|pim|icmp|igmp|gre|nos|pcp|ospf|eigrp|ahp|esp|\d+)\s+(any|host \S+|\S+\/\d+|\S+ [:\.0-9a-fA-F]+)\s+(eq \d+|neq \d+|lt \d+|gt \d+|range \d+ \d+)*\s*(any|host \S+|\S+\/\d+|\S+ [:\.0-9a-fA-F]+)\s*(eq \d+|neq \d+|lt \d+|gt \d+|range \d+ \d+)*( [a-zA-Z0-9\/ ]*)*$/.match(ace)
 => nil 
2.2.1 :028 > ace = '10 permit ip 4.4.4.4 255.255.255.0 8.8.8.8 255.255.255.0'
 => "10 permit ip 4.4.4.4 255.255.255.0 8.8.8.8 255.255.255.0" 
2.2.1 :029 > /\s+(10)\s+(permit|deny)\s+(sctp|tcp|ip|udp|pim|icmp|igmp|gre|nos|pcp|ospf|eigrp|ahp|esp|\d+)\s+(any|host \S+|\S+\/\d+|\S+ [:\.0-9a-fA-F]+)\s+(eq \d+|neq \d+|lt \d+|gt \d+|range \d+ \d+)*\s*(any|host \S+|\S+\/\d+|\S+ [:\.0-9a-fA-F]+)\s*(eq \d+|neq \d+|lt \d+|gt \d+|range \d+ \d+)*( [a-zA-Z0-9\/ ]*)*$/.match(ace)
 => nil 
2.2.1 :030 > 
```

tests/test_ace_v4.rb
```diff
@@ -0,0 +1,64 @@
...
+  # TESTS
+
+  def test_router_create_one
+    acl = Acl.new(@acl_name, @afi)
+    ace = Ace.new(@acl_name, @seqno, @afi)
+    ace.action = 'permit'
+    ace.proto  = 'tcp'
+    ace.v4_src_addr_format = '7.8.9.6 2.3.4.5'
+    ace.v4_src_port_format = 'eq 40'
+    ace.v4_dst_addr_format = '1.2.3.4/32'
```
chrisvanheuveln, line 52: Do we have to support both formats? `x.x.x.x  y.y.y.y` and `x.x.x.x/y`

Just thinking that we can require that the manifest use slash-notation and normalize it to whatever the switch uses.

tests/test_ace_v4.rb
```diff
@@ -0,0 +1,64 @@
...
+    ace.v4_dst_port_format = 'neq 20'
```
chrisvanheuveln, line 53: Is `eq` and `neq` standard across our platforms and yang? If so then okay, otherwise I have some concern about having the user having to know the syntax, and in which case we might have to resort to something like:

```
src_port_eq => 40
dst_port_neq => 20
```

Mike, chime in if you have any thoughts on this one.

lib/cisco_node_utils/cmd_ref/acl_v4.yaml
mikewiebe, line 13: My suggestion would be to use a simpler regexp like the following.

`/<seqno> (/S+) (/S+) (.*)/` and then do all your regexp post processing the the object methods.  The reason I chose (.*) for everything the follows the protocol is because there are many cases you will have to match and it would be simpler to match in a method rather then try and match everything in the yaml.

```
2.2.1 :034 > ace = '10 permit ip 4.4.4.4 255.255.255.0 8.8.8.8 255.255.255.0 fragments'
 => "10 permit ip 4.4.4.4 255.255.255.0 8.8.8.8 255.255.255.0 fragments" 
2.2.1 :035 > /10 (\S+) (\S+) (.*)/.match(ace)
 => #<MatchData "10 permit ip 4.4.4.4 255.255.255.0 8.8.8.8 255.255.255.0 fragments" 1:"permit" 2:"ip" 3:"4.4.4.4 255.255.255.0 8.8.8.8 255.255.255.0 fragments"> 
```

lib/cisco_node_utils/cmd_ref/acl_v4.yaml
```diff
@@ -0,0 +1,25 @@
...
+# For documentation please see:
+#   - README_YAML.md
+#
+---
+_template:
+  config_get: 'show run aclmgr'
+  config_get_token: '/^ip access-list <acl_name>$/'
+  config_set: 'ip access-list <acl_name>'
+
+ace:
```
mikewiebe, line 12: Move this into a separate `ace.yaml` file

lib/cisco_node_utils/cmd_ref/acl_v4.yaml
```diff
@@ -0,0 +1,25 @@
...
+  config_get_token_append: '/\s+(<seqno>)\s+(permit|deny)\s+(sctp|tcp|ip|udp|pim|icmp|igmp|gre|nos|pcp|ospf|eigrp|ahp|esp|\d+)\s+(any|host \S+|\S+\/\d+|\S+ [:\.0-9a-fA-F]+)\s+(eq \d+|neq \d+|lt \d+|gt \d+|range \d+ \d+)*\s*(any|host \S+|\S+\/\d+|\S+ [:\.0-9a-fA-F]+)\s*(eq \d+|neq \d+|lt \d+|gt \d+|range \d+ \d+)*( [a-zA-Z0-9\/ ]*)*$'
+  config_set_append: 
+      '<state> <seqno> <action> <proto> <v4_src_addr_format> <v4_src_port_format> <v4_dst_addr_format> <v4_dst_port_format> <option_format>'
+
+acl:
```
mikewiebe, line 17: Change this to

```
config_get_token: '/^<afi> access-list (\S+)$/'
config_set: '<state> <afi> access-list <acl_name>'
```

mikewiebe:
Please merge the latest develop into your fork to fix the rubocop errors.

lib/cisco_node_utils/acl.rb
yjyongz, line 18: i will change this comment

lib/cisco_node_utils/acl.rb
yjyongz, line 25: accept

lib/cisco_node_utils/ace_v4.rb
yjyongz, line 18: accept

lib/cisco_node_utils/cmd_ref/acl_v4.yaml
chrisvanheuveln, line 9: @mikewiebe I'm not sure that will work right if the nvgen string is `ip` instead of `ipv4`

chrisvanheuveln:
@saqibraza Yes it should. This needs a yaml entry for something like `all_aces:`. We have a number of examples that do this (`all_afs`, `all_neighbors`, etc) that are collected and added to a hash with the proper context to keep them separate.

lib/cisco_node_utils/ace_v4.rb
yjyongz, line 23: changed

lib/cisco_node_utils/cmd_ref/acl_v4.yaml
mikewiebe, line 9: @chrisvanheuveln As we discussed it would work because in the constructor we are passing `ipv4` or `ipv6` so in the `ipv4` case we simply use `'ip'`  when afi == 'ipv4' as the `@get_args` value for `:afi`

chrisvanheuveln, line 9: @mikewiebe cool. XR actually uses `ipv4` for this but we'll just have a separate yaml entry for `cli_ios_xr:` so we're good.

tests/test_ace_v4.rb
yjyongz, line 52: we support both

lib/cisco_node_utils/cmd_ref/acl_v4.yaml
yjyongz, line 12: done

lib/cisco_node_utils/cmd_ref/acl_v4.yaml
yjyongz, line 17: done

lib/cisco_node_utils/cmd_ref/acl_v4.yaml
yjyongz, line 13: Hi Mike,

i decided to keep regular expression like this
 config_get_token_append: '/^(<seqno>)\s+(\S+)\s+(\S+|\d+)\s+(any|host \S+|\S+\/\d+|\S+ [:.0-9a-fA-F]+)\s+(eq \S+|neq \S+|lt \S+|gt \S+|range \S+ \S+)_\s_(any|host \S+|\S+\/\d+|\S+ [:.0-9a-fA-F]+)\s_(eq \S+|neq \S+|lt \S+|gt \S+|range \S+ \S+)_( [a-zA-Z0-9\/ ]_)_$/'

because the getter function of each property need to return match[Integer], change regular expression makes hard to pin point exact location.

tests/test_ace_v4.rb
yjyongz, line 53: Hi Chris, 

can we do like this?
src_port => 'eq 40'
dst_port => 'neq 56'

chrisvanheuveln, line 53: I think you misunderstood the issue on this one - it's not the name of the property that's the question, it's whether `eq` and `neq` strings are standard syntax across platforms and yang.

PR #101, opened 2015-11-23
Feature/grpc rd

RichWellum:
This is a merge of feature/grpc-rd to feature/grpc.

grpc-rd is a merger branch with develop added and route_distinguisher updated to work for IOS XR.

Minitests all pass (some cleanup added). Rubocop happy.

Equivalent merge for puppet side coming soon.


lib/cisco_node_utils/bgp.rb
```diff
@@ -531,6 +527,35 @@ def default_reconnect_interval
       config_get_default('bgp', 'reconnect_interval')
     end
 
+    # Route Distinguisher (Getter/Setter/Default)
```
chrisvanheuveln, line 530: This looks like it's mostly the same code I reviewed for develop. If this is here because of a merge then it would be better if you just committed the merged code first and then submitted the PR for your actual changes (ie. close this one, commit the merge, open new PR).

lib/cisco_node_utils/bgp.rb
```diff
@@ -30,12 +31,7 @@ def initialize(asnum, vrf='default', instantiate=true)
       if @vrf == 'default'
         @get_args = @set_args = { asnum: @asnum }
       else
-        if platform == :ios_xr
-          @rd = 'auto'
-          @get_args = @set_args = { asnum: @asnum, vrf: @vrf, rd: @rd }
-        else
-          @get_args = @set_args = { asnum: @asnum, vrf: @vrf }
-        end
+        @get_args = @set_args = { asnum: @asnum, vrf: @vrf }
```
RichWellum, line 34: Removed special handling for IOS XR.

lib/cisco_node_utils/bgp.rb
```diff
@@ -531,6 +527,35 @@ def default_reconnect_interval
       config_get_default('bgp', 'reconnect_interval')
     end
 
+    # Route Distinguisher (Getter/Setter/Default)
+    # Configure in vrf context
+    def route_distinguisher
+      if platform == :nexus
```
RichWellum, line 533: nexus only

lib/cisco_node_utils/bgp.rb
```diff
@@ -531,6 +527,35 @@ def default_reconnect_interval
...
+        return false unless RouterBgpAF.feature_nv_overlay_evpn_enabled
+      end
+      config_get('bgp', 'route_distinguisher', @get_args)
+    end
+
+    def route_distinguisher=(rd)
+      if platform == :nexus
```
RichWellum, line 540: nexus only

lib/cisco_node_utils/cmd_ref/bgp.yaml
```diff
@@ -219,6 +219,15 @@ reconnect_interval:
     config_set_append: '<state> reconnect-interval <seconds>'
     default_value: 60
 
+route_distinguisher:
+  default_value: ""
+  cli_ios_xr:
+    config_get_token_append: '/^rd (\S+)$/'
```
RichWellum, line 225: just added xr specific yaml.

lib/cisco_node_utils/cmd_ref/bgp.yaml
```diff
@@ -286,6 +295,3 @@ vrf:
   multiple: true
   config_get_token_append: '/^vrf\s+(\S+)$/'
   config_set: ["router bgp <asnum>", "<state> vrf <vrf>"]
-  cli_ios_xr:
```
RichWellum, line 289: Removed special rd handling in yaml/vrf.

lib/cisco_node_utils/bgp.rb
RichWellum, line 530: Sorry - you are correct that would have been better.

To mitigate this I've gone through and marked my changes. Only exception is the test_ file which has a lot of changes - but mainly cleanup and fluff.

Is that ok?

lib/cisco_node_utils/cmd_ref/bgp.yaml
```diff
@@ -219,6 +219,15 @@ reconnect_interval:
     config_set_append: '<state> reconnect-interval <seconds>'
     default_value: 60
 
+route_distinguisher:
+  default_value: ""
+  cli_ios_xr:
+    config_get_token_append: '/^rd (\S+)$/'
+    config_set_append: ' <state> rd <rd>'
```
chrisvanheuveln, line 226: Is the whitespace indention still needed?

lib/cisco_node_utils/bgp.rb
chrisvanheuveln, line 530: Looks like 90% of this is from the merge? This is a pain looking through code that hasn't actually changed.

tests/test_router_bgp.rb
```diff
@@ -43,12 +52,14 @@ def create_bgp_vrf(asnum, vrf)
 end
 
 def setup_default
+  @device.cmd('no router bgp')
```
chrisvanheuveln: Does this actually work? IOS & NX require the asn to remove it.

tests/test_router_bgp.rb
```diff
@@ -22,6 +22,15 @@
 
 XR_NO_VRF_SUPPORT = 'Not supported in a BGP VRF'
 
+# Not a replacement for cisco debugging
```
chrisvanheuveln:
```
# Temporary debug helper. Not for production and not a replacement for cisco debugging.
```

tests/test_router_bgp.rb
```diff
@@ -226,11 +234,10 @@ def test_get_asnum
   end
 
   def test_destroy
```
chrisvanheuveln: I would remove this test and rename the test at line 196 'test_destroy'

tests/test_router_bgp.rb
```diff
@@ -639,19 +637,16 @@ def log_neighbor_changes(bgp)
...
-    bgp.destroy
-  end
-
   def test_default_log_neighbor_changes
-    asnum = 55
-    bgp = RouterBgp.new(asnum)
-    refute(bgp.default_log_neighbor_changes,
-           'bgp log_neighbor_changes default value should be false')
+    bgp = setup_default
+    if platform == :ios_xr
```
chrisvanheuveln: This test should be based on the yaml values instead of hard-coded to platform. I think just change this:

```
/if platform == :ios_xr/if bgp.default_log_neighbor_changes/
```

_(assuming all platforms support this feature)_

tests/test_router_bgp.rb
```diff
@@ -733,22 +725,42 @@ def test_set_get_reconnect_interval
...
       if t == 'test_default'
         bgp = setup_default
       else
         bgp = setup_vrf
       end
-      bgp.router_id = '1.2.3.4'
-      assert_equal('1.2.3.4', bgp.router_id,
+      bgp.router_id = '7.8.9.11'
+      assert_equal('7.8.9.11', bgp.router_id,
                    "vrf #{@vrf}: bgp router_id should be set to '1.2.3.4'")
```
chrisvanheuveln: You changed the rid everywhere but in the error message.

IMO you don't need to state it in the message at all since the assert is going to display <expected>'7.8.9.11' if it fails; ie. this message tells you everything you need:

```
"vrf #{@vrf}: router_id invalid")
```

lib/cisco_node_utils/cmd_ref/bgp.yaml
RichWellum, line 226: For now yes. They have changed their API to be able to do a 'service cli noindent' or similar - but we need to have a discussion about how to do this in our code. We have a US to clean this all up btw.

tests/test_router_bgp.rb

tests/test_router_bgp.rb
RichWellum: Yes it works.

RP/0/RP0/CPU0:ios#sh run router bgp
Mon Nov 23 21:46:06.098 UTC
router bgp 55
 bgp confederation peers
  11
  12
  13
 !
 bgp router-id 1.1.1.1
 bgp log neighbor changes disable
 vrf blue
  rd auto
  bgp router-id 192.168.0.66
 !
!

RP/0/RP0/CPU0:ios#conf
Mon Nov 23 21:46:09.679 UTC

RP/0/RP0/CPU0:ios(config)#no router bgp
RP/0/RP0/CPU0:ios(config)#commit
Mon Nov 23 21:46:17.583 UTC
RP/0/RP0/CPU0:ios(config)#sh run router bgp
Mon Nov 23 21:46:33.733 UTC
% No such configuration item(s)

tests/test_router_bgp.rb

tests/test_router_bgp.rb
RichWellum: Problem is the default behavior is different between Nexus and XR. XR default is on (but not nvgenned), Nexus is off. So I have to differentiate here.

tests/test_router_bgp.rb

tests/test_router_bgp.rb
RichWellum: @chrisvanheuveln Addressed comments - except the one about not using platform check in the test file.

tests/test_router_bgp.rb
chrisvanheuveln: Exactly. That's why I want you to check the default value in order to know whether to assert or refute. If default is true, assert, else, refute.

tests/test_router_bgp.rb
chrisvanheuveln: Okay, it works on XR but it's still going to fail on NX...?

tests/test_router_bgp.rb
RichWellum: Ack IC. I'll give it a go.

tests/test_router_bgp.rb
RichWellum: Actually I shouldn't need it at all - because:

  def setup
    # Disable feature bgp before each test to ensure we
    # are starting with a clean slate for each test.
    super
    if platform == :ios_xr
      config('no router bgp')
    else
      config('no feature bgp')
    end
  end

Will retest to confirm.

lib/cisco_node_utils/bgp.rb
RichWellum, line 34: @chrisvanheuveln Addressed your last two comments, new code pushed.

PR #102, opened 2015-11-23
Collapse feature/fabricpath into develop

mikewiebe:
1) Minitests pass with logic to skip fabricpath tests when not supported.
2) Rubocop passes.


mikewiebe:
@chrisvanheuveln , @dcheriancisco NodeUtils PR for the feature/fabricpath collapse to develop.

CHANGELOG.md
```diff
@@ -3,9 +3,15 @@ Changelog
...
 
+### New feature support
+* Fabric Path
+  * fabricpath_global (@dcheriancisco)
+  * fabricpath_topology (@dcheriancisco)
+
 ### Added
 
 * `Cisco::UnsupportedError` exception class, raised when a command is explicitly marked as unsupported on a particular class of nodes.
+* Extensions to interface and vlan providers to support fabric path. (dcheriancisco)
```
chrisvanheuveln, line 17: `(@dcheriancisco)`

lib/cisco_node_utils/cmd_ref/fabricpath.yaml
```diff
@@ -0,0 +1,178 @@
...
+    - /N5K/
+  kind: boolean
+  config_get: "show run fabricpath"
+  config_get_token: '/^fabricpath multicast aggregate\-routes$/'
+  config_set: "<state> fabricpath multicast aggregate-routes"
+  default_value: false
+  
+allocate_delay:
+  kind: int
+  config_get: "show fabricpath timers"
```
chrisvanheuveln, line 17: I'm hesitant to use feature commands instead of show run config. Is this the only way to get the state? Looks like most if not all can be found there:

```
# sh run fabricpath all
install feature-set fabricpath
feature-set fabricpath

fabricpath domain default
  graceful-restart
  hostname dynamic
  log-adjacency-changes
  authentication-check
fabricpath timers allocate-delay 10
fabricpath timers linkup-delay 10
fabricpath timers transition-delay 10
fabricpath load-balance source-destination 
fabricpath load-balance unicast mixed rotate-amount 0xa include-vlan 
```

Ditto for most of these commands.

lib/cisco_node_utils/cmd_ref/fabricpath.yaml
```diff
@@ -0,0 +1,178 @@
...
+---
+# Fabricpath feature is not available on N3K and N9K
+_exclude: [/N3/, /N9K/]
+
+aggregate_multicast_routes:
+  _exclude:
+    - /N5K/
+  kind: boolean
+  config_get: "show run fabricpath"
+  config_get_token: '/^fabricpath multicast aggregate\-routes$/'
```
chrisvanheuveln, line 11: nit: `/aggregate\-routes/aggregate.routes/`

lib/cisco_node_utils/cmd_ref/fabricpath.yaml
```diff
@@ -0,0 +1,178 @@
...
+  default_value: 10
+
+auto_switch_id: 
+  kind: int
+  config_get: "show fabricpath switch-id"
+  config_get_token: '/^\*\s+(\d+).*No$/'
+
+feature:
+  config_get: "show feature-set"
+  config_get_token: '/^fabricpath[\s]+[\d]+[\s]+(\S+)/'
```
chrisvanheuveln, line 29: `show run | section feature-set` makes for a cleaner config_get_token regexp

lib/cisco_node_utils/cmd_ref/fabricpath.yaml
```diff
@@ -0,0 +1,178 @@
...
+  kind: int
+  config_get: "show fabricpath timers"
+  config_get_token: '/^Allocate Delay Timer\s+:\s+(\d+)/'
+  config_set: "<state> fabricpath timers allocate-delay <delay>"
+  default_value: 10
+
+auto_switch_id: 
+  kind: int
+  config_get: "show fabricpath switch-id"
+  config_get_token: '/^\*\s+(\d+).*No$/'
```
chrisvanheuveln, line 25: `auto_switch_id` doesn't seem to be configurable and I see it's only used as a getter, but I'm wondering if it's really needed at all? If the cli `switch_id` config isn't used do we actually care about auto_switch_id?

lib/cisco_node_utils/cmd_ref/fabricpath.yaml
```diff
@@ -0,0 +1,178 @@
...
+  config_set: "<state> feature-set fabricpath"
+
+feature_install:
+  config_set: "<state> install feature-set fabricpath"
+
+graceful_merge:
+  auto_default: false
+  kind: boolean
+  config_get: "show run fabricpath"
+  config_get_token: '/^fabricpath graceful\-merge disable$/'
```
chrisvanheuveln, line 39: nit: `/graceful.merge/`

lib/cisco_node_utils/fabricpath_global.rb
```diff
@@ -0,0 +1,439 @@
...
+    def initialize(name, instantiate=true)
+      fail TypeError unless name.is_a?(String)
+      fail ArgumentError unless name == 'default'
+      @name = name.downcase
+      @set_params = {}
+
+      create if instantiate
+    end
+
+    def self.globals
```
mikewiebe, line 36: `self.globals`, `self.fabricpath_feature`, `self.fabricpath_feature_set`, `create` and `destroy` methods need to be refactored similar to vxlan_vtep.

https://github.com/cisco/cisco-network-node-utils/blob/feature/vxlan/lib/cisco_node_utils/vxlan_vtep.rb#L35

lib/cisco_node_utils/cmd_ref/fabricpath.yaml
```diff
@@ -0,0 +1,178 @@
...
+  config_set: "<state> fabricpath graceful-merge disable"
+  default_value: true
```
chrisvanheuveln, line 41: IMO this prop should either be renamed `graceful_merge_disable` or the default_value should be false...

Note that the getter returns true when it matches 'disable'... but the graceful-merge execution state is actually false in that case. 

I'm looking at this from the view of graceful-merge state, not grace-merge disable state.

```
(config)# sh run fabricpath all | i graceful-merge
(config)#
(config)# fabricpath graceful-merge disable
(config)# sh run fabricpath all | i graceful-merge
fabricpath graceful-merge disable

(config)# no fabricpath graceful-merge disable
(config)# sh run fabricpath all | i graceful-merge
(config)# 
```

lib/cisco_node_utils/cmd_ref/fabricpath_topology.yaml
```diff
@@ -0,0 +1,35 @@
...
+# Fabricpath feature is not available on N3K and N9K
+_exclude: [/N3/, /N9K/]
+
+all_topos:
+  multiple:
+  config_get: "show fabricpath topology"
+  config_get_token: '/^\S+[\s]+(\d+)[\s]+/'
+
+create:
+  config_set: ["fabricpath topology <topo>" , "end"]
```
chrisvanheuveln, line 12: `/ , "end"//`

Actually, shouldn't this be a template instead? 

```
_template:
  config_get: 'show running fabricpath topology all'
  config_get_token: '/^fabricpath topology <topo>$/'
  config_set: 'fabricpath topology <topo>'
```

lib/cisco_node_utils/cmd_ref/fabricpath_topology.yaml
```diff
@@ -0,0 +1,35 @@
+# fabricpath_topology
+---
+# Fabricpath feature is not available on N3K and N9K
+_exclude: [/N3/, /N9K/]
+
+all_topos:
+  multiple:
+  config_get: "show fabricpath topology"
```
chrisvanheuveln, line 8: Use show runn, not show fabricpath topo...

```
# sh run fabricpath topology all
fabricpath topology 2
  description foo
  member vlan 3 
```

lib/cisco_node_utils/cmd_ref/fabricpath_topology.yaml
```diff
@@ -0,0 +1,35 @@
...
+  config_get_token: '/^\S+[\s]+(\d+)[\s]+/'
+
+create:
+  config_set: ["fabricpath topology <topo>" , "end"]
+
+description:
+  kind: string
+  config_get: "show fabricpath topology"
+  config_get_token: '/^(\S+)[\s]+%d+[\s]+\S+/'
+  config_set: ["fabricpath topology <topo>" , "<state> description <name>", "end"]
```
chrisvanheuveln, line 18: Based on earlier comments, all of these properties need to be reworked to use the _template and updated regex's to use show run output.

lib/cisco_node_utils/fabricpath_global.rb
chrisvanheuveln, line 36: In addition -- considering that fabricpath is installed as a feature-set in a two-step process -- take a look at feature-set fex enablement to see how we've handled that:  
https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/interface.rb#L127

lib/cisco_node_utils/fabricpath_global.rb
```diff
@@ -0,0 +1,439 @@
...
+      FabricpathGlobal.fabricpath_feature_set(:enabled) unless
+      :enabled == FabricpathGlobal.fabricpath_feature
+    end
+
+    def destroy
+      @name = nil
+      FabricpathGlobal.fabricpath_feature_set(:disabled)
+    end
+
+    def my_munge(property, set_val)
```
chrisvanheuveln, line 84: We need some comments to explain why this method is needed.

lib/cisco_node_utils/fabricpath_global.rb
```diff
@@ -0,0 +1,439 @@
...
+    ########################################################
+
+    def aggregate_multicast_routes
+      config_get('fabricpath', 'aggregate_multicast_routes')
+    end
+
+    def aggregate_multicast_routes=(val)
+      @set_params = {}
+      @set_params[:state] = val ? '' : 'no'
+      config_set('fabricpath', 'aggregate_multicast_routes', @set_params)
```
chrisvanheuveln, line 116: There's no need for @set_params since you're only using it for a single key (or two in some cases). Just do:

```
config_set('fabricpath', 'aggregate_multicast_routes', state: val ? '' : 'no')
```

lib/cisco_node_utils/fabricpath_global.rb
dcheriancisco, line 36: In't the code in  def self.fabricpath_feature_set(fabricpath_set) identical to that?

lib/cisco_node_utils/fabricpath_global.rb
```diff
@@ -0,0 +1,439 @@
...
+
+    def allocate_delay
+      config_get('fabricpath', 'allocate_delay')
+    end
+
+    def allocate_delay=(val)
+      @set_params = {}
+      @set_params[:state] = val ? '' : 'no'
+      @set_params[:delay] = val ? val : ''
+      config_set('fabricpath', 'allocate_delay', @set_params)
```
chrisvanheuveln, line 133:
```
state = val ? '' : 'no'
delay = val ? val : ''
config_set('fabricpath', 'allocate_delay', state: state, delay: delay)
```

lib/cisco_node_utils/cmd_ref/fabricpath_topology.yaml
dcheriancisco, line 8: The show runn cannot retrieve the 'default topology' membership details (which is what will be existing in most customer configs), nor the default description.

lib/cisco_node_utils/cmd_ref/fabricpath.yaml
dcheriancisco, line 41: The negating part is handled in the getter/setter code:

```
  graceful_merge_conf = config_get('fabricpath', 'graceful_merge')
  # opposite meaning with the cli
  return :enable if graceful_merge_conf.nil?
  graceful_merge_conf ? :disable : :enable
```

lib/cisco_node_utils/fabricpath_global.rb
chrisvanheuveln, line 36: It is - sorry, please ignore my comment.

lib/cisco_node_utils/cmd_ref/fabricpath.yaml
chrisvanheuveln, line 41: ok, then yaml is fine.

lib/cisco_node_utils/cmd_ref/fabricpath_topology.yaml
chrisvanheuveln, line 8: I don't see that as a problem. If the configuration fails to match the regex then the property is not configured. If it's not configured then it is using the default configuration.

lib/cisco_node_utils/fabricpath_global.rb
```diff
@@ -0,0 +1,439 @@
...
+
+    def auto_switch_id
+      config_get('fabricpath', 'auto_switch_id')
+    end
+
+    def graceful_merge
+      graceful_merge_conf = config_get('fabricpath', 'graceful_merge')
+      # opposite meaning with the cli
+      return true if graceful_merge_conf.nil?
+      graceful_merge_conf ? false : true
```
chrisvanheuveln, line 150: Just replace entire method with:

```
def graceful_merge
  # invert result because graceful-merge nvgens with 'disable' keyword
  not config_get('fabricpath', 'graceful_merge')
end
```

lib/cisco_node_utils/fabricpath_topology.rb
```diff
@@ -0,0 +1,150 @@
...
+require_relative 'node_util'
+require_relative 'fabricpath_global'
+
+module Cisco
+  # node_utils class for fabricpath_topology
+  class FabricpathTopo < NodeUtil
+    attr_reader :topo_id
+
+    def initialize(topo_id, instantiate=true)
+      @topo_id = topo_id.to_s
```
chrisvanheuveln, line 28: Why the to_s if you want it to be an integer? Consider:

```
@topo_id = topo_id.to_i
fail ArgumentError unless @topo_id > 0
```

You also won't need the regex check if you use that.

lib/cisco_node_utils/fabricpath_global.rb
chrisvanheuveln, line 36: I remember now why I didn't like this... it's because the feature checker returns four possible symbol states (as the fex feature does as well). 
It would have been better to just have it return true/false. If it's not both installed and enabled then it's false. All the extra symbol checking just complicates things and it's not going to hurt to re-enter the install command if it's already installed.
Anyway, what you have is fine but feel free to rework it. :-)

lib/cisco_node_utils/fabricpath_topology.rb
```diff
@@ -0,0 +1,150 @@
...
+      topo_list.each do |id|
+        hash[id] = FabricpathTopo.new(id, false)
+      end
+      hash
+    end
+
+    def create
+      fabricpath_feature_set(:enabled) unless :enabled == fabricpath_feature
+      config_set('fabricpath_topology', 'create',
+                 topo: @topo_id) unless @topo_id == '0'
```
chrisvanheuveln, line 52: Shouldn't need the zero check if we catch it at initialize. See prev comment.

lib/cisco_node_utils/fabricpath_topology.rb
```diff
@@ -0,0 +1,150 @@
...
+
+    def fabricpath_feature
+      FabricpathGlobal.fabricpath_feature
+    end
+
+    def fabricpath_feature_set(fabricpath_set)
+      FabricpathGlobal.fabricpath_feature_set(fabricpath_set)
+    end
+
+    def state
```
chrisvanheuveln, line 75: It's not clear to me why this method is needed. There's no setter and there don't appear to be any callers or related configs?

lib/cisco_node_utils/fabricpath_topology.rb
```diff
@@ -0,0 +1,150 @@
...
+    end
+
+    def member_vlans
+      str = config_get('fabricpath_topology', 'member_vlans', @topo_id)
+      return [] if str == '--'
+      str.gsub!('-', '..')
+      if /,/.match(str)
+        str.split(/\s*,\s*/)
+      else
+        str.lines.to_a
```
chrisvanheuveln, line 97: Do you need this else clause? The split will still give an array if str is a non-comma string.

lib/cisco_node_utils/fabricpath_topology.rb
```diff
@@ -0,0 +1,150 @@
...
+      if desc.empty?
+        @set_params[:topo] = @topo_id
+        @set_params[:state] = 'no'
+        @set_params[:name] = ''
+      else
+        @set_params[:topo] = @topo_id
+        @set_params[:state] = ''
+        @set_params[:name] = desc
+      end
+      config_set('fabricpath_topology', 'description', @set_params)
```
chrisvanheuveln, line 141: Why is the setter/getter called 'topo_name' if the property is 'description' ?  The agents / manifest should be expecting 'description'.

lib/cisco_node_utils/interface.rb
```diff
@@ -96,6 +97,22 @@ def default_description
       config_get_default('interface', 'description')
     end
 
+    def enable_pim_sparse_mode
+      state = config_get('interface', 'pim_sparse_mode', @name)
+      state ? true : false
+    end
```
chrisvanheuveln, line 103: 1. This should be `pim_sparse_mode` or possibly `ip_pim_sparse_mode`
2. You don't need the state var.

e.g.

```
def pim_sparse_mode
  config_get('interface', 'pim_sparse_mode', @name)
end
```

lib/cisco_node_utils/interface.rb
```diff
@@ -114,6 +131,31 @@ def default_encapsulation_dot1q
       config_get_default('interface', 'encapsulation_dot1q')
     end
 
+    def encapsulation_profile_vni
+      val = config_get('interface', 'encapsulation_vni', @name)
+      debug "val from get is #{val}"
+      return '' if val.nil?
+      val.first.strip
+    end
```
chrisvanheuveln, line 139: 1. I don't see any yaml entries for `encapsulation_vni` in this diff or in the branch?
2. debug in the wild (use CiscoLogger.debug)
3. Should be able to just call config_get() by itself. The return and strip can be removed.

lib/cisco_node_utils/vlan.rb
```diff
@@ -60,6 +61,45 @@ def cli_error_check(result)
...
 
+    def fabricpath_feature
+      FabricpathGlobal.fabricpath_feature
+    end
+
+    def fabricpath_feature_set(fabricpath_set)
+      FabricpathGlobal.fabricpath_feature_set(fabricpath_set)
+    end
+
+    def mode
```
chrisvanheuveln, line 72: This needs a feature enabled check.

tests/test_fabricpath_topology.rb
```diff
@@ -0,0 +1,98 @@
...
+  def teardown
+    # teardown runs at the end of each test
+    no_feature_fabricpath
+    super
+  end
+
+  def no_feature_fabricpath
+    # Turn the feature off for a clean test.
+    config('no feature-set fabricpath')
+    config('no install feature-set fabricpath')
```
chrisvanheuveln, line 40: Do you actually need to do a 'no install' every time? It's somewhat time-consuming to do it for every test case. I would think the 'no feature-set' by itself would be sufficient. Ditto for the test_fabricpath_global file.

lib/cisco_node_utils/vlan.rb
```diff
@@ -60,6 +61,45 @@ def cli_error_check(result)
...
+      result = config_get('vlan', 'mode', @vlan_id)
+      return default_mode if result.nil?
+      case result
+      when /fabricpath/i
+        return 'fabricpath'
+      when /ce/i
+        return 'ce'
+      end
```
chrisvanheuveln, line 80: We shouldn't need this regex's directly in the code if we can handle it in the yaml. (I'm harping on the 'show run' again)

lib/cisco_node_utils/vlan.rb
```diff
@@ -60,6 +61,45 @@ def cli_error_check(result)
...
+    end
+
+    def mode=(str)
+      str = str.to_s
+      if str.empty?
+        result = config_set('vlan', 'mode', @vlan_id, 'no', '')
+      else
+        if 'fabricpath' == str
+          fabricpath_feature_set(:enabled) unless
```
chrisvanheuveln, line 89: This feature check should occur at the beginning of the method; ie. if fabricpath is not enabled then the 'no' command above will raise a CliError.

lib/cisco_node_utils/interface.rb
```diff
@@ -96,6 +97,22 @@ def default_description
...
+      state = config_get('interface', 'pim_sparse_mode', @name)
+      state ? true : false
+    end
+
+    def enable_pim_sparse_mode=(state)
+      no_cmd = (state ? '' : 'no')
+      config_set('interface', 'pim_sparse_mode', @name, no_cmd)
+    rescue Cisco::CliError => e
+      raise "[#{@name}] '#{e.command}' : #{e.clierror}"
+    end
```
chrisvanheuveln, line 110: There needs to be a feature pim check:
e.g.

```
def pim_sparse_mode=(state)
  return false unless Pim.feature_enabled
   ...
end
```

Smitha's working on the Pim.feature_enabled part. We can either dupe it here temporarily or wait for her code.

mikewiebe:
@dcheriancisco After discussions with @chrisvanheuveln we decided to go ahead and merge this branch and open a new US to address the refactoring comments so that you can complete the development of the puppet providers.  The impact to puppet should be minimal after the refactoring.

PR #107, opened 2015-11-25
non-vrf handing for XR

RichWellum:
Using Glenn's preferred method to deal with bgp global properties that are not supported at a VRF level. Associated minitest changes.

Couple of fix ups to other properties I was skipping in the minitests instead of reasing unsupported.

Rubocop happy.

Minitests:
XR:
63 runs, 204 assertions, 0 failures, 0 errors, 9 skips

Nexus:
63 runs, 242 assertions, 0 failures, 0 errors, 0 skips

### DIFF
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -266,6 +266,7 @@ module Cisco
 
     # Cluster Id (Getter/Setter/Default)
     def cluster_id
+      return nil if platform == :ios_xr && @vrf != 'default'
       config_get('bgp', 'cluster_id', @get_args)
     end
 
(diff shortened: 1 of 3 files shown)


lib/cisco_node_utils/bgp.rb
```diff
@@ -524,6 +582,7 @@ def reconnect_interval=(seconds)
     end
 
     def default_reconnect_interval
+      return nil if platform == :ios_xr
```
ceburkett: Doesn't config_get_default already return nil if the yaml doesn't specify XR config?

RichWellum: Looks like it from the readme:

When a feature or attribute is excluded in this way, attempting to call `config_get` or `config_set` on an excluded node will result in a `Cisco::UnsupportedError` being raised. Calling `config_get_default` on such a node will always return `nil`.

I'll remove it and retest.

RichWellum: Yup that worked. I'll update in a sec.

chburket: :+1:

PR #109, opened 2015-12-01
Minor cleanups for cisco_interface and yaml

chrisvanheuveln:
test_interface.rb minitest passes on n3k, n31, n9k, n9kv


lib/cisco_node_utils/cmd_ref/interface.yaml
```diff
@@ -24,8 +24,8 @@ all_interfaces:
 
 channel_group:
```
RichWellum, line 25: :+1:

PR #110, opened 2015-12-02
CMD_ref, Node_util and minitest files for cisco_vni- specific to 9k/3k.

alok-aggarwal:
Please review 9k specific code in vni.rb and yaml files. The 7k code still needs to be refactored. 

Here are the methods to be reviewed:
feature
feature_set
Create
mapped_vlan
mapped_vlan=(vlan)
default_vlan

Look for this condition 
 if /N9K/.match(node.product_id)

No Rubocop errors.
Minitest passes:
# Running:

Node under test:
- name  - agent-lab1-nx
- type  - N9K-NXOSV

TestVni#test_on_off = 6.48 s = .
TestVni#test_multiple_vnis_vlans = 13.08 s = .
TestVni#test_mapped_vlan = 5.35 s = .

Finished in 24.904488s, 0.1205 runs/s, 0.4015 assertions/s.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/vni.yaml
+++ b/lib/cisco_node_utils/cmd_ref/vni.yaml
@@ -1,0 +1,51 @@
+# vni
+---
+all_vnis:
+  config_get: "show vni"
+  config_get_token: '/^(\d+)\s/'
+
+bridge_domain:
+  config_get: "show vni"
+  config_get_token: [ '/^%d\s+\S+\s+(\d+)/' ]
+  config_set: ["bridge-domain %d", "%s member vni %d", "end"]
+  default_value: ~
+
+bridge_domain_activate:
+  config_set: ["%s system bridge-domain add %d", "end"]
+
+create:
+  config_set: ["vni %s" , "end"]
+
+destroy:
+  config_set: "no vni %s"
+
+encap_dot1q:
+  config_set: ["encapsulation profile vni %s", "%s dot1q %s vni %s", "end"]
+  default_value: ~
+
+feature:
+  config_get: "show feature"
+  config_get_token: '/^vni\s+\d+\s+(\S+)/'
+  config_set: "%s feature vni"
+
+feature_n9k:
+  config_get: " show running | i ^feature"
+  config_get_token: '/^feature vn-segment-vlan-based$/'
+  config_set: "<state> feature vn-segment-vlan-based"
+
+feature_nv_overlay:
+  config_get: " show running | i ^feature"
+  config_get_token: '/^feature nv overlay$/'
+  config_set: "<state> feature nv overlay"
+
+mapped_vlan:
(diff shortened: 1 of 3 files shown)


lib/cisco_node_utils/cmd_ref/vni.yaml
```diff
@@ -0,0 +1,51 @@
+# vni
+---
+all_vnis:
+  config_get: "show vni"
+  config_get_token: '/^(\d+)\s/'
+
+bridge_domain:
```
chrisvanheuveln, line 7: Are bridge-domains exclusively used with vni? 
If so then fine; otherwise I think we should move this out into it's own bridge_domain.yaml file. 

But you don't seem to use this in the vni code anyway?

lib/cisco_node_utils/cmd_ref/vni.yaml
```diff
@@ -0,0 +1,51 @@
...
+  config_get: "show vni"
```
chrisvanheuveln, line 8: Most of the time it's better to get state from running-config instead of exec commands. Is there a reason why you can't get it from running-config?

lib/cisco_node_utils/cmd_ref/vni.yaml
```diff
@@ -0,0 +1,51 @@
...
+  config_get_token: [ '/^%d\s+\S+\s+(\d+)/' ]
+  config_set: ["bridge-domain %d", "%s member vni %d", "end"]
```
chrisvanheuveln, line 10: 1. config_set() does the "end" for you so you don't need to specify it here.
2. Please use key-values instead of printf. 
   https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/cmd_ref/README_YAML.md#key-value-wildcards

lib/cisco_node_utils/cmd_ref/vni.yaml
alok-aggarwal, line 7: The 7k code still needs to be refactored as part of US52662. I have made a note of these comments. They will apply when we do US52662.

Here are the methods to be reviewed for now.
feature
feature_set
Create
mapped_vlan
mapped_vlan=(vlan)
default_vlan

Look for this condition 
if /N9K/.match(node.product_id)

lib/cisco_node_utils/cmd_ref/vni.yaml
alok-aggarwal, line 8: Please see above comment.

lib/cisco_node_utils/cmd_ref/vni.yaml
alok-aggarwal, line 10: Please see above comment.

chrisvanheuveln:
Alok will track these comments in US52662 and address them during the refactor.

PR #99, opened 2015-11-21
acl.rb ace.rb acl.yaml test_acl.rb test_ace.rb commit

yjyongz:
[21:54 shenma-laas]($):$ rubocop tests/test_ace.rb 
Inspecting 1 file
.

1 file inspected, no offenses detected
[21:54 shenma-laas]($):$ rubocop lib/cisco_node_utils/acl.rb 
Inspecting 1 file
.

1 file inspected, no offenses detected
[21:54 shenma-laas]($):$ pwd
/tftpboot/shenma/cisco-node/cisco-network-node-utils

[21:54 shenma-laas]($):$ rubocop tests/test_acl.rb 
Inspecting 1 file
.

1 file inspected, no offenses detected
[21:52 shenma-laas]:$ rubocop lib/cisco_node_utils/ace.rb 
Inspecting 1 file
.

1 file inspected, no offenses detected

PPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPP
ruby test_ace.rb -v -- 10.122.197.183 admin admin
Run options: -v -- --seed 9837
# Running:

Node under test:
- name  - agent-lab3-nx
- type  - N9K-NXOSV
- image - bootflash:///nxos.7.0.3.I2.1.bin

TestAceV4#test_router_create_destroy_ipv4_ace_one = 8.41 s = .

Finished in 8.411624s, 0.1189 runs/s, 1.4266 assertions/s.

1 runs, 12 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for MiniTest to /auto/tftp-sjc-users3/shenma/cisco-node/cisco-network-node-utils/tests/coverage. 0.0 / 0.0 LOC (100.0%) covered.
[21:56 shenma-laas]:$ ruby test_acl.rb -v -- 10.122.197.183 admin admin
Run options: -v -- --seed 1073
# Running:

Node under test:
- name  - agent-lab3-nx
- type  - N9K-NXOSV
- image - bootflash:///nxos.7.0.3.I2.1.bin

TestAcl#test_router_stats_enable_v6 = 6.72 s = .
TestAcl#test_router_stats_enable = 4.46 s = .
TestAcl#test_router_create_acl_v6 = 4.05 s = .
TestAcl#test_router_create_acl_v4 = 5.01 s = .

Finished in 20.243165s, 0.1976 runs/s, 1.1856 assertions/s.

4 runs, 24 assertions, 0 failures, 0 errors, 0 skips
[...]

PR #111, opened 2015-12-02
Separating ACL changes from ACE changes

saqibraza:
Do we need to take ace out of acl.yaml?

### DIFF
--- a/lib/cisco_node_utils/acl.rb
+++ b/lib/cisco_node_utils/acl.rb
@@ -23,11 +23,9 @@ module Cisco
     # instantiate: true = create acl instance
     def initialize(acl_name, afi, instantiate=true)
       fail TypeError unless acl_name.is_a?(String)
-      fail ArgumentError 'we expect ip or ipv6' unless
+      fail ArgumentError 'Argument afi must be ip or ipv6' unless
         afi == 'ip' || afi == 'ipv6'
-      @acl_name = acl_name
-      @afi = afi
-      @set_args = @get_args = { acl_name: @acl_name, afi: @afi }
+      @set_args = @get_args = { acl_name: acl_name, afi: afi }
       create if instantiate
     end
 
@@ -67,20 +65,38 @@ module Cisco
     # ----------
     # PROPERTIES
     # ----------
-    # getter stats perentry info
     def stats_per_entry
       config_get('acl', 'stats_per_entry', @get_args)
     end
 
-    # setter stats perentry info
     def stats_per_entry=(state)
       @set_args[:state] = (state ? '' : 'no')
       config_set('acl', 'stats_per_entry', @set_args)
     end
 
-    # default stats perentry info
     def default_stats_per_entry
       config_get_default('acl', 'stats_per_entry')
+    end
+
+    def fragments
+      config_get('acl', 'fragments', @get_args)
+    end
+
+    def fragments=(permit)
+      @set_args[:state] = (permit ? '' : 'no')
+      if permit
+        @set_args[:permit] = permit
+        config_set('acl', 'fragments', @set_args)
+      else
+        @set_args[:permit] = 'permit-all'
+        config_set('acl', 'fragments', @set_args)
+        @set_args[:permit] = 'deny-all'
+        config_set('acl', 'fragments', @set_args)
+      end
+    end
+
+    def default_fragments
+      config_get_default('acl', 'fragments')
     end
 
     # acl == overide func
--- a/lib/cisco_node_utils/acl.rb
+++ b/lib/cisco_node_utils/acl.rb
@@ -1,0 +1,91 @@
+# Copyright (c) 2014-2015 Cisco and/or its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative 'node_util'
+
+module Cisco
+  # Acl - node utility class for ACL configuration
+  class Acl < NodeUtil
+    attr_reader :acl_name, :afi
+
+    # acl_name: name of the acl
+    # instantiate: true = create acl instance
+    def initialize(acl_name, afi, instantiate=true)
+      fail TypeError unless acl_name.is_a?(String)
+      fail ArgumentError 'we expect ip or ipv6' unless
+        afi == 'ip' || afi == 'ipv6'
+      @acl_name = acl_name
+      @afi = afi
+      @set_args = @get_args = { acl_name: @acl_name, afi: @afi }
+      create if instantiate
+    end
+
+    # it will return all acls in the switch
+    def self.acls
+      afis = %w(ip ipv6)
+      acl_hash = {}
+      afis.each do |afi|
+        acl_hash[afi] = {}
+        get_args = { afi: afi }
+        instances = config_get('acl', 'all_acl', get_args)
+
+        next if instances.nil?
+
+        instances.each do |name|
+          acl_hash[afi][name] = Acl.new(name, afi, false)
+        end
+      end
+      acl_hash
+    end
+
+    # config ip access-list and create
+    def create
+      config_acl('')
+    end
+
+    # Destroy acl instance
+    def destroy
+      config_acl('no')
+    end
+
+    def config_acl(state)
+      @set_args[:state] = state
+      config_set('acl', 'acl', @set_args)
+    end
+
+    # ----------
+    # PROPERTIES
+    # ----------
+    # getter stats perentry info
+    def stats_per_entry
+      config_get('acl', 'stats_per_entry', @get_args)
+    end
+
+    # setter stats perentry info
+    def stats_per_entry=(state)
+      @set_args[:state] = (state ? '' : 'no')
+      config_set('acl', 'stats_per_entry', @set_args)
+    end
+
+    # default stats perentry info
+    def default_stats_per_entry
+      config_get_default('acl', 'stats_per_entry')
+    end
+
+    # acl == overide func
+    def ==(other)
+      acl_name == other.acl_name && afi == other.afi
+    end
+  end
+end
--- a/lib/cisco_node_utils/acl.rb
+++ b/lib/cisco_node_utils/acl.rb
@@ -35,8 +35,7 @@ module Cisco
       acl_hash = {}
       afis.each do |afi|
         acl_hash[afi] = {}
-        get_args = { afi: afi }
-        instances = config_get('acl', 'all_acl', get_args)
+        instances = config_get('acl', 'all_acls', afi: afi)
 
         next if instances.nil?
 
@@ -82,17 +81,11 @@ module Cisco
       config_get('acl', 'fragments', @get_args)
     end
 
-    def fragments=(permit)
-      @set_args[:state] = (permit ? '' : 'no')
-      if permit
-        @set_args[:permit] = permit
-        config_set('acl', 'fragments', @set_args)
-      else
-        @set_args[:permit] = 'permit-all'
-        config_set('acl', 'fragments', @set_args)
-        @set_args[:permit] = 'deny-all'
-        config_set('acl', 'fragments', @set_args)
-      end
+    def fragments=(action)
+      @set_args[:state] = (action ? '' : 'no')
+      action = fragments unless action
+      @set_args[:action] = action
+      config_set('acl', 'fragments', @set_args) if action
     end
 
     def default_fragments
--- a/lib/cisco_node_utils/acl.rb
+++ b/lib/cisco_node_utils/acl.rb
@@ -21,11 +21,11 @@ module Cisco
 
     # acl_name: name of the acl
     # instantiate: true = create acl instance
-    def initialize(acl_name, afi, instantiate=true)
+
+    def initialize(afi, acl_name, instantiate=true)
       fail TypeError unless acl_name.is_a?(String)
-      fail ArgumentError 'Argument afi must be ip or ipv6' unless
-        afi == 'ip' || afi == 'ipv6'
-      @set_args = @get_args = { acl_name: acl_name, afi: afi }
+      afi = 'ip' if afi[/ipv4/] # TBD platform-specific
+      @set_args = @get_args = { afi: afi, acl_name: acl_name }
       create if instantiate
     end
 
@@ -39,8 +39,8 @@ module Cisco
 
         next if instances.nil?
 
-        instances.each do |name|
-          acl_hash[afi][name] = Acl.new(name, afi, false)
+        instances.each do |acl_name|
+          acl_hash[afi][acl_name] = Acl.new(afi, acl_name, false)
         end
       end
       acl_hash
--- a/lib/cisco_node_utils/cmd_ref/acl.yaml
+++ b/lib/cisco_node_utils/cmd_ref/acl.yaml
@@ -1,0 +1,33 @@
+# Command Reference Common ACL
+#
+# For documentation please see:
+#   - README_YAML.md
+#
+---
+_template:
+  config_get: 'show run aclmgr'
+  config_get_token: '/^<afi> access-list <acl_name>$/'
+  config_set: '<afi> access-list <acl_name>'
+
+ace:
+  config_get_token_append: '/^(<seqno>) .*$'
+  config_set_append:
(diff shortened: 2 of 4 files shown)


mikewiebe:
Make sure ACL comments from (PR  https://github.com/cisco/cisco-network-node-utils/pull/99) are addressed in this review

lib/cisco_node_utils/acl.rb
```diff
@@ -0,0 +1,91 @@
...
+module Cisco
+  # Acl - node utility class for ACL configuration
+  class Acl < NodeUtil
+    attr_reader :acl_name, :afi
+
+    # acl_name: name of the acl
+    # instantiate: true = create acl instance
+    def initialize(acl_name, afi, instantiate=true)
+      fail TypeError unless acl_name.is_a?(String)
+      fail ArgumentError 'we expect ip or ipv6' unless
```
chrisvanheuveln: Please fix the error msg per review comment in #99. Should be something like:

`fail ArgumentError 'Argument afi must be ip or ipv6' unless...`

lib/cisco_node_utils/acl.rb
```diff
@@ -0,0 +1,91 @@
...
+        afi == 'ip' || afi == 'ipv6'
+      @acl_name = acl_name
+      @afi = afi
+      @set_args = @get_args = { acl_name: @acl_name, afi: @afi }
```
chrisvanheuveln: You don't use these class variables anywhere else in this file, just remove them:

```
-      @acl_name = acl_name
-      @afi = afi
-      @set_args = @get_args = { acl_name: @acl_name, afi: @afi }
+      @set_args = @get_args = { acl_name: acl_name, afi: afi }
```

lib/cisco_node_utils/acl.rb
```diff
@@ -0,0 +1,91 @@
...
+      @set_args[:state] = state
+      config_set('acl', 'acl', @set_args)
+    end
+
+    # ----------
+    # PROPERTIES
+    # ----------
+    # getter stats perentry info
+    def stats_per_entry
+      config_get('acl', 'stats_per_entry', @get_args)
```
chrisvanheuveln, line 68: Broken. You don't have a yaml entry for `stats_per_entry`

chrisvanheuveln:
@saqibraza _Do we need to take ace out of acl.yaml?_  

You can keep them in acl.yaml if you want or move them out. Your call.
 In this case I don't think it matters that much, just be aware that all config_get/config_set calls will refer to 'acl' instead of 'ace', which is okay.

lib/cisco_node_utils/acl.rb
```diff
@@ -0,0 +1,91 @@
...
+
+    def config_acl(state)
+      @set_args[:state] = state
+      config_set('acl', 'acl', @set_args)
+    end
+
+    # ----------
+    # PROPERTIES
+    # ----------
+    # getter stats perentry info
```
chrisvanheuveln: nit: these getter/setter comments don't add much value since you know what they are just by looking at the method names. I would remove them.

lib/cisco_node_utils/cmd_ref/acl.yaml
```diff
@@ -0,0 +1,33 @@
...
+#   - README_YAML.md
+#
+---
+_template:
+  config_get: 'show run aclmgr'
+  config_get_token: '/^<afi> access-list <acl_name>$/'
+  config_set: '<afi> access-list <acl_name>'
+
+ace:
+  config_get_token_append: '/^(<seqno>) .*$'
```
chrisvanheuveln: broken.

```
-  config_get_token_append: '/^(<seqno>) .*$'
+  config_get_token_append: '/^(<seqno>) .*$/'
```

lib/cisco_node_utils/cmd_ref/acl.yaml
```diff
@@ -0,0 +1,33 @@
...
+                      
+all_ace:
+  multiple:
+  config_get_token_append: '/^(\d+) .*$/'
+
+all_acl:
+  multiple:
+  config_get_token: '/^<afi> access-list (\S+)$/'
+
+stats_perentry:
```
chrisvanheuveln: camelcase (see prev comment)

tests/test_acl.rb
```diff
@@ -0,0 +1,123 @@
...
+    hash = Acl.acls
+    found = false
+    hash['ip'].each do |acl|
+      found = true if acl[1].acl_name == rtr.acl_name
+    end
+
+    assert_equal(found, true,
+                 "#{rtr.afi} acl #{rtr.acl_name}"\
+                 ' is not in the system')
+
```
chrisvanheuveln: You can simplify this somewhat and make it mostly reuseable for v6:

```
  def test_create_acl_v4
-    rtr = Acl.new(@acl_name_v4, 'ip')
-    hash = Acl.acls
-    found = false
-    hash['ip'].each do |acl|
-      found = true if acl[1].acl_name == rtr.acl_name
-    end
-
-    assert_equal(found, true,
-                 "#{rtr.afi} acl #{rtr.acl_name}"\
-                 ' is not in the system')
+    acl_name = @acl_name_v4
+    afi = 'ip'
+    rtr = Acl.new(acl_name, afi)
+    acls = Acl.acls
+    hash['ip'].key?(@acl_name_v4)
+    assert(acls[afi].key?(acl_name), "ACL #{afi} #{acl_name} is not configured")

```

tests/test_acl.rb
```diff
@@ -0,0 +1,123 @@
...
+    rtr = Acl.new(@acl_name_v4, 'ip')
+    # setter function
+    rtr.stats_per_entry = true
+    @default_show_command = "show runn | sec 'ip access-list #{@acl_name_v4}'"
+    assert_show_match(pattern: /statistics per-entry/,
+                      msg:     'failed to enable stats')
+
+    # getter function
+    val = rtr.stats_per_entry
+    assert_equal(val, true, 'value is not true')
```
chrisvanheuveln: Don't use assert_equal()  and refute_equal() for boolean tests; use assert() and refute()

tests/test_acl.rb
```diff
@@ -0,0 +1,123 @@
...
+                 "#{rtr.afi} acl #{rtr.acl_name}"\
+                 ' is not in the system')
+
+    @default_show_command = "show runn | i 'ip access-list #{@acl_name_v4}'"
+    assert_show_match(pattern: /^ip access-list #{@acl_name_v4}$/,
+                      msg:     "failed to create acl #{@acl_name_v4}")
+    rtr.destroy
+    refute_show_match(pattern: /^ip access-list #{@acl_name_v4}$/,
+                      msg:     "failed to destroy acl #{@acl_name_v4}")
+  end
```
chrisvanheuveln: You can simplify this somewhat and test both v4 and v6 in the same tester:

```
-  def test_create_acl_v4
-    rtr = Acl.new(@acl_name_v4, 'ip')
-    hash = Acl.acls
-    found = false
-    hash['ip'].each do |acl|
-      found = true if acl[1].acl_name == rtr.acl_name
-    end
-
-    assert_equal(found, true,
-                 "#{rtr.afi} acl #{rtr.acl_name}"\
-                 ' is not in the system')
-
-    @default_show_command = "show runn | i 'ip access-list #{@acl_name_v4}'"
-    assert_show_match(pattern: /^ip access-list #{@acl_name_v4}$/,
-                      msg:     "failed to create acl #{@acl_name_v4}")
-    rtr.destroy
-    refute_show_match(pattern: /^ip access-list #{@acl_name_v4}$/,
-                      msg:     "failed to destroy acl #{@acl_name_v4}")
-  end
+  def create_acl(afi, acl_name)
+    rtr = Acl.new(acl_name, afi)
+    assert(Acl.acls[afi].key?(acl_name), "ACL #{afi} #{acl_name} is not configured")
+
+    @default_show_command = "show runn | i 'ip access-list #{acl_name}'"
+    assert_show_match(pattern: /^#{afi} access-list #{acl_name}$/,
+                      msg:     "failed to create ACL #{afi} #{acl_name}")
+    rtr.destroy
+    refute_show_match(pattern: /^#{afi} access-list #{acl_name}$/,
+                      msg:     "failed to destroy ACL #{afi} #{acl_name}")
end

+  def test_create_acl
+    %w(ip ipv6).each do { |afi|
+      acl_name = afi[/ipv6/] ? @acl_name_v6 : @acl_name_v4
+      create_acl(afi, acl_name)
+    end
+  end
```

tests/test_acl.rb
```diff
@@ -0,0 +1,123 @@
...
+
+  def test_stats_enable
```
chrisvanheuveln: I would use a similar approach as my comment above to remove the redundancy by testing both afi's in a single test method.

lib/cisco_node_utils/acl.rb
```diff
@@ -0,0 +1,91 @@
...
+require_relative 'node_util'
+
+module Cisco
+  # Acl - node utility class for ACL configuration
+  class Acl < NodeUtil
+    attr_reader :acl_name, :afi
+
+    # acl_name: name of the acl
+    # instantiate: true = create acl instance
+    def initialize(acl_name, afi, instantiate=true)
```
chrisvanheuveln: nit: This is a bit of an afterthought but the order of your parameters may make better sense if swapped; ie. should your manifest titles be:
 ` cisco_acl { 'acl_name  ipv4':`  or would they be better as `cisco_acl { 'ipv4 acl_name': ` ? The difference is minor but I would probably order my manifest by afi's, so all ipv4 acls and then all the ipv6 acls. But like I said this is just a nit so you can keep it as is if you prefer.

lib/cisco_node_utils/acl.rb
yjyongz: it will cause rubocop failure (80 chars limit).

chrisvanheuveln: unlikely

```
      fail ArgumentError 'Argument afi must be ip or ipv6' unless
        afi == 'ip' || afi == 'ipv6'
```

lib/cisco_node_utils/acl.rb
```diff
@@ -0,0 +1,107 @@
...
+    end
+
+    # it will return all acls in the switch
+    def self.acls
+      afis = %w(ip ipv6)
+      acl_hash = {}
+      afis.each do |afi|
+        acl_hash[afi] = {}
+        get_args = { afi: afi }
+        instances = config_get('acl', 'all_acl', get_args)
```
chrisvanheuveln: `get_args` is unnecessary. Just use `afi: afi` here.

lib/cisco_node_utils/acl.rb
```diff
@@ -0,0 +1,107 @@
...
+      if permit
+        @set_args[:permit] = permit
+        config_set('acl', 'fragments', @set_args)
+      else
+        @set_args[:permit] = 'permit-all'
+        config_set('acl', 'fragments', @set_args)
+        @set_args[:permit] = 'deny-all'
+        config_set('acl', 'fragments', @set_args)
+      end
+    end
```
chrisvanheuveln, line 89: There's no need for all of these config_sets. Just do:

```
    def fragments=(permit)
      @set_args[:state] = (permit ? '' : 'no')
      permit = fragments unless permit
      config_set('acl', 'fragments', @set_args)
    end
```

lib/cisco_node_utils/cmd_ref/acl.yaml
```diff
@@ -0,0 +1,29 @@
...
+_template:
+  config_get: 'show run aclmgr'
+  config_get_token: '/^<afi> access-list <acl_name>$/'
+  config_set: '<afi> access-list <acl_name>'
+
+acl:
+  config_get_token: '/^<afi> access-list (\S+)$/'
+  config_set: '<state> <afi> access-list <acl_name>'
+                      
+all_acl:
```
chrisvanheuveln: Nit: To keep with our coding convention I would prefer to rename this tag `all_acls`

lib/cisco_node_utils/cmd_ref/acl.yaml
```diff
@@ -0,0 +1,29 @@
...
+  multiple:
+  config_get_token: '/^<afi> access-list (\S+)$/'
+
+fragments:
+  config_get_token_append: '/fragments permit-all$|fragments deny-all$/'
+  config_set_append: '<state> fragments <permit>'
+  default_value: nil
```
chrisvanheuveln: Use `~` here instead of `nil`. 

https://github.com/saqibraza/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/cmd_ref/README_YAML.md#default_value

tests/test_acl.rb
```diff
@@ -0,0 +1,161 @@
...
+    @default_show_command = "show runn | sec '#{afi} access-list #{acl_name}'"
+
+    # setter function
+    rtr.fragments = nil
+    refute_show_match(pattern: /fragments #{@permit}|fragments #{@deny}/,
+                      msg:     'failed to unset set fragments')
+
+    # getter function
+    val = rtr.fragments
+    assert_equal(val, 'nil', 'value not correct')
```
chrisvanheuveln: `assert_nil(val)`

The error message is okay but doesn't add a lot of value since the assert failure will already display `expected: nil`

tests/test_acl.rb
```diff
@@ -0,0 +1,161 @@
...
+  end
+
+  def fragments(afi, acl_name)
+    rtr = Acl.new(acl_name, afi)
+
+    @default_show_command = "show runn | sec '#{afi} access-list #{acl_name}'"
+    assert_show_match(pattern: /^#{afi} access-list #{acl_name}$/,
+                      msg:     "failed to create acl #{acl_name}")
+
+    # set 'no frgaments ...'
```
chrisvanheuveln: `fragments`

tests/test_acl.rb
```diff
@@ -0,0 +1,161 @@
...
+    unset_fragments(rtr, afi, acl_name)
+
+    # set 'fragments permit-all' from nothing set
+    set_fragments(rtr, afi, acl_name, @permit)
```
chrisvanheuveln, line 142: nit: These `@permit` `@deny` class vars are okay but you're basically using them as constants. And they don't save that much typing so I would just use `permit-all` and `deny-all` instead since it makes the code clearer. Just a style preference - keep what you have if you like that better.

chrisvanheuveln:
Please don't add any more new properties to this PR. The code creep is extending this US and I'd like to see it closed out soon. Thanks.

tests/test_acl.rb
```diff
@@ -0,0 +1,161 @@
...
+                      msg:     "failed to destroy acl #{acl_name}")
+  end
+
+  def test_fragments
+    %w(ip ipv6).each do |afi|
+      acl_name = afi[/ipv6/] ? @acl_name_v6 : @acl_name_v4
+      fragments(afi, acl_name)
+    end
+  end
+end
```
chrisvanheuveln, line 171: Update CHANGELOG.md

lib/cisco_node_utils/acl.rb
saqibraza, line 89: I don't understand what 'permit = fragments unless permit' is doing?

When we get a nil value for permit, we need to do "no fragments permit-all" and "no fragments deny-all". Else permit will be either "permit-all", or "deny-all", in which case we need to do "fragments <permit>"

chrisvanheuveln, line 89: Yes I understand the intent and my code will do the appropriate command. 

`permit = fragments unless permit` does this:

If permit is an empty string then:  set permit to the _current_ permit value by calling the `fragments` getter.

If the current value is 'deny-all' then permit is set to 'deny-all'

TBH, `permit` is a poor choice for this variable name since it can also mean deny-all. I would use `action` instead.

saqibraza, line 89: understood

saqibraza, line 89: Is this okay: 

def fragments=(action)
      @set_args[:state] = (action ? '' : 'no')
      action = fragments unless action
      @set_args[:action] = action
      config_set('acl', 'fragments', @set_args) if action
    end
- action = fragments can still set action to nil

chrisvanheuveln, line 89: It can but the agent shouldn't be calling to remove it if it's already gone.
As a code goodness though it makes sense, so yes it's okay to code it this way.

tests/test_acl.rb
saqibraza, line 171: Can you please guide me what to put in CHANGELOG.md?

tests/test_acl.rb
```diff
@@ -0,0 +1,161 @@
...
+    @default_show_command = "show runn | sec '#{afi} access-list #{acl_name}'"
+
+    # setter function
+    rtr.fragments = option
+    assert_show_match(pattern: /fragments #{option}/,
+                      msg:     'failed to set fragments #{option} ' + option)
+
+    # getter function
+    val = rtr.fragments
+    assert_equal(val, option, 'value not correct')
```
chrisvanheuveln: `assert_equal(option, rtr.fragments)`

PR #112, opened 2015-12-03
ace node util part

yjyongz:
ace part only for node util dev

### DIFF
--- a/lib/cisco_node_utils/ace.rb
+++ b/lib/cisco_node_utils/ace.rb
@@ -44,7 +44,7 @@ module Cisco
                  'portgroup \S+)?'\
                  ' *(?<option>[a-zA-Z0-9\-\/ ]*)*'
     end
-    
+
     # Create a hash of all aces under give acl_name.
     def self.aces
       regexp_str = '(?<seqno>\d+) (?<action>\S+)'\
@@ -69,13 +69,11 @@ module Cisco
           get_args[:acl_name] = name
           aces = config_get('acl', 'all_ace', get_args)
           next if aces.nil?
-          #puts aces
           aces.each do |ace|
             regexp = Regexp.new(regexp_str)
             match = regexp.match(ace)
             item = Ace.new(name, match[1].to_i, afi)
             item.seqno = match[1].to_i
-            #puts " #{match[1]} #{match[2]} #{match[3]}"
             item.action = match[2]
             item.proto = match[3]
             item.src_addr = match[4]
--- a/lib/cisco_node_utils/ace.rb
+++ b/lib/cisco_node_utils/ace.rb
@@ -79,13 +79,7 @@ module Cisco
     def ace_set(attrs)
       state = attrs.empty? ? 'no ' : ''
       @set_args[:state] = state
-      @set_args[:action] = attrs[:action]
-      @set_args[:proto] = attrs[:proto]
-      @set_args[:src_addr] = attrs[:src_addr]
-      @set_args[:src_port] = attrs[:src_port]
-      @set_args[:dst_addr] = attrs[:dst_addr]
-      @set_args[:dst_port] = attrs[:dst_port]
-      @set_args[:option_format] = attrs[:option_format]
+      @set_args.merge!(attrs)
       config_set('acl', 'ace', @set_args)
     end
 
--- a/lib/cisco_node_utils/ace.rb
+++ b/lib/cisco_node_utils/ace.rb
@@ -1,0 +1,227 @@
+# Copyright (c) 2014-2015 Cisco and/or its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative 'node_util'
+
+module Cisco
+  # Ace - node utility class for Ace Configuration
+  class Ace < NodeUtil
+    attr_reader :acl_name, :seqno, :afi
+    # acl_name is name of acl
+    # seqno is sequence number of ace
+    # afi is either v4 or v6
+    def initialize(acl_name, seqno, afi)
+      fail TypeError unless
+        acl_name.is_a?(String) && seqno.is_a?(Integer) && afi.is_a?(String)
+      fail ArgumentError 'we expect ip or ipv6' unless
+        afi == 'ip' || afi == 'ipv6'
+      @acl_name = acl_name
+      @afi = afi
+      @seqno = seqno
+      @set_args = @get_args = { acl_name: @acl_name, seqno: @seqno, afi: @afi,
+                                src_port: '', dst_port: '',
+                                option_format: '' }
+      @regexp_str = '(?<seqno>\d+) (?<action>\S+)'\
+                 ' *(?<proto>\d+|\S+)'\
+                 ' *(?<src_addr>any|host \S+|\S+\/\d+|\S+ [:\.0-9a-fA-F]+|'\
+                 'addrgroup \S+)*'\
+                 ' *(?<src_port>eq \S+|neq \S+|lt \S+|''gt \S+|range \S+ \S+|'\
+                 'portgroup \S+)?'\
+                 ' *(?<dst_addr>any|host \S+|\S+\/\d+|\S+ [:\.0-9a-fA-F]+|'\
+                 'addrgroup \S+)'\
+                 ' *(?<dst_port>eq \S+|neq \S+|lt \S+|gt \S+|range \S+ \S+|'\
+                 'portgroup \S+)?'\
+                 ' *(?<option>[a-zA-Z0-9\-\/ ]*)*'
+    end
+    
+    # Create a hash of all aces under give acl_name.
+    def self.aces
+      regexp_str = '(?<seqno>\d+) (?<action>\S+)'\
+                 ' *(?<proto>\d+|\S+)'\
+                 ' *(?<src_addr>any|host \S+|\S+\/\d+|\S+ [:\.0-9a-fA-F]+|'\
+                 'addrgroup \S+)*'\
+                 ' *(?<src_port>eq \S+|neq \S+|lt \S+|''gt \S+|range \S+ \S+|'\
+                 'portgroup \S+)?'\
+                 ' *(?<dst_addr>any|host \S+|\S+\/\d+|\S+ [:\.0-9a-fA-F]+|'\
+                 'addrgroup \S+)'\
+                 ' *(?<dst_port>eq \S+|neq \S+|lt \S+|gt \S+|range \S+ \S+|'\
+                 'portgroup \S+)?'\
+                 ' *(?<option>[a-zA-Z0-9\-\/ ]*)*'
+      afis = %w(ip ipv6)
+      hash = {}
+      afis.each do |afi|
+        get_args = { afi: afi }
+        instances = config_get('acl', 'all_acl', get_args)
+        next if instances.nil?
+        instances.each do |name|
+          hash[name] = {}
+          get_args[:acl_name] = name
+          aces = config_get('acl', 'all_ace', get_args)
+          next if aces.nil?
+          #puts aces
+          aces.each do |ace|
+            regexp = Regexp.new(regexp_str)
+            match = regexp.match(ace)
+            item = Ace.new(name, match[1].to_i, afi)
+            item.seqno = match[1].to_i
+            #puts " #{match[1]} #{match[2]} #{match[3]}"
+            item.action = match[2]
+            item.proto = match[3]
+            item.src_addr = match[4]
+            item.src_port = match[5]
+            item.dst_addr = match[6]
+            item.dst_port = match[7]
+            item.option_format = match[8]
+            hash[name][match[1].to_i] = item
+          end
+        end
+      end
+      hash
+    end
+
+    # Destroy a router instance; disable feature on last instance
+    def destroy
+      config_ace('no')
+    end
+
+    # First create acl
+    def config_ace(state='')
+      @set_args[:state] = state
+      config_set('acl', 'ace', @set_args)
+    end
+
+    # ----------
+    # PROPERTIES
+    # ----------
+    # getter of seqno
+    def seqno
+      str = config_get('acl', 'ace', @get_args)
+      return if str.nil?
+
+      regexp = Regexp.new(@regexp_str)
+      match = regexp.match(str)
+      match[1]
+    end
+
+    # setter of seqno
+    def seqno=(state)
+      @set_args[:state] = state ? '' : 'no'
+    end
+
+    # getter of action
+    def action
+      str = config_get('acl', 'ace', @get_args)
+      return if str.nil?
+
+      regexp = Regexp.new(@regexp_str)
+      match = regexp.match(str)
+      match[2]
+    end
+
+    # setter of action
+    def action=(action)
+      @set_args[:action] = action
+    end
+
+    # getter of proto
+    def proto
+      str = config_get('acl', 'ace', @get_args)
+      return if str.nil?
+
+      regexp = Regexp.new(@regexp_str)
+      match = regexp.match(str)
+      match[3]
+    end
+
+    # setter of proto
+    def proto=(proto)
+      @set_args[:proto] = proto
+    end
+
+    # getter of src_addr
+    def src_addr
+      str = config_get('acl', 'ace', @get_args)
+      return if str.nil?
+
+      regexp = Regexp.new(@regexp_str)
+      match = regexp.match(str)
+      match[4]
+    end
+
+    # setter of src_addr
+    def src_addr=(src_addr)
+      @set_args[:src_addr] = src_addr
+    end
+
+    # getter of src_port
+    def src_port
+      str = config_get('acl', 'ace', @get_args)
+      return if str.nil?
+
+      regexp = Regexp.new(@regexp_str)
+      match = regexp.match(str)
+      match[5]
+    end
+
+    # setter of src_port_format
+    def src_port=(src_port)
+      @set_args[:src_port] = src_port
+    end
+
+    # getter of dst_addr_format
+    def dst_addr
+      str = config_get('acl', 'ace', @get_args)
+      return if str.nil?
+
+      regexp = Regexp.new(@regexp_str)
+      match = regexp.match(str)
+      match[6]
+    end
+
+    # setter of dst_addr_format
+    def dst_addr=(dst_addr)
+      @set_args[:dst_addr] = dst_addr
+    end
+
+    # getter of dst_port_format
+    def dst_port
+      str = config_get('acl', 'ace', @get_args)
+      return if str.nil?
+
+      regexp = Regexp.new(@regexp_str)
+      match = regexp.match(str)
+      match[7]
+    end
+
+    # setter of dst_port_format
+    def dst_port=(src_port)
+      @set_args[:dst_port] = src_port
+    end
+
+    # getter of option_format
+    def option_format
+      str = config_get('acl', 'ace', @get_args)
+      return if str.nil?
+
+      regexp = Regexp.new(@regexp_str)
+      match = regexp.match(str)
+      match[8]
+    end
+
+    # setter of option_format
+    def option_format=(option)
+      @set_args[:option_format] = option
+    end
+  end
+end
--- a/lib/cisco_node_utils/ace.rb
+++ b/lib/cisco_node_utils/ace.rb
@@ -18,10 +18,11 @@ module Cisco
   # Ace - node utility class for Ace Configuration
   class Ace < NodeUtil
     attr_reader :acl_name, :seqno, :afi
+
     # acl_name is name of acl
     # seqno is sequence number of ace
     # afi is either v4 or v6
-    def initialize(acl_name, seqno, afi)
+    def initialize(afi, acl_name, seqno)
       fail TypeError unless
         acl_name.is_a?(String) && seqno.is_a?(Integer) && afi.is_a?(String)
       fail ArgumentError 'we expect ip or ipv6' unless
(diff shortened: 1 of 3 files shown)


mikewiebe:
Make sure ACE comments from (PR  https://github.com/cisco/cisco-network-node-utils/pull/99) are addressed in this review

lib/cisco_node_utils/ace.rb
```diff
@@ -0,0 +1,225 @@
...
+      fail TypeError unless
+        acl_name.is_a?(String) && seqno.is_a?(Integer) && afi.is_a?(String)
+      fail ArgumentError 'we expect ip or ipv6' unless
+        afi == 'ip' || afi == 'ipv6'
+      @acl_name = acl_name
+      @afi = afi
+      @seqno = seqno
+      @set_args = @get_args = { acl_name: @acl_name, seqno: @seqno, afi: @afi,
+                                src_port: '', dst_port: '',
+                                option_format: '' }
```
chrisvanheuveln: You don't need these class vars, and you don't need to specify unused keys for your set/get_args:

```
-      @acl_name = acl_name
-      @afi = afi
-      @seqno = seqno
-      @set_args = @get_args = { acl_name: @acl_name, seqno: @seqno, afi: @afi,
-                                src_port: '', dst_port: '',
-                                option_format: '' }
+      @set_args = @get_args = { acl_name: acl_name, seqno: seqno, afi: afi)
```

lib/cisco_node_utils/ace.rb
```diff
@@ -0,0 +1,225 @@
...
+  # Ace - node utility class for Ace Configuration
+  class Ace < NodeUtil
+    attr_reader :acl_name, :seqno, :afi
+    # acl_name is name of acl
+    # seqno is sequence number of ace
+    # afi is either v4 or v6
+    def initialize(acl_name, seqno, afi)
+      fail TypeError unless
+        acl_name.is_a?(String) && seqno.is_a?(Integer) && afi.is_a?(String)
+      fail ArgumentError 'we expect ip or ipv6' unless
```
chrisvanheuveln, line 28: `fail ArgumentError 'Argument afi must be ip or ipv6' unless...`

lib/cisco_node_utils/ace.rb
```diff
@@ -0,0 +1,225 @@
...
+                 ' *(?<proto>\d+|\S+)'\
+                 ' *(?<src_addr>any|host \S+|\S+\/\d+|\S+ [:\.0-9a-fA-F]+|'\
+                 'addrgroup \S+)*'\
+                 ' *(?<src_port>eq \S+|neq \S+|lt \S+|''gt \S+|range \S+ \S+|'\
+                 'portgroup \S+)?'\
+                 ' *(?<dst_addr>any|host \S+|\S+\/\d+|\S+ [:\.0-9a-fA-F]+|'\
+                 'addrgroup \S+)'\
+                 ' *(?<dst_port>eq \S+|neq \S+|lt \S+|gt \S+|range \S+ \S+|'\
+                 'portgroup \S+)?'\
+                 ' *(?<option>[a-zA-Z0-9\-\/ ]*)*'
```
chrisvanheuveln: What is the difference between regexp_str and @regexp_str ?

lib/cisco_node_utils/ace.rb
```diff
@@ -0,0 +1,225 @@
...
+    end
+
+    # Create a hash of all aces under give acl_name.
+    def self.aces
+      regexp_str = '(?<seqno>\d+) (?<action>\S+)'\
```
chrisvanheuveln: You defined the same regexp_str as @regexp_str in initialize but you don't need it in either place. Remove them both.

lib/cisco_node_utils/ace.rb
```diff
@@ -0,0 +1,225 @@
+# Copyright (c) 2014-2015 Cisco and/or its affiliates.
```
chrisvanheuveln, line 1: I'd like to put this review on hold for now. The ACL code in #111 looks pretty good (once final comments are addressed) so I'd like to see that one get wrapped up and merged soon because I have some thoughts on this ace code that will rely on the ACL code being in place and will hopefully simplify the ace code a bit more. I think we should have a short design meeting tomorrow or monday to discuss. 

edit: I set up a call for today at 2:30 EST.

lib/cisco_node_utils/ace.rb
```diff
@@ -0,0 +1,225 @@
...
+            item.src_port = match[5]
+            item.dst_addr = match[6]
+            item.dst_port = match[7]
+            item.option_format = match[8]
+            hash[name][match[1].to_i] = item
+          end
+        end
+      end
+      hash
+    end
```
chrisvanheuveln, line 75: You don't need to set properties in self.aces, it should look like this instead:

```
  def self.aces
    # Create a hash of AFIs/ACLs/ACEs. We only need the ACE seqno, not the properties.
    # ip access-list bar
    #   10 deny tcp any any
    #   20 permit ip any any
    def self.aces
      afis = %w(ip ipv6)
      hash_final = {}
      afis.each do |afi|
        # Get a list of all acls for this afi
        acls = config_get('acl', 'all_acls', afi: afi)
        #             cli: 'ip access-list foo'
        #      config_get: ['foo']
        next if acls.nil?

        # Walk each acl and get a list of aces for the given acl
        acls.each do |acl_name|
          hash_tmp[acl_name] = {}
          aces = config_get('acl', 'all_aces', afi: afi)
          #             cli: '10 deny tcp any any'
          #      config_get: ['10', '20']
          next if aces.nil?

          aces.each do |seqno|
            hash_tmp[acl_name][seqno] = Ace.new(afi, acl, seqno, false)
          end
        end
        puts "hash_tmp: #{hash_temp}"
      end
      hash_final
    end
  end
```

lib/cisco_node_utils/ace.rb
```diff
@@ -0,0 +1,225 @@
...
+
```
chrisvanheuveln, line 76: You need a common ace getter that just grabs the entire ace and sorts it out with a regex. 
Note: You need to use Regexp.new in order to have it auto-generate the match tags.

```
  # Common ace getter
  def ace_get
    str = config_get('acl', 'ace', @get_args)
    return nil if str.nil?
    # Possible matches
    #   10 permit ip host 1.1.1.1
    #   20 permit ip host 1.1.1.0/24
    #   30 permit ip host 1.1.1.0 0.0.0.255
    #   40 permit ip addrgroup 
    # rubocop:disable Metrics/LineLength
    regexp =
      Regexp.new('(?<seqno>\d+) (?<action>\S+)'\
                 ' *(?<proto>\d+|\S+)'\
                 ' *(?<src_addr>any|host \S+|\S+ [\d.]+|addrgroup \S+)*'\
                 ' *(?<src_port>eq \S+|neq \S+|lt \S+|''gt \S+|range \S+ \S+|portgroup \S+)?'\
                 ' *(?<dst_addr>any|host \S+|\S+ [\d.]+|addrgroup \S+)'\
                 ' *(?<dst_port>eq \S+|neq \S+|lt \S+|gt \S+|range \S+ \S+|portgroup \S+)?'\
                 ' *(?<option>[a-zA-Z0-9\-\/ ]*)*')
    # rubocop:enable Metrics/LineLength

    # Return a regexp match hash
    regexp.match(str)
  end
```

Now you can get the individual matches just by referencing the hash key:

```
  # Getters
  def action
    match = ace_get
    return nil if match.nil?
    match[:action]
  end

  def proto
    match = ace_get
    return nil if match.nil?
    match[:proto]
  end

  def src_addr
    match = ace_get
    return nil if match.nil?
    match[:src_addr]
     ...
```

And the setter is very simple:

```
  # Common setter. Put the values you need in a hash and pass it in.
  # attrs = {:action=>'permit', :proto=>'ip', :src =>'host 1.1.1.1'}
  def ace_set(attrs)
    state = attrs.empty? 'no ' : ''
    config_set('acl', 'ace', @set_args, attrs)
  end
```

lib/cisco_node_utils/ace.rb
```diff
@@ -0,0 +1,225 @@
...
+    # Destroy a router instance; disable feature on last instance
+    def destroy
+      config_ace('no')
+    end
+
+    # First create acl
+    def config_ace(state='')
+      @set_args[:state] = state
+      config_set('acl', 'ace', @set_args)
+    end
```
chrisvanheuveln, line 84: Should not need these `destroy` and `config_ace` methods.

lib/cisco_node_utils/ace.rb
```diff
@@ -0,0 +1,225 @@
...
+
+    # ----------
+    # PROPERTIES
+    # ----------
+    # getter of seqno
+    def seqno
```
chrisvanheuveln: `seqno` is your namevar so you don't need `seqno` and `seqno=` methods.

lib/cisco_node_utils/ace.rb
```diff
@@ -0,0 +1,225 @@
...
+      match[1]
+    end
+
+    # setter of seqno
+    def seqno=(state)
+      @set_args[:state] = state ? '' : 'no'
+    end
+
+    # getter of action
+    def action
```
chrisvanheuveln, line 89: Replace `action`, `proto`, `src_addr`, `src_port`, `dst_addr`, `dst_port` getters & setters with the method examples I showed in the earlier comment.

lib/cisco_node_utils/ace.rb
```diff
@@ -0,0 +1,225 @@
...
+      match[7]
+    end
+
+    # setter of dst_port_format
+    def dst_port=(src_port)
+      @set_args[:dst_port] = src_port
+    end
+
+    # getter of option_format
+    def option_format
```
chrisvanheuveln: Let's discuss this one after the rewrites above. It will have to change but I'm mulling over some options.

lib/cisco_node_utils/cmd_ref/acl.yaml
```diff
@@ -0,0 +1,33 @@
...
+---
+_template:
+  config_get: 'show run aclmgr'
+  config_get_token: '/^<afi> access-list <acl_name>$/'
+  config_set: '<afi> access-list <acl_name>'
+
+ace:
+  config_get_token_append: '/^<seqno> .+$/'
+  config_set_append:
+      '<state> <seqno> <action> <proto> <src_addr> <src_port> <dst_addr> <dst_port> <option_format>'
```
chrisvanheuveln: Does that syntax work? I'm surprised that yaml doesn't complain about the split line.

lib/cisco_node_utils/cmd_ref/acl.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+acl:
+  config_get_token: '/^<afi> access-list (\S+)$/'
+  config_set: '<state> <afi> access-list <acl_name>'
+                      
+all_ace:
+  multiple:
+  config_get_token_append: '/^\d+ .+$/'
+
+all_acl:
```
chrisvanheuveln: Use `all_aces` and `all_acls` to keep with our naming conventions.

lib/cisco_node_utils/ace.rb
yjyongz: Hi Chris
my acl,yaml file it has 
<src_port> and <dst_port> tag , i have to fill these in. 
if i leave it empty, does infra automatically fill it for me?

lib/cisco_node_utils/ace.rb
yjyongz, line 28: this change will hit rubocop failure (80 chars limit)

lib/cisco_node_utils/ace.rb
yjyongz, line 75: Chris, 
hash_tmp[acl_name][seqno] = Ace.new(afi, acl, seqno, false) 
this entry leave properties underpopulated. 
etc. src_addr, src_port

lib/cisco_node_utils/cmd_ref/acl.yaml
yjyongz: done

lib/cisco_node_utils/cmd_ref/acl.yaml
yjyongz: yes it works it passed mini-test

lib/cisco_node_utils/ace.rb
chrisvanheuveln: If you don't set a tag then it's nil, so it's essentially ignored by yaml and our cmd_ref code.

Furthermore, src_port, dst_port, and option_format are properties that will be populated by your getters and setters; they should not be set in initialize.

lib/cisco_node_utils/ace.rb
chrisvanheuveln, line 75: I think that you have a misunderstanding about how self.aces is used. It is not used to populate properties. Let's discuss these changes in a design meeting.

lib/cisco_node_utils/cmd_ref/acl.yaml
chrisvanheuveln: I'm surprised but okay. Please move it up to the config_set_append line anyway so that I can find it when I grep through yaml files.

lib/cisco_node_utils/ace.rb
```diff
@@ -0,0 +1,225 @@
...
+require_relative 'node_util'
+
+module Cisco
+  # Ace - node utility class for Ace Configuration
+  class Ace < NodeUtil
+    attr_reader :acl_name, :seqno, :afi
+    # acl_name is name of acl
+    # seqno is sequence number of ace
+    # afi is either v4 or v6
+    def initialize(acl_name, seqno, afi)
```
chrisvanheuveln: Let's order these params in the order that the cli uses them: `(afi, acl_name, seqno)`

tests/test_ace.rb
```diff
@@ -0,0 +1,134 @@
...
+  end
+
+  def no_ip_access_list_foo
+    # Turn the feature off for a clean test.
+    config('no ip access-list ' + @acl_name)
+  end
+
+  # TESTS
+
+  def test_create_destroy_ipv4_ace_one
```
chrisvanheuveln: You're going to have a lot of duplicated code if you create separate v4 and v6 tests that are essentially testing the same things. You can structure your code to use a common test method that's used for both - see the ACL test code here for an example: https://github.com/cisco/cisco-network-node-utils/pull/111/files#diff-5dde18e7fed82ed58d46532b7f59c873R43

The only difference is that you'll need to pass in some v4/v6 specific variables as well, which you could do by setting up a hash first. The benefit of the hash is that you can use it in the call to ace_set as well, e.g.:

```
  def test_create_destroy_ace_one
    props = {
      'ip' => {
        :afi => 'ip'
        :acl_name => @acl_name_v4,
        :proto' => 'tcp',
        :src_addr => '7.8.9.6 2.3.4.5',
        :src_port => 'eq 40',
        :dst_addr => '1.2.3.4/32',
        :dst_port => 'neq 20',
        :option => 'precedence critical',
      },
      'ipv6' => {
        :afi => 'ipv6'
        :acl_name => @acl_name_v6,
        :proto => '6',
        :src_addr => 'addrgroup fi',
        :src_port => 'eq 30',
        :dst_addr => '1::7/32',
        :dst_port => 'neq 20',
        :option => 'dscp cs2 fragments packet-length eq 30',
      }
    }

    %w(ip ipv6).each do |afi|
      create_destroy_ace_one(props[afi])
    end
  end

  def create_destroy_ace_one(p)

    # Test seqno 10 with permit
    p[:action] = 'permit'
    ace = Ace.new(p[:afi], p[:acl_name], 10)
    ace.ace_set(p)
    assert_equal(p[:action], ace.action)
    assert_equal(p[:src_addr], ace.src_addr)
    assert_equal(p[:src_port], ace.src_port)
      ...
    # I don't think we need assert_show_match for this test. May revisit.

    # Test seqno 20 with deny any
    p[:action] = 'deny'
    p[:src_addr] = 'any'
    p[:dst_addr] = 'any'
    ace = Ace.new(p[:afi], p[:acl_name], 20)
    ace.ace_set(p)
    assert_equal(p[:action], ace.action)
    assert_equal(p[:src_addr], ace.src_addr)
    assert_equal(p[:src_port], ace.src_port)
      ...
```

chrisvanheuveln:
Update CHANGELOG.md

lib/cisco_node_utils/ace.rb
yjyongz, line 76: changed

tests/test_ace.rb
yjyongz: looks good ... i like this

lib/cisco_node_utils/cmd_ref/acl.yaml
```diff
@@ -0,0 +1,33 @@
...
+  config_set_append:
+      '<state> <seqno> <action> <proto> <src_addr> <src_port> <dst_addr> <dst_port> <option_format>'
+
+acl:
+  config_get_token: '/^<afi> access-list (\S+)$/'
+  config_set: '<state> <afi> access-list <acl_name>'
+                      
+all_ace:
+  multiple:
+  config_get_token_append: '/^\d+ .+$/'
```
chrisvanheuveln: `'/^(\d+) /'`

lib/cisco_node_utils/cmd_ref/acl.yaml
```diff
@@ -0,0 +1,33 @@
...
+#   - README_YAML.md
+#
+---
+_template:
+  config_get: 'show run aclmgr'
+  config_get_token: '/^<afi> access-list <acl_name>$/'
+  config_set: '<afi> access-list <acl_name>'
+
+ace:
+  config_get_token_append: '/^<seqno> .+$/'
```
chrisvanheuveln, line 13: `'/^<seqno> (.*)$/'`

lib/cisco_node_utils/ace.rb
chrisvanheuveln, line 76: ace_set needs a little more to work right. You can look at bgp_neighbor_af for an example of using set_args_keys helper methods, but this is essentially how it should look:

```
  def ace_set(attrs)
    set_args_keys(attrs)
    @set_args[:state] = attrs.empty? 'no ' : ''
    config_set('acl', 'ace', @set_args)
  end

    def set_args_keys(hash={})
      keys = { afi: @afi, acl_name: @acl_name, seqno: @seqno }
      @get_args = @set_args = keys
      @set_args = @get_args.merge!(hash) unless hash.empty?
    end
```

lib/cisco_node_utils/ace.rb
```diff
@@ -0,0 +1,179 @@
...
+  class Ace < NodeUtil
+    attr_reader :acl_name, :seqno, :afi
+
+    # acl_name is name of acl
+    # seqno is sequence number of ace
+    # afi is either v4 or v6
+    def initialize(afi, acl_name, seqno)
+      fail TypeError unless
+        acl_name.is_a?(String) && seqno.is_a?(Integer) && afi.is_a?(String)
+      fail ArgumentError 'we expect ip or ipv6' unless
```
chrisvanheuveln, line 28: Please fix this error msg. See previous review comments.

lib/cisco_node_utils/ace.rb
```diff
@@ -0,0 +1,179 @@
...
+      regexp = Regexp.new('(?<seqno>\d+) (?<action>\S+)'\
+                 ' *(?<proto>\d+|\S+)'\
+                 ' *(?<src_addr>any|host \S+|\S+\/\d+|\S+ [:\.0-9a-fA-F]+|'\
+                 'addrgroup \S+)*'\
+                 ' *(?<src_port>eq \S+|neq \S+|lt \S+|''gt \S+|range \S+ \S+|'\
+                 'portgroup \S+)?'\
+                 ' *(?<dst_addr>any|host \S+|\S+\/\d+|\S+ [:\.0-9a-fA-F]+|'\
+                 'addrgroup \S+)'\
+                 ' *(?<dst_port>eq \S+|neq \S+|lt \S+|gt \S+|range \S+ \S+|'\
+                 'portgroup \S+)?'\
```
chrisvanheuveln, line 50: Please don't wrap these regexp lines. See previous review comment for an example.

lib/cisco_node_utils/ace.rb
```diff
@@ -0,0 +1,179 @@
...
+                 ' *(?<option_format>[a-zA-Z0-9\-\/ ]*)*')
+      regexp.match(str)
+    end
+
+    # Create a hash of all aces under give acl_name.
+    def self.aces
```
chrisvanheuveln, line 56: Please move this class method under the initialize method. Our convention is to have all of the class methods at the top.

lib/cisco_node_utils/ace.rb
```diff
@@ -0,0 +1,179 @@
...
+      hash
+    end
+
+    # common setter. Put the values you need in a hash and pass it in.
+    # attrs = {:action=>'permit', :proto=>'ip', :src =>'host 1.1.1.1'}
+    def ace_set(attrs)
+      state = attrs.empty? ? 'no ' : ''
+      @set_args[:state] = state
+      @set_args.merge!(attrs) unless attrs.empty?
+      config_set('acl', 'ace', @set_args)
```
chrisvanheuveln, line 83: We discussed that this method needs to have a set_args_keys() call in order to work correctly. See previous review comment.

lib/cisco_node_utils/ace.rb
```diff
@@ -0,0 +1,179 @@
...
+    def action
+      match = ace_get
+      return nil if match.nil?
+
+      match[:action]
+    end
+
+    # setter of action
+    def action=(action)
+      @set_args[:action] = action
```
chrisvanheuveln, line 98: This can't possibly have worked (as in configured the switch) without calling ace_set? Did you test this code? 

```
  def action=(action)
    @set_args[:action] = action
    ace_set
  end
```

lib/cisco_node_utils/cmd_ref/acl.yaml
chrisvanheuveln: Are you going to address this comment?

lib/cisco_node_utils/cmd_ref/acl.yaml
```diff
@@ -0,0 +1,32 @@
...
+  multiple:
+  config_get_token_append: '/^(\d+) .+$/'
+
+all_acls:
+  multiple:
+  config_get_token: '/^<afi> access-list (\S+)$/'
+
+stats_per_entry:
+  kind: boolean
+  config_get_token_append: '/statistics per-entry$/'
```
chrisvanheuveln, line 30: You need to anchor this regex:

`config_get_token_append: '/^statistics per-entry$/'`

lib/cisco_node_utils/ace.rb
yjyongz, line 98: Hi Christ,
yes i definitely run the mini-test for this change
ace_set is the final step to create ace instance.
before ace_set, we should fill up all the properties like seqno, action, protocol etc..

chrisvanheuveln, line 98: Ah, of course that's right. nm.

PR #118, opened 2015-12-08
channel_group setter fixes

chrisvanheuveln:
-yaml now uses default false
-puppet tests pass
-minitest interface.rb: 43 runs, 319 assertions, 0 failures, 0 errors, 1 skips

### DIFF
--- a/lib/cisco_node_utils/interface.rb
+++ b/lib/cisco_node_utils/interface.rb
@@ -84,19 +84,21 @@ module Cisco
 
     def channel_group=(val)
       fail "channel_group is not supported on #{name}" unless @name[/Ethernet/i]
-      if val.to_s.empty?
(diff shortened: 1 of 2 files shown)


lib/cisco_node_utils/interface.rb
```diff
@@ -84,19 +84,21 @@ def channel_group
 
     def channel_group=(val)
       fail "channel_group is not supported on #{name}" unless @name[/Ethernet/i]
```
mikewiebe: I know this is not your code.. but should `#{name}` be `#{@name}` instead?

chrisvanheuveln: It probably works as is since this file has an `attr :name` defined I agree it should be changed.

PR #119, opened 2015-12-08
Feature/grpc nbr af

RichWellum:
Node Utils for BGP Neighbor Address-Family nexus equivalence.

Mini test runs fine on both platforms:

XR:
13 runs, 242 assertions, 0 failures, 0 errors, 0 skips

Nexus:
13 runs, 341 assertions, 0 failures, 0 errors, 0 skips

Rubocop happy across the board.

Please review.


lib/cisco_node_utils/bgp_neighbor_af.rb
```diff
@@ -70,6 +70,11 @@ def validate_args(asn, vrf, nbr, af)
       fail ArgumentError, "'af' must be an array specifying afi and safi" unless
         af.is_a?(Array) || af.length == 2
 
+      # XR BGP does not support <address>/<mask>
+      if platform == :ios_xr && nbr['/'] == '/'
+        fail ArgumentError, "IOS XR does not support 'slash' notation"
```
glennmatthews: UnsupportedError?

lib/cisco_node_utils/bgp_neighbor_af.rb
```diff
@@ -110,10 +115,12 @@ def destroy
 
     # Returns ['<map1>', '<map2>']
     def advertise_map_exist
+      fail ArgumentError if platform == :ios_xr
```
glennmatthews: Why doesn't the default _exclude behavior (return nil for the getter, raise UnsupportedError for the setter) work for this property? Why do you need to explicitly raise an error?

lib/cisco_node_utils/bgp_neighbor_af.rb
```diff
@@ -257,7 +269,7 @@ def default_additional_paths_send
     def default_originate_get
       val = config_get('bgp_neighbor_af', 'default_originate', @get_args)
       return nil unless val
-      (val[/route-map/]) ? val.split.last : true
+      (val[/route-/]) ? val.split.last : true
```
glennmatthews: would it be better to do `/route-(map|policy)/`?

lib/cisco_node_utils/bgp_neighbor_af.rb
```diff
@@ -544,23 +563,49 @@ def default_route_reflector_client
...
       val
     end
 
+    # XR: 'send-community-ebgp' and 'send-extended-community-ebgp' are
+    #  the equivalents of the NXOS: standard | extended functionality
+    #  Returns: node, 'send-community-ebgp', 'send-extended-community-ebgp' or
+    #  'send-community-ebgp send-extended-community-ebgp' which is the 'both'
+    # keyword equivalent
+    def send_comm_iosxr_get(val)
+      if val == ['send-community-ebgp', 'send-extended-community-ebgp']
```
glennmatthews, line 583: Is the ordering here guaranteed or can it differ?

lib/cisco_node_utils/bgp_neighbor_af.rb
```diff
@@ -544,23 +563,49 @@ def default_route_reflector_client
...
       return 'standard' if val[/send-community/] # Workaround
       val
     end
 
+    # XR: 'send-community-ebgp' and 'send-extended-community-ebgp' are
+    #  the equivalents of the NXOS: standard | extended functionality
+    #  Returns: node, 'send-community-ebgp', 'send-extended-community-ebgp' or
+    #  'send-community-ebgp send-extended-community-ebgp' which is the 'both'
+    # keyword equivalent
+    def send_comm_iosxr_get(val)
```
glennmatthews: Nitpick: I'd prefer `ios_xr` rather than `iosxr` in order to remain self-consistent.

lib/cisco_node_utils/bgp_neighbor_af.rb
```diff
@@ -579,6 +624,47 @@ def send_community=(val)
       config_set('bgp_neighbor_af', 'send_community', @set_args)
     end
 
+    def send_comm_iosxr_set(val)
+      if val[/none/]
```
glennmatthews: is it correct for these to be regexp matches instead of just plain string equality checks?

lib/cisco_node_utils/cmd_ref/bgp_neighbor_af.yaml
```diff
@@ -122,54 +143,77 @@ prefix_list_out:
...
-  config_set_append: '<state> send-community <attr>'
-  default_value: 'none'
+  cli_nexus:
+    config_get_token_append: '/^send-community(?: .*)?/'
+    config_set_append: '<state> send-community <attr>'
+    default_value: 'none'
+  cli_ios_xr:
+    multiple:
+    # XR four seperate commands: send-community-ebgp  send-community-gshut-ebgp
+    # send-extended-community-ebgp  send-multicast-attributes
```
glennmatthews: Remove reference to `send-multicast-attributes` unless you're _sure_ it's relevant to send-community.

tests/test_bgp_neighbor_af.rb
```diff
@@ -30,13 +30,20 @@ class TestRouterBgpNeighborAF < CiscoTestCase
   def setup
     super
     if @@reset_feat
-      config('no feature bgp', 'feature bgp')
-      config('no nv overlay evpn', 'nv overlay evpn')
+      if platform == :nexus
+        config('no feature bgp', 'feature bgp')
+        config('no nv overlay evpn', 'nv overlay evpn')
```
glennmatthews: Why not move these configs into cleanup_ios_xr and setup_ios_xr?

tests/test_bgp_neighbor_af.rb
```diff
@@ -56,6 +63,67 @@ def clean_af(af_args, ebgp=true)
...
+    # Global requirements
+    # route policies
+    str1 = 'route-map-in-name'
+    str2 = 'route-map-out-name'
+    config("route-policy #{str1}", 'end-policy')
+    config("route-policy #{str2}", 'end-policy')
+    config("route-policy #{str1.reverse}", 'end-policy')
+    config("route-policy #{str2.reverse}", 'end-policy')
+    config('route-policy foo_bar', 'end-policy')
+    config('route-policy baz_inga', 'end-policy')
```
glennmatthews, line 83: Would it be reasonable to merge these configs together into a single 'config()' call?

tests/test_bgp_neighbor_af.rb
```diff
@@ -117,6 +184,16 @@ def test_nbrs_with_masks
...
       asn, vrf, nbr, af = v
       nbr += (nbr[/:/]) ? '/64' : '/16'
+      if platform == :ios_xr
+        # XR does not do slash masks and neighbors
+        # so make sure to catch the error
+        assert_raises(ArgumentError, 'Neighbors do not support /') do
+          RouterBgpNeighborAF.new(asn, vrf, nbr, af, true)
+        end
+        # Finish the test case with regular neighbor
+        # to avoid skipping whole testcase
```
glennmatthews: This seems dubious - this test case is explicitly `test_nbrs_with_masks` - doesn't look like it does anything that isn't already covered by `test_nbr_af_create_destroy` above and I think proceeding with a non-mask neighbor is not in the spirit of the test. Would be better to 'return' after the above assertion.

tests/test_bgp_neighbor_af.rb
```diff
@@ -233,7 +322,7 @@ def test_tri_states
     @@matrix.values.each do |af_args|
       af, dbg = clean_af(af_args)
 
-      unless dbg.include?('l2vpn/evpn')
+      unless dbg.include?('l2vpn/evpn') || platform == :ios_xr
```
glennmatthews: If these keywords aren't supported on XR, please add appropriate negative tests.

tests/test_bgp_neighbor_af.rb
```diff
@@ -267,6 +356,12 @@ def test_advertise_map
   def advertise_map(af, dbg)
     %w(advertise_map_exist advertise_map_non_exist).each do |k|
       v = %w(foo bar)
+      if platform == :ios_xr
+        assert_raises(ArgumentError) do
```
glennmatthews: As above, I would really expect the default behavior (getter and default return `nil` and setter raises an `UnsupportedError`) rather than this ArgumentError.

lib/cisco_node_utils/cmd_ref/bgp_neighbor_af.yaml
```diff
@@ -94,27 +109,33 @@ max_prefix_threshold:
   default_value: ~
 
 max_prefix_warning:
-  default_value: ~
+  cli_ios_xr:
+    default_value: 75
```
glennmatthews, line 111: This seems very weird given that the max_prefix_warning getter for NXOS returns either `true` or `nil` - is it really correct that `warning-only` means `75` on XR?

(edit) Is this supposed to be for max_prefix_threshold instead?

tests/test_bgp_neighbor_af.rb
```diff
@@ -552,6 +670,60 @@ def send_community(af, dbg)
...
+    v = 'extended'
+    af.send_community = v
+    assert_equal('send-extended-community-ebgp', af.send_community,
+                 "Test 2a. #{dbg} Failed to set '#{v}' from 'both'")
+    af.send_community = v
+    assert_equal('send-extended-community-ebgp', af.send_community,
+                 "Test 2b. #{dbg} Failed to set '#{v}' from 'extended'")
+    v = 'standard'
+    af.send_community = v
+    # XR it's additive
```
glennmatthews: This seems suspect to me and likely to break idempotence. If I ask node-utils to set send_community = 'standard', I'd expect send_community to report as 'standard' (and this is the behavior on Nexus, right?), not as 'both'.

tests/test_bgp_neighbor_af.rb
```diff
@@ -572,29 +744,61 @@ def test_soo
...
+      assert_raises(Cisco::UnsupportedError) do
         af.soo = val
       end
-      # SOO is only allowed in non-default VRF
-      return
     end
+  end
+
+  # --------------------------------
+  def test_weight
```
glennmatthews, line 738: Was there not a test for 'weight' previously?

lib/cisco_node_utils/bgp_neighbor_af.rb
RichWellum: Ack

lib/cisco_node_utils/bgp_neighbor_af.rb
RichWellum: Good point - I forget that. I'll remove them and check it's working.

RichWellum: I see why I did this now, these are mainly for the 'default' properties - where the yaml isn't _excluded. I'll rework this.

lib/cisco_node_utils/bgp_neighbor_af.rb
RichWellum: Yes that's better.

lib/cisco_node_utils/bgp_neighbor_af.rb
RichWellum: Ack.

lib/cisco_node_utils/bgp_neighbor_af.rb
RichWellum, line 583: It seems to be yeah - I know I just got lucky...

I added the following TC to my minitest to verify:

none - extended - standard, then none - standard - extended.

getter/setters worked fine.

lib/cisco_node_utils/bgp_neighbor_af.rb
RichWellum: Original code which I copied - but it does seem unnecessary. I'll check with Chris.

RichWellum: Chris recalls it as being because the regexp was neater in the code, especially in the switch statement.

lib/cisco_node_utils/cmd_ref/bgp_neighbor_af.yaml
RichWellum, line 111: yeah almost certainly - and that's what I am testing in mini test.

lib/cisco_node_utils/cmd_ref/bgp_neighbor_af.yaml
RichWellum: Ack.

tests/test_bgp_neighbor_af.rb
RichWellum: Because those methods are for xr only and the configs are nexus?

tests/test_bgp_neighbor_af.rb
RichWellum, line 83: TBh I don't think it adds much and makes it a little less readable imo.

tests/test_bgp_neighbor_af.rb
RichWellum: Fair enough. BTW when I add a 'return' rubocop tells me:

Non-local exit from iterator, without return value. `next`, `break`, `Array#find`, `Array#any?`, etc. is preferred. [Lint/NonLocalExitFromIterator] [2 times]

This is probably why I elected to add a dummy address and just let it run through. Any idea what I can use instead? break / next doesn't work in this context.

glennmatthews: Seems like `next` would be exactly the right thing to do. No?

tests/test_bgp_neighbor_af.rb
RichWellum: Yup done.

tests/test_bgp_neighbor_af.rb
RichWellum: Yup done.

tests/test_bgp_neighbor_af.rb
RichWellum, line 738: There was not - checked with Chris - thinks it was just missed.

tests/test_bgp_neighbor_af.rb
RichWellum: Yes - got it to work.

tests/test_bgp_neighbor_af.rb
RichWellum: In Nexus the behavior is:
standard or extended or both

So if you have extended then do standard it replaces the extended (not additive).

In XR the behavior is
standard or extended or standard and extended.

So when you do extended after a standard, you get both configured which is equivalent of 'both'.

So I think this is correct - but not 100% sure about the nexus behavior.

This correct @chrisvanheuveln ?

tests/test_bgp_neighbor_af.rb
RichWellum, line 738: That correct @chrisvanheuveln  ?

lib/cisco_node_utils/cmd_ref/bgp_neighbor_af.yaml
RichWellum, line 111: You were correct. Fixed.

tests/test_bgp_neighbor_af.rb
RichWellum: @glennmatthews Addressed all your comments with this new pull, except this one here which I think is the correct behavior.

Rubocop and minittest on both platforms is happy.

chrisvanheuveln: Not quite right. NU should set the state to the manifest's desired state, so if the manifest says 'standard' then it needs to set to 'standard'. 

In the (manual) nexus case, if the switch state is currently 'extended', then setting the cli to standard would result in a 'both' state, but that's not what the manifest wants, so NU needs to first remove the 'extended' before setting it to 'standard' to avoid the 'both' state.

RichWellum: Ack got it. Subtle difference between the CLI and what a puppet manifest wants. Will rework.

lib/cisco_node_utils/bgp_neighbor_af.rb
```diff
@@ -213,6 +218,7 @@ def additional_paths_receive
     end
 
     def additional_paths_receive=(val)
+      # return if val.nil?
```
glennmatthews: remove?

lib/cisco_node_utils/bgp_neighbor_af.rb
```diff
@@ -223,7 +229,8 @@ def additional_paths_receive=(val)
     end
 
     def default_additional_paths_receive
-      config_get_default('bgp_neighbor_af', 'additional_paths_receive').to_sym
+      ret = config_get_default('bgp_neighbor_af', 'additional_paths_receive')
+      ret.to_sym unless ret.to_s == ''
```
glennmatthews: Would this be more correct?

`ret.to_sym unless ret.nil?`

glennmatthews: Also, it shouldn't be too hard to extend CommandReference to let the YAML specify `kind: symbol` and have CmdRef automatically do this conversion.

lib/cisco_node_utils/bgp_neighbor_af.rb
```diff
@@ -544,23 +557,49 @@ def default_route_reflector_client
...
     def send_community
       val = config_get('bgp_neighbor_af', 'send_community', @get_args)
       return default_send_community if val.nil?
+      platform == :nexus ? send_comm_nexus_get(val) : send_comm_ios_xr_get(val)
+    end
+
+    # Nexus: <state> send-community [ both | extended | standard ]
+    #  NOTE: 'standard' is default but does not nvgen on some platforms
+    #  Returns: none, both, extended, or standard
+    def send_comm_nexus_get(val)
```
glennmatthews: Since the 'get' work is done in `send_community`, maybe rename these methods to `send_community_nexus` and `send_community_ios_xr`?

lib/cisco_node_utils/bgp_neighbor_af.rb
```diff
@@ -544,23 +557,49 @@ def default_route_reflector_client
...
+    #  the equivalents of the NXOS: standard | extended functionality
+    #  Returns: node, 'send-community-ebgp', 'send-extended-community-ebgp' or
+    #  'send-community-ebgp send-extended-community-ebgp' which is the 'both'
+    # keyword equivalent
+    def send_comm_ios_xr_get(val)
+      if val == ['send-community-ebgp', 'send-extended-community-ebgp']
+        val = 'both'
+      else
+        # Properties are an array
+        val = val.join
```
glennmatthews: This seems wrong - XR and Nexus need to return the same values. Don't you need to convert `'send-community-ebgp'` to `'standard'` and `'send-extended-community-ebgp'` to `'extended'`?

lib/cisco_node_utils/bgp_neighbor_af.rb
```diff
@@ -579,6 +618,47 @@ def send_community=(val)
...
+          if val == 'standard'
+            set_args_keys(state: 'no', attr: 'send-community-ebgp')
+          else # val == 'extended'
+            set_args_keys(state: 'no', attr: 'send-extended-community-ebgp')
+          end
+          config_set('bgp_neighbor_af', 'send_community', @set_args)
+          state = ''
+        end
+      end
+      if val == 'both'
```
glennmatthews: I find the flow of this function quite confusing due to the intermingling of if, else, when. Would the following work instead?

``` ruby
def send_comm_ios_xr_set(val)
  case val
  when 'none'
    set_args_keys(state: 'no', attr: 'send-community-ebgp')
    config_set('bgp_neighbor_af', 'send_community', @set_args')
    set_args_keys(state: 'no', attr: 'send-extended-community-ebgp')
    config_set('bgp_neighbor_af', 'send_community', @set_args')
  when 'standard'
    set_args_keys(state: '', attr: 'send-community-ebgp')
    config_set('bgp_neighbor_af', 'send_community', @set_args')
    set_args_keys(state: 'no', attr: 'send-extended-community-ebgp')
    config_set('bgp_neighbor_af', 'send_community', @set_args')
  when 'extended'
    set_args_keys(state: 'no', attr: 'send-community-ebgp')
    config_set('bgp_neighbor_af', 'send_community', @set_args')
    set_args_keys(state: '', attr: 'send-extended-community-ebgp')
    config_set('bgp_neighbor_af', 'send_community', @set_args')
  when 'both'
    set_args_keys(state: '', attr: 'send-community-ebgp')
    config_set('bgp_neighbor_af', 'send_community', @set_args')
    set_args_keys(state: '', attr: 'send-extended-community-ebgp')
    config_set('bgp_neighbor_af', 'send_community', @set_args')
  end
end
```

tests/test_bgp_neighbor_af.rb
glennmatthews: Sure - I was wondering why not make the methods applicable to both platforms rather than putting Nexus configuration locally and XR configuration in a method.

tests/test_bgp_neighbor_af.rb
```diff
@@ -64,15 +132,15 @@ def clean_af(af_args, ebgp=true)
...
   @@matrix = { # rubocop:disable Style/ClassVars
     # 1  => [1, 'default', '10:1::1', %w(ipv4 multicast)], # UNSUPPORTED
-    # 2  => [1, 'default', '10:1::1', %w(ipv4 unicast)],
+    # 2  => [1, 'default', '10:1::1', %w(ipv4 unicast)], # UNSUPPORTED in XR
     # 3  => [1, 'default', '10:1::1', %w(ipv6 multicast)],
     # 4  => [1, 'default', '10:1::1', %w(ipv6 unicast)],
     # 5  => [1, 'default', '1.1.1.1', %w(ipv4 multicast)],
     6  => [1, 'default', '1.1.1.1', %w(ipv4 unicast)],
     # 7  => [1, 'default', '1.1.1.1', %w(ipv6 multicast)],
-    8  => [1, 'default', '1.1.1.1', %w(ipv6 unicast)],
```
glennmatthews, line 73: Is this intended - you are no longer including ipv6 unicast in the active matrix?

tests/test_bgp_neighbor_af.rb
```diff
@@ -117,6 +184,15 @@ def test_nbrs_with_masks
     @@matrix.each do |k, v|
       asn, vrf, nbr, af = v
       nbr += (nbr[/:/]) ? '/64' : '/16'
+      if platform == :ios_xr
+        # XR does not do slash masks and neighbors
+        # so make sure to catch the error
+        assert_raises(ArgumentError, 'Neighbors do not support /') do
```
glennmatthews: This is no longer correct since you changed it to raise an UnsupportedError, right?

tests/test_bgp_neighbor_af.rb
```diff
@@ -117,6 +184,15 @@ def test_nbrs_with_masks
...
+          RouterBgpNeighborAF.new(asn, vrf, nbr, af, true)
+        end
+        next
+      end
+      next if platform == :ios_xr
```
glennmatthews, line 198: redundant?

tests/test_bgp_neighbor_af.rb
```diff
@@ -150,15 +233,17 @@ def test_props_bool
...
       :next_hop_self,
-      :next_hop_third_party,
-      :suppress_inactive,
     ]
 
-    props = [:disable_peer_as_check] if dbg.include?('l2vpn/evpn')
+    if platform == :nexus
+      props << :next_hop_third_party
+      props << :suppress_inactive
+      props = [:disable_peer_as_check] if dbg.include?('l2vpn/evpn')
```
glennmatthews, line 252: By moving this inside here, Nexus overrides props with `[:disable_peer_as_check]` when testing l2vpn/evpn, but XR uses the base array. Is that right?

tests/test_bgp_neighbor_af.rb
```diff
@@ -230,9 +319,36 @@ def props_string(af, dbg)
...
+        [:enable, :disable, :inherit, 'enable', 'disable', 'inherit'
+        ].each do |val|
+          assert_raises(UnsupportedError,
+                        "This feature #{k}, #{val} is not supported}") do
+            af.send("#{k}=", val)
+          end
+        end
+      end
+
+      %w(soft_reconfiguration_in).each do |k|
```
glennmatthews, line 345: This block is the same for XR and NX right? Can you refactor to avoid duplicate code?

tests/test_bgp_neighbor_af.rb
```diff
@@ -267,6 +383,12 @@ def test_advertise_map
   def advertise_map(af, dbg)
     %w(advertise_map_exist advertise_map_non_exist).each do |k|
       v = %w(foo bar)
+      if platform == :ios_xr
+        assert_raises(UnsupportedError) do
+          af.send("#{k}=", v)
+        end
+        break
```
glennmatthews: not `next`?

tests/test_bgp_neighbor_af.rb
```diff
@@ -357,14 +491,28 @@ def default_originate(af, dbg)
...
+    end
 
     # Test default_state
-    af.default_originate_set(af.default_default_originate)
-    refute(af.default_originate,
-           "Test 6. #{dbg} Failed to set state to default")
+    if platform == :ios_xr
+      # XR does not support a default route-policy
+      assert_raises(ArgumentError) do
+        af.default_originate_set(af.default_default_originate)
```
glennmatthews: This seems like it should work on XR - this is just a 'no default_originate' right?

tests/test_bgp_neighbor_af.rb
```diff
@@ -407,7 +555,7 @@ def max_prefix(af, dbg)
     af.max_prefix_set(limit, threshold, warning)
     assert_equal(limit, af.max_prefix_limit,
                  "Test 4a. #{dbg} Failed to set limit to '#{limit}'")
-    assert_equal(threshold, af.max_prefix_threshold,
+    assert_equal((platform == :nexus ? threshold : 75), af.max_prefix_threshold,
```
glennmatthews: Is there no way to set 'no threshold' on XR?

tests/test_bgp_neighbor_af.rb
```diff
@@ -552,9 +697,95 @@ def send_community(af, dbg)
...
+    v = 'both'
+    af.send_community = v
+    assert_equal(v, af.send_community,
+                 "Test 1a. #{dbg} Failed to set '#{v}' from None")
+    af.send_community = v
+    assert_equal(v, af.send_community,
+                 "Test 1b. #{dbg} Failed to set '#{v}' from None")
+    v = 'extended'
+    af.send_community = v
+    assert_equal('send-extended-community-ebgp', af.send_community,
```
glennmatthews: As commented above in the node_util class - the getter needs to match the setter for idempotence to work. If you set 'extended', the getter should return 'extended' not 'send-extended-community-ebgp'.

lib/cisco_node_utils/bgp_neighbor_af.rb
RichWellum: Ack.

lib/cisco_node_utils/bgp_neighbor_af.rb
RichWellum: I shortened it so I could get the terniary on one line. You ok with:

```
  platform == :nexus ? send_community_nexus(val) : send_community_xr(val)
```

?

lib/cisco_node_utils/bgp_neighbor_af.rb
RichWellum: So I've learned something here I didn't know before (idempotency etc). If I do this, then because the getter is basically the same it means I can share a lot more code between the two platforms - especially the minitest code.

Without this change I was converting ['send-community-ebgp', 'send-extended-community-ebgp'] to 'both' in the setter - but no need now.

Tested - seems to work better.

lib/cisco_node_utils/bgp_neighbor_af.rb
RichWellum: Much better thanks!

tests/test_bgp_neighbor_af.rb
RichWellum: Yeah makes sense - will do this.

tests/test_bgp_neighbor_af.rb
RichWellum, line 73: Yeah I felt it wasn't adding much other than another whole cycle of running through the tests which are already pretty slow. So I went with: ipv4 uni default vrf, same with a real vrf and the l2vpn case. I'll put it back in if you want?

RichWellum, line 73: I'll add back in but note I've also added the vrf case in too.

tests/test_bgp_neighbor_af.rb
RichWellum: Yeah fixed. The raise argumenterror was working because it was catching a different argument error.

tests/test_bgp_neighbor_af.rb
RichWellum, line 198: Ack.

tests/test_bgp_neighbor_af.rb
RichWellum, line 252: That's correct yes.

tests/test_bgp_neighbor_af.rb
RichWellum, line 345: it's the same apart from I raise an error and don't do an assert. I tried for a long time. I always ended up with some issue - as it's test code I decided to split it out.

tests/test_bgp_neighbor_af.rb
RichWellum: Yup.

tests/test_bgp_neighbor_af.rb
RichWellum: It's not allowed in XR, in XR you need to explicitly name the route-policy because you can have more than one...

tests/test_bgp_neighbor_af.rb
RichWellum: There is and it works, but this is testing threshold = nil. On XR it appears that the default is 75 though. So this is testing that.

max_prefix_threshold:
  cli_ios_xr:
    default_value: 75
  cli_nexus:
    default_value: ~

tests/test_bgp_neighbor_af.rb
RichWellum: Yes with your suggestions I was able to dump the XR specific testing and use the nexus test code.

lib/cisco_node_utils/bgp_neighbor_af.rb
RichWellum: I thought I had a nil vs '' issue and was following: http://stackoverflow.com/questions/247948/is-there-a-better-way-of-checking-nil-or-length-0-of-a-string-in-ruby

But I replaced it with nil? and it's passing so not sure what I was seeing.

RichWellum: Made the command_reference change as suggested too.

tests/test_bgp_neighbor_af.rb
glennmatthews: I would have thought that threshold = nil would be the way to set 'no threshold'. If 'nil' means '75', what means 'none'?

tests/test_bgp_neighbor_af.rb
glennmatthews: Ok, but how do you revert to the default behavior of default_originate being disabled?

RichWellum: Well I looked up the command and see that I am completely wrong about the functionality. You CAN do default-originate without a policy-map, which means 'no should work. So reworking this code. I think my getter was wrong and gave me the wrong impression that it wasn't supported.

RichWellum: Yup got it working now.

tests/test_bgp_neighbor_af.rb
chrisvanheuveln: @glennmatthews It looks like threshold is not optional on XR (as it is on NX), 
or rather XR will nvgen the default threshold, whereas NX appears to have no (known) default threshold (checking against 'show run bgp all' and 'show ip bgp neigh x.x.x.x').

Setting the yaml default_value to 75 for XR should be fine, however the test should continue to work for either platform without modification.

lib/cisco_node_utils/bgp_neighbor_af.rb
```diff
@@ -70,6 +70,13 @@ def validate_args(asn, vrf, nbr, af)
       fail ArgumentError, "'af' must be an array specifying afi and safi" unless
         af.is_a?(Array) || af.length == 2
 
+      # XR BGP does not support <address>/<mask>
```
RichWellum, line 73: @glennmatthews Addressed your last round of comments. Rubocop and mini test happy.

tests/test_bgp_neighbor_af.rb
glennmatthews, line 252: Just to be clear:

```
props = [:as_override, :next_hop_self] # XR, not l2vpn/evpn
props = [:as_override, :next_hop_self] # XR, l2vpn/evpn
props = [:as_override, :next_hop_self, :next_hop_third_party, :suppress_inactive] # Nexus, not l2vpn/evpn
props = [:disable_peer_as_check] # Nexus, l2vpn/evpn
```

So it's intentional that Nexus does _not_ test the other properties under l2vpn, but _does_ test only `:disable_peer_as_check`; and conversely, that XR _does_ test the other properties (which Nexus doesn't) under l2vpn but does _not_ test `:disable_peer_as_check`?

RichWellum, line 252: Yes exactly, I tested it pretty thoroughly. For XR the only properties supported are:
 :as_override,
 :next_hop_self,

Under all af's.

For Nexus they are all supported, except when a l2vpn in which case the only disable_peer_as_check is supported.

glennmatthews, line 252: OK. Consider adding a comment and/or negative tests as this is not obvious just looking at the code.

glennmatthews:
:+1:

tests/test_bgp_neighbor_af.rb
RichWellum, line 252: Sure.

PR #121, opened 2015-12-08
review and pull request for cisco acl remark

bansalpradeep:


chrisvanheuveln:
Pradeep, sorry I don't have the background on this one. Was there a reason why it had to be a separate standalone provider instead of co-existing with ace.rb ?

The ace entries for remark look like they should fit right in with the other aces:

```
ip access-list foo
  10 remark this is a remark
  20 remark this is a remark2
```

I'll sync up with Mike and see if there's more to this than it seems.

chrisvanheuveln:
As discussed with @mikewiebe and @bansalpradeep, this should just be another property under ace.rb instead of being its own provider. Pradeep with rework this accordingly.

PR #125, opened 2015-12-10
Merge cisco_os_shim content into cisco_node_utils

glennmatthews:
To reduce Gem dependency hell, migrate the contents of cisco_os_shim (https://github.com/cisco/cisco-nxapi/tree/develop) into the cisco_node_utils package.

The abstract client code is under `lib/cisco_node_utils/client/` and the NXAPI and gRPC client implementations are in `lib/cisco_node_utils/client/nxapi` and `lib/cisco_node_utils/grpc` respectively. 

`CiscoLogger` now lives in `lib/cisco_node_utils/logger.rb` and is renamed to `Cisco::Logger`.

I simplified the nested namespaces for the client classes a bit - some examples of the new namespacing:

``` ruby
Cisco::Client.create(...)
Cisco::Client::NXAPI.new()
Cisco::Client::GRPC.new()
Cisco::Client::ConnectionRefused
Cisco::Client::NXAPI::CliError
```

RSpec tests pass. I ran the `test_nxapi`, `test_grpc`, `test_logger` minitests migrated from cisco_os_shim and they are passing. I ran a few standard minitests (`test_interface`, `test_bgp`) and they are working as well. Travis-CI should be passing now.


robert-w-gries:
Why are you merging into `feature/grpc` instead of `develop`?

glennmatthews:
Because this is still grpc-specific content. develop doesn't have any grpc code yet

robert-w-gries:
Since our `cisco-nxapi/feature/grpc` branch was collapsed into `cisco-nxapi/develop`, I don't see the problem with merging our `cisco-nxapi/develop` code into `node-utils/develop`.

The nexus side is already using this code in the `cisco_os_shim` gem.

glennmatthews:
node-utils/develop is still using cisco_nxapi, not cisco_os_shim.

robert-w-gries:
:+1:

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ Changelog
 ### Added
 
 * `Cisco::UnsupportedError` exception class, raised when a command is explicitly marked as unsupported on a particular class of nodes.
+* Added client support for gRPC on IOS XR 6.0.0.
+* Smart dependency installation - installing this gem will install `grpc` on IOS XR and Linux environments, but not on NX-OS environments.
```
mikewiebe: This will conversely install nxapi on NX-OS right?

mikewiebe:
@glennmatthews Can you identify specific areas you had to refactor for the merge vs. code that was simply moved from one place to another?  Want to focus review on those areas rater then just re-review code that has already been reviewed given the size of these diffs.

bin/git/hooks/pre-commit/validate-diffs
```diff
@@ -25,7 +25,3 @@ fi
 git diff --cached --name-only --diff-filter=A -z $against |
         LC_ALL=C tr -d '[ -~]\0' | wc -c
 check_rc "Rename non-ASCII file name(s) before committing"
-
-# If there are whitespace errors, print the offending file names and fail.
```
glennmatthews, line 29: Removed this because the NXAPI and gRPC tests include trailing whitespace as part of the expected literal output from the node. Rubocop will catch trailing whitespace in Ruby code so this doesn't give us a lot of risk.

lib/cisco_node_utils/bgp.rb
```diff
@@ -17,6 +17,7 @@
 require_relative 'cisco_cmn_utils'
 require_relative 'node_util'
 require_relative 'bgp_af'
+require_relative 'logger'
```
glennmatthews, line 20: (new code in this branch) I don't know if this is explicitly needed but it doesn't hurt.

glennmatthews:
I added a couple comments inline - but in general, the first commit (9715d84) is purely moving files in from the other repo with the minimum changes needed to get it working, the second commit (6534263) is pretty self-contained, and the more meaty refactorings are in the later commits (eadf8c8, 257b717). You may find it useful to view the individual commits for review purposes.

glennmatthews:
Forgot to mention - I tested the grpc behavior change (the latest commit) against both 'old' XR code and 'new' XR code and it works in both cases - it's backwards compatible with the old code.

.gitignore
```diff
@@ -2,6 +2,7 @@
 pkg/*
```
RichWellum, line 2: :+1: Thanks for doing this - should make life a lot easier.

mikewiebe:
:+1:

PR #129, opened 2015-12-14
Added SNMP Notification Receiver

jonnytdevops:
This is currently failing rubocop on the following:

```
lib/cisco_node_utils/snmp_notification_receiver.rb:24:5: C: Assignment Branch Condition size for initialize is too high. [41.61/40]
    def initialize(name,
    ^^^
lib/cisco_node_utils/snmp_notification_receiver.rb:24:5: C: Cyclomatic complexity for initialize is too high. [20/17]
    def initialize(name,
    ^^^
lib/cisco_node_utils/snmp_notification_receiver.rb:24:5: C: Perceived complexity for initialize is too high. [20/19]
```

However, this needs to be changed in the upstream rubocop code. My current code for this class is appropriate for the task, and is in line with other the other classes that has been merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,8 @@ Changelog
 * RADIUS
   * radius_global (@jonnytpuppet)
   * radius_server (@jonnytpuppet)
+* SNMP
+  * snmp_notification_receiver (@jonnytpuppet)
 * SYSLOG
   * syslog_server (@jonnytpuppet)
   * syslog_setting (@jonnytpuppet)
--- a/lib/cisco_node_utils/cmd_ref/snmp_notification_receiver.yaml
+++ b/lib/cisco_node_utils/cmd_ref/snmp_notification_receiver.yaml
@@ -1,0 +1,50 @@
+# snmp_notification_receiver
+---
+port:
+  config_get: "show running-config snmp all"
+  config_get_token: '/^snmp-server host %s.*version.* udp-port (\d+).*$/'
+  default_value: null
+
+receivers:
+  config_get: "show running snmp all"
+  config_get_token: '/^snmp-server host (\S+) (traps|informs)/'
+  config_set: "<state> snmp-server host <ip> <type> version <version> <security> <username> <udp_port>"
+  multiple:
+
+security:
+  config_get: "show running-config snmp all"
+  config_get_token: '/^snmp-server host %s.* version 3 (auth|noauth|priv).*$/'
+  default_value: null
+
+source_interface:
+  config_get: "show running-config snmp all"
+  config_get_token: '/^snmp-server host %s.* source-interface (\S+).*$/'
+  config_set: "snmp-server host <ip> source-interface <source_interface> <port>"
+  default_value: null
+
+type:
+  config_get: "show running-config snmp all"
+  config_get_token: '/^snmp-server host %s (traps|informs).*$/'
+  default_value: null
+
+username:
+  config_get: "show running-config snmp all"
+  config_get_token: '/^snmp-server host %s.*version.* (\S+)$/'
+  default_value: null
+
+username_with_port:
+  config_get: "show running-config snmp all"
+  config_get_token: '/^snmp-server host %s.*version.* (\S+) udp-port/'
+  default_value: null
+
+version:
+  config_get: "show running-config snmp all"
+  config_get_token: '/^snmp-server host %s.* version (\S+).*$/'
+  default_value: null
+
+vrf:
+  config_get: "show running-config snmp all"
+  config_get_token: '/^snmp-server host %s.* use-vrf (\S+).*$/'
+  config_set: "snmp-server host <ip> use-vrf <vrf> <port>"
+  default_value: null
+
--- a/lib/cisco_node_utils/snmp_notification_receiver.rb
+++ b/lib/cisco_node_utils/snmp_notification_receiver.rb
@@ -1,0 +1,158 @@
+# November 2015, Jonathan Tripathy
+#
+# Copyright (c) 2014-2015 Cisco and/or its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative 'node_util'
+
+module Cisco
+  # SnmpNotificationReceiver - node utility class for SNMP server management
+  class SnmpNotificationReceiver < NodeUtil
+    attr_reader :name
+
+    def initialize(name,
+                   instantiate:      true,
+                   type:             '',
+                   version:          '',
+                   security:         '',
+                   username:         '',
+                   port:             '',
+                   vrf:              '',
+                   source_interface: '')
+
+      fail TypeError unless name.is_a?(String)
+      @name = name
+
+      fail TypeError unless type.is_a?(String)
+
+      fail TypeError unless version.is_a?(String) || version.is_a?(Integer)
+
+      fail TypeError unless security.is_a?(String)
+
+      fail TypeError unless username.is_a?(String)
+
+      fail TypeError unless port.is_a?(String) || port.is_a?(Integer)
+
+      fail TypeError unless vrf.is_a?(String)
+
+      fail TypeError unless source_interface.is_a?(String)
+
+      return unless instantiate
+
+      # Mandatory Properties
+      fail TypeError unless name.length > 0
+      fail TypeError unless type.length > 0
+
+      if version.is_a?(Integer)
+        fail TypeError if version <= 0
+      else
+        fail TypeError if version.length <= 0
+      end
+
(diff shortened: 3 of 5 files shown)


glennmatthews:
You can change lib/.rubocop.yml to add the appropriate new upper bounds for these metrics as part of this PR.

Alternately, consider splitting the argument validation into a separate validation function - see snmpuser.rb for an example.

lib/cisco_node_utils/snmp_notification_receiver.rb
```diff
@@ -0,0 +1,158 @@
...
+
+    def initialize(name,
+                   instantiate=true,
+                   type='',
+                   version='',
+                   security='',
+                   username='',
+                   port='',
+                   vrf='',
+                   source_interface='')
```
glennmatthews: Consider using named args instead:

``` ruby
def initialize(name,
               instantiate: true, type: '', version: '', security: '', username: '',
               port: '', vrf: '', source_interface: '')
```

This will make the instantiation calls easier to read and less error prone:

``` ruby
    receiver = Cisco::SnmpNotificationReceiver.new(
      id,
      instantiate:      true,
      type:             'informs',
      version:          '3',
      security:         'priv',
      username:         'ab',
      port:             '45',
      vrf:              'red',
      source_interface: 'ethernet1/3')
```

lib/cisco_node_utils/snmp_notification_receiver.rb
```diff
@@ -0,0 +1,158 @@
...
+
+      fail TypeError unless name.is_a?(String)
+      @name = name
+
+      fail TypeError unless type.is_a?(String)
+
+      fail TypeError unless version.is_a?(String)
```
glennmatthews: so `version: 3` would be rejected as invalid? Seems unfortunate.

lib/cisco_node_utils/snmp_notification_receiver.rb
```diff
@@ -0,0 +1,158 @@
...
+
+      fail TypeError unless security.is_a?(String)
+
+      fail TypeError unless username.is_a?(String)
+
+      fail TypeError unless port.is_a?(String)
```
glennmatthews: ditto - why not port as an integer?

lib/cisco_node_utils/snmp_notification_receiver.rb
```diff
@@ -0,0 +1,158 @@
...
+      fail TypeError unless name.length > 0
+      fail TypeError unless type.length > 0
+      fail TypeError unless version.length > 0
+      fail TypeError unless username.length > 0
+
+      config_set('snmp_notification_receiver',
+                 'receivers',
+                 state:    '',
+                 ip:       "#{name}",
+                 type:     "#{type}",
```
glennmatthews: if you're forcing these to be strings to begin with, why the extra quoting? Why not just `ip: name, type: type`?

lib/cisco_node_utils/snmp_notification_receiver.rb
```diff
@@ -0,0 +1,158 @@
...
+                 version:  "#{version}",
+                 security: security.empty? ? '' : "#{security}",
```
glennmatthews: given that security is a string, this is literally equivalent to `security: security`

tests/test_snmp_notification_receiver.rb
```diff
@@ -0,0 +1,134 @@
...
+    assert_equal(Cisco::SnmpNotificationReceiver.receivers[id2].version, '3')
+    assert_equal(Cisco::SnmpNotificationReceiver.receivers[id2].vrf, 'red')
+    assert_equal(Cisco::SnmpNotificationReceiver.receivers[id2].security,
+                 'auth')
+
+    receiver.destroy
+    receiver2.destroy
+    refute_includes(Cisco::SnmpNotificationReceiver.receivers, id)
+    refute_includes(Cisco::SnmpNotificationReceiver.receivers, id2)
+  end
```
glennmatthews, line 136: Is there no way to change any of the properties without doing a destroy and recreate? If I want to change the username for a receiver from 'ab' to 'cd', I have to call `instance.destroy()` and `instance.new()` rather than something like `instance.username = 'cd'`?

lib/cisco_node_utils/snmp_notification_receiver.rb
jonnytdevops: Everything is a string in the Cisco CLI. I can port to a Integer though if you would like. It just means I would have to cast in the Puppet provider and create a separate munge method for integers.

tests/test_snmp_notification_receiver.rb
jonnytdevops, line 136: This is a limitation in the Cisco CLI that does not allow you to change properties independently of each other. This was discussed in the an email thread with the Cisco SNMP guys for the snmp_user provider.

glennmatthews, line 136: Fair enough -  snmpuser does the same approach but I was mis-remembering that it had setters for individual properties too. This is fine.

lib/cisco_node_utils/snmp_notification_receiver.rb
glennmatthews: It's fine to accept strings as valid input still - I just don't want to reject integers unnecessarily.

tests/test_snmp_notification_receiver.rb
ajjohnsonpdx, line 136: Hi @mikewiebe - Is there anything you're needing to merge this PR?

mikewiebe, line 136: Hi @ajjohnsonpdx @glennmatthews provided quite a few review comments which I believe are not yet resolved?

jonnytdevops:
@glennmatthews @mikewiebe Can this please be merged, the code has been updated, thanks

CHANGELOG.md
```diff
@@ -44,6 +44,8 @@ Changelog
 * RADIUS
   * radius_global (@jonnytpuppet)
   * radius_server (@jonnytpuppet)
+*SNMP
```
glennmatthews: missing space between `*` and `SNMP` is causing the markdown to render this as italics instead of a bullet list - please fix.

lib/cisco_node_utils/cmd_ref/snmp_notification_receiver.yaml
```diff
@@ -0,0 +1,50 @@
+# snmp_notification_receiver
+---
+port:
+  config_get: "show running-config snmp all"
+  config_get_token: '/^snmp-server host %s.*version.* udp-port (\d+).*$/'
```
glennmatthews, line 5: For future work it would be preferable to use named keys in the config_get_token as well (`'/^snmp-server host <ip>.*version.* udp-port (\d+).*$/'`) but this isn't critical at this time.

lib/cisco_node_utils/snmp_notification_receiver.rb
```diff
@@ -0,0 +1,158 @@
...
+
+      fail TypeError unless type.is_a?(String)
+
+      fail TypeError unless version.is_a?(String) || version.is_a?(Integer)
+
+      fail TypeError unless security.is_a?(String)
+
+      fail TypeError unless username.is_a?(String)
+
+      fail TypeError unless port.is_a?(String) || version.is_a?(Integer)
```
glennmatthews: copy/paste error - should be `port.is_a?(Integer)`

lib/cisco_node_utils/snmp_notification_receiver.rb
```diff
@@ -0,0 +1,158 @@
...
+      fail TypeError unless vrf.is_a?(String)
+
+      fail TypeError unless source_interface.is_a?(String)
+
+      return unless instantiate
+
+      # Mandatory Properties
+      fail TypeError unless name.length > 0
+      fail TypeError unless type.length > 0
+      fail TypeError unless version.length > 0
```
glennmatthews: This will raise a NoMethodError if version is a Integer.

lib/cisco_node_utils/snmp_notification_receiver.rb
```diff
@@ -0,0 +1,158 @@
...
+        endpoint = 'username'
+      end
+
+      val = config_get('snmp_notification_receiver', endpoint, @name)
+      val
+    end
+
+    def version
+      val = config_get('snmp_notification_receiver', 'version', @name)
+      val
```
glennmatthews: Assign and return is unnecessary in all of these getters - they can be rewritten more succinctly:

``` ruby
def version
  config_get('snmp_notification_receiver', 'version', @name)
end
```

tests/test_snmp_notification_receiver.rb
```diff
@@ -0,0 +1,137 @@
...
+
+  def no_snmpnotificationreceiver
+    # Turn the feature off for a clean test.
+    config('no snmp-server host 4.5.6.7 informs version 3 priv ab udp-port 45',
+           'no snmp-server host 8.9.10.11 traps version 3 auth cd udp-port 46')
+  end
+
+  # TESTS
+
+  def test_snmpnotificationreceiver_create_destroy_single
```
glennmatthews: The convention we are trying to move towards is to not include the class name (`snmpnotificationreceiver`) in the test method name as it's redundant with the test class name (`TestSnmpNotificationReceiver::test_snmpnotificationreceiver_create_destroy_single`). Can you remove throughout?

jonnytdevops:
@glennmatthews @chrisvanheuveln @mikewiebe Happy New Year! I've made the requested changes. It would be appreciated if this could be merged. Thanks

glennmatthews:
:+1:

PR #130, opened 2015-12-15
Get all_vnis for 9k and destoy method destroys vlan-vni mapping

alok-aggarwal:
1. Added code to get all_vnis for n9k
2. Added code to destroy the vlan-vni mapping when destroy is called

Minitest passes:

sjc-ads-1546:262> ruby-2.1.1 tests/test_vni.rb -v -- 10.122.197.174 admin admin
Run options: -v -- --seed 23235
# Running:

Node under test:
- name  - agent-lab1-nx
- type  - N9K-NXOSV

TestVni#test_on_off = 7.40 s = .
TestVni#test_multiple_vnis_vlans = 12.25 s = .
TestVni#test_mapped_vlan = 5.34 s = .

Finished in 24.986780s, 0.1201 runs/s, 0.4002 assertions/s.

3 runs, 10 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for MiniTest to /nobackup/aloaggar/pnc/cisco-network-node-utils/coverage. 403 / 560 LOC (71.96%) covered.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/vni.yaml
+++ b/lib/cisco_node_utils/cmd_ref/vni.yaml
@@ -1,13 +1,13 @@
 # vni
 ---
 all_vnis:
-  config_get: "show vni"
-  config_get_token: '/^(\d+)\s/'
-
-all_vnis_n9k:
-  multiple: true
-  config_get: "show running vlan"
-  config_get_token: '/vn-segment (\d+)/'
+  /N7K/:
+    config_get: "show vni"
+    config_get_token: '/^(\d+)\s/'
+  else:
+    multiple: true
+    config_get: "show running vlan"
+    config_get_token: '/vn-segment (\d+)/'
 
 bridge_domain:
   config_get: "show vni"
@@ -22,21 +22,21 @@ create:
   config_set: ["vni %s" , "end"]
 
 destroy:
-  config_set: "no vni %s"
+  config_set: "no vni <vni>"
 
 encap_dot1q:
   config_set: ["encapsulation profile vni %s", "%s dot1q %s vni %s", "end"]
   default_value: ~
 
 feature:
-  config_get: "show feature"
-  config_get_token: '/^vni\s+\d+\s+(\S+)/'
-  config_set: "%s feature vni"
-
-feature_n9k:
-  config_get: " show running | i ^feature"
-  config_get_token: '/^feature vn-segment-vlan-based$/'
-  config_set: "<state> feature vn-segment-vlan-based"
+  /N7K/:
+    config_get: "show feature"
+    config_get_token: '/^vni\s+\d+\s+(\S+)/'
+    config_set: "<state> feature vni"
+  else:
+    config_get: " show running | i ^feature"
+    config_get_token: '/^feature vn-segment-vlan-based$/'
+    config_set: "<state> feature vn-segment-vlan-based"
 
 feature_nv_overlay:
   config_get: " show running | i ^feature"
--- a/lib/cisco_node_utils/cmd_ref/vni.yaml
+++ b/lib/cisco_node_utils/cmd_ref/vni.yaml
@@ -4,6 +4,11 @@ all_vnis:
   config_get: "show vni"
   config_get_token: '/^(\d+)\s/'
 
+all_vnis_n9k:
+  multiple: true
+  config_get: "show running vlan"
+  config_get_token: '/vn-segment (\d+)/'
+
 bridge_domain:
   config_get: "show vni"
   config_get_token: [ '/^%d\s+\S+\s+(\d+)/' ]
@@ -41,7 +46,7 @@ feature_nv_overlay:
 mapped_vlan:
   config_get: 'show running vlan'
   config_get_token: ['/^vlan <vlan>$/', '/^vn-segment (\d+)$/']
-  config_set: ['vlan <vlan>', '<state> vn-segment <vni>']
+  config_set: ['vlan <vlan>', '<state> vn-segment <vni> ; end']
   default_value: ""
 
 shutdown:
--- a/lib/cisco_node_utils/vni.rb
+++ b/lib/cisco_node_utils/vni.rb
@@ -32,9 +32,15 @@ module Cisco
 
     def self.vnis
       hash = {}
-      feature = config_get('vni', 'feature')
-      return hash if (:enabled != feature.to_sym)
-      vni_list = config_get('vni', 'all_vnis')
+      if /N9K/.match(node.product_id)
+        feature = config_get('vni', 'feature_n9k')
+        return hash if feature.nil?
+        vni_list = config_get('vni', 'all_vnis_n9k')
+      else
+        feature = config_get('vni', 'feature')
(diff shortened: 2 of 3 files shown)


lib/cisco_node_utils/cmd_ref/vni.yaml
```diff
@@ -4,6 +4,11 @@ all_vnis:
   config_get: "show vni"
   config_get_token: '/^(\d+)\s/'
 
+all_vnis_n9k:
```
chrisvanheuveln: Don't use a platform-specific tag, just update the existing `all_vnis` tag on line 3 to differentiate:

```
all_vnis:
  /N7K/:
    config_get: "show vni"           ### TBD: Refactor to use show run
    config_get_token: '/^(\d+)\s/'
  else:
    multiple:
    config_get: "show running vlan"
    config_get_token: '/vn-segment (\d+)/'
```

For the else case, should that use `multiple:` as well?

lib/cisco_node_utils/cmd_ref/vni.yaml
```diff
@@ -41,7 +46,7 @@ feature_nv_overlay:
 mapped_vlan:
   config_get: 'show running vlan'
   config_get_token: ['/^vlan <vlan>$/', '/^vn-segment (\d+)$/']
-  config_set: ['vlan <vlan>', '<state> vn-segment <vni>']
+  config_set: ['vlan <vlan>', '<state> vn-segment <vni> ; end']
```
chrisvanheuveln, line 49: Do you still need the `; end` here? We tried it as a test but I wasn't sure if it's actually needed anymore.

lib/cisco_node_utils/vni.rb
```diff
@@ -32,9 +32,15 @@ def initialize(vni_id, instantiate=true)
...
-      vni_list = config_get('vni', 'all_vnis')
+      if /N9K/.match(node.product_id)
+        feature = config_get('vni', 'feature_n9k')
+        return hash if feature.nil?
+        vni_list = config_get('vni', 'all_vnis_n9k')
+      else
+        feature = config_get('vni', 'feature')
+        return hash if (:enabled != feature.to_sym)
+        vni_list = config_get('vni', 'all_vnis')
+      end
```
chrisvanheuveln: Again, this platform callout should not be used. Try this instead to handle all platforms:

```
      return hash if config_get('vni', 'feature').nil?
      vni_list = config_get('vni', 'all_vnis')
```

Yaml:

```
feature:
  /N7K/:
    config_get_token: '/^feature vni/'
    config_set: 'feature vni'
  else
    config_get_token: '/^feature vn-segment-vlan-based$/'
    config_set: 'feature vn-segment-vlan-based'
```

lib/cisco_node_utils/vni.rb
```diff
@@ -91,7 +97,13 @@ def create
...
 
     def destroy
-      config_set('vni', 'destroy', @vni_id)
+      if /N9K/.match(node.product_id)
+        # Just destroy the vni-vlan mapping
+        vlan = mapped_vlan
+        config_set('vni', 'mapped_vlan', vlan: vlan, state: 'no', vni: @vni_id)
+      else
+        config_set('vni', 'destroy', @vni_id)
+      end
```
chrisvanheuveln, line 89: <del>This is okay for now but I'm exploring some ways to do this without the platform check.<del>

(edit: Glenn told me we can pass `vlan` unconditionally so that's the answer for this one.)
1. I would get rid of the `vlan = mapped_vlan`
2. You've got a mix of printf-style and key-val style yaml. Can you rework this to just use key-val? We're planning on doing that for the refactor anyway but it would be good to just fix it now.

(ignore my earlier code suggestion, use this instead)

```
      def destroy
        # Platform differences: For vni's mapped under the vlan config just remove the vni config and
        # not the vlan config; for standalone vni configs remove the entire vni config. The yaml cmdref
        # will ignore the unneeded keys for the standalone vni config.
        config_set('vni', 'destroy', state: 'no', vlan: mapped_vlan, vni: @vni_id)
      end
```

lib/cisco_node_utils/vni.rb
```diff
@@ -232,16 +244,21 @@ def mapped_vlan
...
     def mapped_vlan=(vlan)
-      if vlan == default_vlan
+      if vlan == default_mapped_vlan
         state = 'no'
         vlan = mapped_vlan
       else
         state = ''
       end
-      config_set('vni', 'mapped_vlan', vlan: vlan, state: state, vni: @vni_id)
+      # Nothing to be done if vlan is nil
```
chrisvanheuveln: This should raise an error if vlan is nil. I would just remove the vlan.nil? check altogether.

lib/cisco_node_utils/cmd_ref/vni.yaml
alok-aggarwal, line 49: Yes we need it. Without the ; end, the result[2][body] is empty. It doesn't contain any message.

lib/cisco_node_utils/cmd_ref/vni.yaml
alok-aggarwal: Done.

yes we need multiple to get vn-segments mapped to different vlans

lib/cisco_node_utils/vni.rb
alok-aggarwal: Done

lib/cisco_node_utils/vni.rb
alok-aggarwal, line 89: Didn't quite understand.

for N7K, we need "no vni 20000"
for N9K, we need 
vlan 200
  no vn-segment 20000

Not sure how the method proposed by you (single config_get) will work for both the syntaxes.

I have gotten rid of vlan = mapped_vlan though.

```
def destroy
  if /N9K/.match(node.product_id)
    # Just destroy the vni-vlan mapping
    config_set('vni', 'mapped_vlan', vlan: mapped_vlan, state: 'no', vni: @vni_id)
  else
    config_set('vni', 'destroy', vni: @vni_id)
  end
end
```

lib/cisco_node_utils/vni.rb
alok-aggarwal: Done.

lib/cisco_node_utils/vni.rb
chrisvanheuveln, line 89:
```
destroy:
  /N7K/:
    config_set: "no vni <vni>"
  /N9K/:
    config_set: ['vlan <vlan>', 'no vn-segment <vni> ; end']
```

```
    def destroy
      # Platform differences: For vni's mapped under the vlan config just remove
      # the vni config and not the vlan config; for standalone vni configs
      # remove the entire vni config. The yaml cmdref will ignore the unneeded
      # keys for the standalone vni config.
      config_set('vni', 'destroy', state: 'no', vlan: mapped_vlan, vni: @vni_id)
    end

```

chrisvanheuveln:
Alok, I had to test out a couple of theories and ended up refactoring a bunch of this code; so I think I'll go ahead and merge your diff, then submit a follow-up PR with my changes for you and Deepak to review.

PR #133, opened 2015-12-16
Vni cleanup

chrisvanheuveln:
```
vni refactor #1

This is step 1 of the refactor, there is still plenty to fix in here.

The main thrust of this was adding platform checks for the n7k vs n9k differences. I suspect this needs more tweaking. Deepak & Alok should give these changes a once-over in case I excluded something I shouldn't have.

I updated the yaml to use key-val and added platform checks there too.

Fixed the feature methods.

Added additional tests to minitest along with platform checks.
```


chrisvanheuveln:
I picked up some changes after merging so I'll close this and re-open with just the vni changes.

PR #134, opened 2015-12-16
   vni refactor #1

chrisvanheuveln:
```
This is step 1 of the refactor, there is still plenty to fix in here.

The main thrust of this was adding platform checks for the n7k vs n9k differences. I suspect this needs more tweaking. Deepak & Alok should give these changes a once-over in case I excluded something I shouldn't have.

I updated the yaml to use key-val and added platform checks there too.

Fixed the feature methods.

Added additional tests to minitest along with platform checks.
```

### DIFF
--- a/lib/cisco_node_utils/vni.rb
+++ b/lib/cisco_node_utils/vni.rb
@@ -32,10 +32,8 @@ module Cisco
 
     def self.vnis
       hash = {}
-      feature = config_get('vni', 'feature')
-      return hash if (:enabled != feature.to_sym)
       vni_list = config_get('vni', 'all_vnis')
-      return hash if vni_list.nil? || vni_list == {}
+      return hash if vni_list.nil?
 
       vni_list.each do |id|
         hash[id] = Vni.new(id, false)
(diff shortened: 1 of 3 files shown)


mikewiebe:
@chrisvanheuveln @alok-aggarwal Will https://github.com/cisco/cisco-network-puppet-module/pull/110 need to change based on this refactor?

chrisvanheuveln:
It doesn't at this point but will need additional work when round 2 occurs.

mikewiebe:
:+1: Go ahead and merge after you fix the conflicts.

lib/cisco_node_utils/vni.rb
```diff
@@ -43,55 +41,46 @@ def self.vnis
...
+    rescue Cisco::CliError => e
+      # cmd will syntax reject when feature is not enabled
+      raise unless e.clierror =~ /Syntax error/
+      return false
     end
 
-    def feature_set(vni_set)
-      curr = feature
-      return if curr == vni_set
+    def self.feature_vni_enable
```
alok-aggarwal: This isn't correct. feature vn-segment-vlan-based can be enabled independent of feature nv overlay.

But if both are enabled, feature nv overlay should be disabled prior to disabling feature vn-segment-vlan-based.

alok-aggarwal: Also note that it takes about 7-8 seconds to turn of feature nv overlay. Thats why i had the sleep(8) there, so that the minitest doesn't get confused.

lib/cisco_node_utils/vni.rb
```diff
@@ -43,55 +41,46 @@ def self.vnis
...
+      # feature nv overlay is a dependency for feature vni
+      Vni.feature_nv_overlay_enable unless Vni.feature_nv_overlay_enabled
```
alok-aggarwal, line 55: This isn't correct. feature vn-segment-vlan-based can be enabled independent of feature nv overlay.

But if both are enabled, feature nv overlay should be disabled prior to disabling feature vn-segment-vlan-based.

alok-aggarwal, line 55: Also note that it takes about 7-8 seconds to turn of feature nv overlay. Thats why i had the sleep(8) there, so that the minitest doesn't get confused.

chrisvanheuveln, line 55: Got it. Since we don't want to do any feature disablement for these anyway it's just a matter of removing the comment. Thanks.

chrisvanheuveln:
I still have some pre-merge code in here so I'll kill this PR as well and fix.

lib/cisco_node_utils/vni.rb
alok-aggarwal, line 55: Well I'd say just remove the whole thing:
Vni.feature_nv_overlay_enable unless Vni.feature_nv_overlay_enabled. We don't need to enable nv overlay for vn-segment-vlan-based.

But another point is I have a test in minitest which tests feature on/off. In that the test will fail if nv overlay is also on be some other provider and we don't disable it before disabling vn-segment-vlan-based
I suggest we keep feature on/off test and the disablement part unless you feel otherwise.

PR #136, opened 2015-12-16
Implemented IOS XR support for BGP address-family commands and properties and refactored tests to use getters/setters.

brturn:


glennmatthews:
Looks like you have a mismatch between your git user configuration and your github account - the commits don't seem to be associated with your 'brturn' github account.

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -98,21 +98,41 @@ def set_args_keys(hash={}) # rubocop:disable Style/AccessorMethodName
     #                      PROPERTIES                      #
     ########################################################
 
+    # IOS XR?
+    def ios_xr?
```
glennmatthews, line 102: I like this - consider moving it into the NodeUtil class so it can be shared to other provider classes.

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -397,20 +446,24 @@ def dampening=(damp_array)
...
-      elsif route_map.is_a? String
-        # 'dampening route-map WORD' command
-        route_map = "route-map #{damp_array}"
-        route_map.strip!
-        Cisco::Logger.debug("Dampening 'dampening #{route_map}'")
       else
-        # Array not in a valid format
-        fail ArgumentError
+        # 'dampening route-map WORD' command
+        if platform == :nexus
```
glennmatthews: Might as well create a `nexus?` method on NodeUtils as well.

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -502,12 +556,16 @@ def networks=(should_list)
       return if delta_hash.values.flatten.empty?
       [:add, :remove].each do |action|
         Cisco::Logger.debug("networks delta #{@get_args}\n #{action}: " \
-                            "#{delta_hash[action]}")
-        delta_hash[action].each do |network, route_map|
+                          "#{delta_hash[action]}")
```
glennmatthews: This indentation change looks suspect.

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -533,7 +591,7 @@ def redistribute=(should)
       return if delta_hash.values.flatten.empty?
       [:add, :remove].each do |action|
         Cisco::Logger.debug("redistribute delta #{@get_args}\n #{action}: " \
-                            "#{delta_hash[action]}")
+                          "#{delta_hash[action]}")
```
glennmatthews: ditto comment about indentation. Yet rubocop is happy with it, it looks like - I'm confused. :-)

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -1,44 +1,52 @@
...
-  config_set_append: '<state> additional-paths send'
+  config_set_append: ' <state> additional-paths send'
   default_value: false
 
 advertise_l2vpn_evpn:
+  _exclude: [cli_ios_xr]
   kind: boolean
   config_get_token_append: '/^advertise l2vpn evpn$/'
-  config_set_append: '<state> advertise l2vpn evpn'
+  config_set_append: ' <state> advertise l2vpn evpn'
```
glennmatthews: Something's off here - if this command doesn't apply to XR, the leading space isn't needed. If it does apply, the _exclude is wrong. Same comment on dampen_igp_metric below.

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -83,12 +104,28 @@ dampening_suppress_time:
...
-  config_set_append: '<state> default-information originate'
-  default_value: false
+  cli_nexus:
+    config_get_token_append: '/^default-information originate$/'
+    config_set_append: ' <state> default-information originate'
+    default_value: false
+  cli_ios_xr:
+    # IOS XR does not have default-information under address-family,
+    #   so we must replace config get/set
+    # Perhaps better to fix in ruby code?
```
glennmatthews: Yeah this looks sketchy. :-)

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -98,21 +98,41 @@ def set_args_keys(hash={}) # rubocop:disable Style/AccessorMethodName
...
     #
     def client_to_client
+      return nil if vrf_unsupported_xr?
       state = config_get('bgp_af', 'client_to_client', @get_args)
       state ? true : false
     end
 
     def client_to_client=(state)
+      fail_vrf_unsupported_xr('client-to-client')
+      state = !state if ios_xr?
```
robert-w-gries, line 136: Can you add a comment that explains this line?  Not sure why it needs to be inversed.

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -98,21 +98,41 @@ def set_args_keys(hash={}) # rubocop:disable Style/AccessorMethodName
...
+    # IOS XR?
+    def ios_xr?
+      platform == :ios_xr
+    end
+
+    def vrf_unsupported_xr?
+      ios_xr? && @vrf != 'default'
+    end
+
+    def fail_vrf_unsupported_xr(reason)
```
robert-w-gries: Nit: rename method to `fail_if_vrf_unsupported_xr` so it's clear that the method does not always call the `fail` method

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -160,6 +180,32 @@ def default_next_hop_route_map
...
     end
 
+    def next_hop_route_policy
+      return nil if vrf_unsupported_xr?
+      config_get('bgp_af', 'next_hop_route_policy', @get_args)
+    end
+
+    def next_hop_route_policy=(route_policy)
+      fail_vrf_unsupported_xr('nexthop route-policy')
+      route_policy.strip!
```
robert-w-gries: Adding `kind: string` to the yaml entry should eliminate the need for this line

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
brturn: After Charles explained the bug to me, I did a replace-all.  I consider this a latent bug and if it doesn't hurt Nexus, then adding a space may possibly eliminate future copy-paste recurrences of the bug.

I've removed the space from all nexus-only lines, but left it in all XR and XR/NX lines.

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
brturn: Not sure what the right fix is?  Currently there is a check for vrf-unsupported features under XR, so I imagine this could be added as an address-family-unsupported feature?  But the other fix feels like a bit of a hack to begin with.  Any advice?

glennmatthews: If this is a 'bgp_af' property that XR BGP doesn't support under any address families, then this property should be excluded on XR. (I guess on XR, from the comment, that this command is configured in some other scope? In that context it would then be appropriate to have a future work item to add a version of this property under the appropriate object for that scope.)

RichWellum: FYI:

It's not supported under bgp af:

RP/0/RP0/CPU0:ios(config)#router bgp 5
RP/0/RP0/CPU0:ios(config-bgp)#address-family ipv4 unicast
RP/0/RP0/CPU0:ios(config-bgp-af)#bgp d?
dampening
RP/0/RP0/CPU0:ios(config-bgp-af)#bgp ?
  attribute-download  Configure attribute download for this address-family
  bestpath            Change default route selection criteria
  client-to-client    Configure client to client route reflection
  dampening           Enable route-flap dampening
  label-delay         Specify delay for batching label processing
  origin-as           BGP origin-AS knobs
  scan-time           Configure background scanner interval for this address-family

So _exclude here for XR, then add the logic minitest to catch the UnsupportedError. Unless it's a special case where for example it works under the af for a default vrf, but not a non-default.

tests/test_bgp_af.rb
```diff
@@ -314,65 +371,59 @@ def test_additional_paths_selection
     #
     # Set and verify
     #
+    if platform == :ios_xr
+      config_ios_xr_dependencies(asn)
+      config('route-policy drop_all', '#drop_all', 'end-policy')
```
robert-w-gries: `'#drop_all'` seems like a no-op to me.  It was pulled in from #87, does anyone know the reason for that command being there?

Edit:  [This line](https://github.com/cisco/cisco-network-node-utils/pull/136/files#diff-09b95f5c8a820ab9f6db3a11831816deR289) has the same command

brturn: Out of my area of expertise, can you suggest an alternative?  It looks like XR needs some additional setup before configuring the route-policy for additional-paths.  I think this is meant to be a mock policy?  Would it be better to setup a 'realistic' policy instead?

brturn:
@glennmatthews 

> Looks like you have a mismatch between your git user configuration and your github account
> the commits don't seem to be associated with your 'brturn' github account.

New to github; not sure what you are seeing as a problem?  It looks fine on my end.

glennmatthews:
Do you have the email address 'brturn@cisco.com' associated with your GitHub user account? GitHub doesn't appear to be recognizing that 'brturn@cisco.com' is GitHub user @brturn.

(Edit: looks to be fixed now - thanks!)

tests/test_bgp_af.rb
chburket: I think it would probably be fine to just remove the '#drop_all' arg from the config call.  Not sure why Geert had it in there.

[edit] I just tried this on XR and it works fine without the "#drop_all"

RichWellum: Why not move this into config_ios_xr_dependencies(asn)?

You can look at bgp neighbor af for similar:
  def setup_ios_xr
...
 # Global requirements
    # route policies
    str1 = 'route-map-in-name'
    str2 = 'route-map-out-name'
    config("route-policy #{str1}", 'end-policy')
    config("route-policy #{str2}", 'end-policy')
    config("route-policy #{str1.reverse}", 'end-policy')
    config("route-policy #{str2.reverse}", 'end-policy')
    config('route-policy foo_bar', 'end-policy')
    config('route-policy baz_inga', 'end-policy')

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -1,41 +1,49 @@
...
-    - 'vrf <vrf>'
-    - 'address-family <afi> <safi>'
+    - ' vrf <vrf>'
+    - ' address-family <afi> <safi>'
+  cli_nexus:
+    config_get: 'show running bgp all'
+  cli_ios_xr:
+    config_get: 'show running-config router bgp'
 
 additional_paths_install:
```
chburket, line 17: I think that additional_paths_install is not supported on XR at all (even under default vrf).  Can you double check?

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -98,21 +98,49 @@ def set_args_keys(hash={}) # rubocop:disable Style/AccessorMethodName
...
+    # Nexus?
+    def nexus?
+      platform == :nexus
+    end
+
+    def vrf_unsupported_xr?
+      ios_xr? && @vrf != 'default'
+    end
+
+    def fail_if_vrf_unsupported_xr(reason)
```
chburket, line 115: Can you fix this so there are no extra spaces in the error message?

``` ruby
      fail Cisco::UnsupportedError.new(
        'bgp_af', reason, 'set',
        "#{reason} is not configurable on a "\
        'per-VRF basis on IOS XR') if vrf_unsupported_xr?
```

glennmatthews:
:+1:

brturn:
@chburket @RichWellum @robert-w-gries Any last comments?  Thumbs up?

robert-w-gries:
:+1:

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
brturn, line 17: You are correct - I'll add that to be fixed in my next task.

lib/cisco_node_utils/bgp_af.rb
brturn, line 115: Fixed

chburket:
:+1:

tests/test_bgp_af.rb
```diff
@@ -106,6 +110,41 @@ def test_collection_not_empty
     end
   end
 
+  def config_ios_xr_dependencies(asn, vrf='red', casty='unicast')
+    # These dependencies are required on ios xr
+
+    # router bgp 55
```
RichWellum: Nit. I think you can condense this example in the comments? 

And there's a couple of unnecessary comments in there like 'Configure'.

Also not sure what calling the variable 'casty' represents?

Is the loopback needed?

tests/test_bgp_af.rb
```diff
@@ -118,114 +157,151 @@ def test_default_information_originate
...
+    assert_raises(Cisco::UnsupportedError) do
+      bgp_af.next_hop_route_policy = '  drop_all  '
+    end
+  end
+
+  ##
+  ## next_hop route-map or route-policy
+  ##
+  def test_next_hop_route_map_or_policy
+    asn = '55'
```
RichWellum: You set the asn and vrf and af etc several times. Maybe a helper method? See bgp global for examples.

tests/test_bgp_af.rb
```diff
@@ -269,22 +330,17 @@ def test_additional_paths
...
     bgp_af.additional_paths_receive = true
-    bgp_af.additional_paths_install = true
-
-    af_string = get_bgp_af_cfg(asn, vrf, af)
-
-    [pattern_send, pattern_receive, pattern_install].each do |pat|
-      assert_match(pat, af_string,
-                   "Error: '#{pat}' is not configured and should be")
-    end
+    bgp_af.additional_paths_install = true if platform == :nexus
```
RichWellum: Should we set these for both platforms and then catch the error? Ditto to similar.

lib/cisco_node_utils/bgp_af.rb
RichWellum, line 102: If you don't do this in this commit - can you add a bug to our backlog to track this please?

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -142,7 +170,7 @@ def next_hop_route_map
     end
 
     def next_hop_route_map=(route_map)
-      route_map.strip!
+      #      route_map.strip!
```
RichWellum: Remove funky comment?

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -160,6 +188,32 @@ def default_next_hop_route_map
...
     end
 
+    def next_hop_route_policy
+      return nil if vrf_unsupported_xr?
+      config_get('bgp_af', 'next_hop_route_policy', @get_args)
+    end
+
+    def next_hop_route_policy=(route_policy)
+      fail_if_vrf_unsupported_xr('nexthop route-policy')
+      #      route_policy.strip!
```
RichWellum: Ditto comment.

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -397,20 +454,24 @@ def dampening=(damp_array)
...
-        # Array not in a valid format
-        fail ArgumentError
+        # 'dampening route-map WORD' command
+        if nexus?
+          route_map = "route-map #{damp_array}"
+          route_map.strip!
+          Cisco::Logger.debug("Dampening 'dampening #{route_map}'")
+        elsif ios_xr?
+          route_policy = "route-policy #{damp_array}"
+          #          route_policy.strip!
```
RichWellum: Comment ditto.

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -96,36 +119,60 @@ feature_nv_overlay_evpn:
 
 maximum_paths:
   kind: int
-  config_get_token_append: '/^maximum-paths (\d+)$/'
-  config_set_append: '<state> maximum-paths <num>'
+  cli_nexus:
+    config_get_token_append: '/^maximum-paths (\d+)$/'
+    config_set_append: '<state> maximum-paths <num>'
+  cli_ios_xr:
+    # BT: Is "maximum-paths" without specialization equal to "maximum-paths ebgp"?
```
RichWellum: Comment in the wild? Did you answer this question?

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -96,36 +119,60 @@ feature_nv_overlay_evpn:
...
 network:
   multiple: true
-  config_get_token_append: '/^network (\S+) ?(?:route-map )?(\S+)?$/'
-  config_set_append: '<state> network <network> <route_map>'
+  cli_nexus:
+    config_get_token_append: '/^network (\S+) ?(?:route-map )?(\S+)?$/'
+    config_set_append: '<state> network <network> <route_map>'
+  cli_ios_xr:
+    config_get_token_append: '/^network (\S+) ?(?:route-policy )?(\S+)?$/'
+    config_set_append: ' <state> network <network> <route_policy>'
```
RichWellum: The set append could be common to both right? I though that was the approach we are taking where possible. Applies globally.

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -96,36 +119,60 @@ feature_nv_overlay_evpn:
...
-  config_set_append: '<state> redistribute <protocol>'
+  config_set_append: ' <state> redistribute <protocol>'
+  cli_nexus:
+    config_get_token_append: '/^redistribute (\S+ ?\S+?) ?(?:route-map (\S+))?$/'
+  cli_ios_xr:
+    config_get_token_append: '/^redistribute (\S+ ?\S+?) ?(?:route-policy (\S+))?$/'
   default_value: []
 
 redistribute_policy:
   # route-map/policy is optional on some platforms, required on others
```
RichWellum: Don't think this comments adds much.

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -96,36 +119,60 @@ feature_nv_overlay_evpn:
...
-  config_set_append: '<state> redistribute <protocol> route-map <policy>'
+  cli_nexus:
+    config_set_append: '<state> redistribute <protocol> route-map <policy>'
+  cli_ios_xr:
+    config_set_append: ' <state> redistribute <protocol> route-policy <policy>'
```
RichWellum, line 207: I think these sets could be combined very easily. In the nu_setter - you'd format the string and pass that as the token. E.g instead of passing <policy> you'd pass <policy-string> equiv where it is route-map for nexus and route-policy for xr. We've done that in several places already.

tests/test_bgp_af.rb
```diff
@@ -28,7 +28,11 @@ def setup
     super
     # Disable and enable feature bgp before each test to ensure we
     # are starting with a clean slate for each test.
-    config('no feature bgp', 'feature bgp')
+    if platform == :nexus
+      config('no feature bgp', 'feature bgp')
+    elsif platform == :ios_xr
```
RichWellum, line 33: Personally I think the elsif platform = xyz, could just be replaced with 'else' throughout the code. We only support two platforms currently. No biggy.

tests/test_bgp_af.rb
```diff
@@ -314,65 +370,59 @@ def test_additional_paths_selection
     #
     # Set and verify
     #
+    if platform == :ios_xr
+      config_ios_xr_dependencies(asn)
+      config('route-policy drop_all', 'end-policy')
```
RichWellum: Think I mentioned in a previous pr# this should go into the method above.

tests/test_bgp_af.rb
```diff
@@ -314,65 +370,59 @@ def test_additional_paths_selection
...
-    val = bgp_af.default_advertise_l2vpn_evpn
-    bgp_af.advertise_l2vpn_evpn = val
-    assert_equal(val, bgp_af.advertise_l2vpn_evpn,
-                 'Error: advertise l2vpn evpn value does not match default' \
-                 'value')
+    initial = bgp_af.advertise_l2vpn_evpn
+
+    if platform == :ios_xr
+      assert_raises(Cisco::UnsupportedError) do
+        bgp_af.advertise_l2vpn_evpn = !initial
```
RichWellum: Do you need to do !initial? It's simply not supported regardless of what you set.

tests/test_bgp_af.rb
```diff
@@ -390,123 +440,78 @@ def test_dampen_igp_metric
...
+  def dampening_vrf_xr_neg_test
     asn = '101'
+    vrf = 'red'
+    af = %w(ipv4 unicast)
+    bgp_af = RouterBgpAF.new(asn, vrf, af)
+    assert_nil(bgp_af.dampening,
+               'dampening should return nil on XR with non-default vrf')
+    assert_raises(Cisco::UnsupportedError) do
+      bgp_af.dampening = %w(one two)
+    end
```
RichWellum: bgp_af.destroy ? Can you check you're doing this consistently for all test methods. Also in some you add an extra line before it and some you don't - can you make it consistent please?

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -160,6 +188,32 @@ def default_next_hop_route_map
...
+    end
+
+    def next_hop_route_policy=(route_policy)
+      fail_if_vrf_unsupported_xr('nexthop route-policy')
+      #      route_policy.strip!
+      if route_policy.empty?
+        state = 'no'
+        # Dummy route policy required if not configured.
+        if next_hop_route_policy.empty?
+          route_policy = 'dummy_routepolicy'
```
RichWellum: I'd like to know if the team thinks that adding a dummy route_policy is ok, or is it better to fail (from the manifest) and add support for route-policy asap?

Good example is rd auto - we just failed until someone added it.

tests/test_bgp_af.rb
```diff
@@ -118,114 +157,151 @@ def test_default_information_originate
...
+  end
 
-    assert_match(pattern, af_string,
-                 "Error: 'client-to-client reflection' is not configured " \
-                   'and should be')
+  ##
+  ## client-to-client reflection (BT: trying a different style)
+  ##
+  def test_client_to_client
+    asn = '55'
```
RichWellum: Note that l2vpn is perfectly valid address-family:

router bgp 55
 address-family l2vpn evpn

You should add these tests throughout. You can see bgp neighbor af for examples on how to do this within a matrix that can be called for each test method. It's a harness I think should be adopted.

Ditto ipv6 and multicast.

tests/test_bgp_af.rb
brturn: Cleaned this up; loopback does not appear to be needed.

"casty" is just capturing the unicast/multicast of the address-family, it is only used in one place.  This will not be needed after migrating to a matrix-based test harness.

tests/test_bgp_af.rb
brturn: Wasn't sure how much of the code to muck with for this story.  It does seem very redundant.

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
brturn, line 207: Didn't quite understand you here:  What is the "nu_setter"?  Are you suggesting passing "route-map" from the test harness when running on Nexus, and "route-policy" when running on XR, then just dropping that string into a platform-common command line here?

I was not so sure of the route-map/policy split, Geert mentioned that some versions of IOS may have both.  Code would be much cleaner if we collapsed them and dealt with the special case of having both at some later date.

tests/test_bgp_af.rb
brturn: Agreed.

tests/test_bgp_af.rb
brturn: I was told that setting it to the value it is already set at would cause no commands to occur.  I saw this happen in one of the tests and came up with the !initial as a general answer to the issue.  If I'm mistaken about the no-op, let me know!

tests/test_bgp_af.rb
brturn: Agreed.

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
RichWellum, line 207: nu_setter : node_utils setter function.

No nothing to do with the test file; I am suggesting something like what I am doing in bgp af neighbor:

```
def default_originate_set(state, map=nil)
      if platform == :ios_xr
        map = "route-policy #{map}" unless map.nil?
      else
        map = "route-map #{map}" unless map.nil?
      end
      set_args_keys(state: (state ? '' : 'no'), map: map)
      config_set('bgp_neighbor_af', 'default_originate', @set_args)
    end
```

So map is then the token in the yaml:

```
default_originate:
  default_value: false
  cli_ios_xr:
    config_get_token_append: '/^default-originate(?: route-policy .*)?/'
    config_set_append: ' <state> default-originate <map>'
  cli_nexus:
    config_get_token_append: '/^default-originate(?: route-map .*)?/'
    config_set_append: '<state> default-originate <map>'
```

Of course I didn't make the set_append common which kind of makes this redundant, but I was expecting to clean that all up with the yaml User Story.

tests/test_bgp_af.rb
RichWellum: This is really a nit, so if it's too much work then don't worry about it.

tests/test_bgp_af.rb
RichWellum: Right - I see the statement above now. Carry on. :)

tests/test_bgp_af.rb
RichWellum, line 33: This is a nit btw.

tests/test_bgp_af.rb
RichWellum: Ack.

robert-w-gries:
@brturn Can you please merge upstream `feature/grpc` branch into your fork, so you can pull in #145 and fix Travis?

brturn:
Merged upstream - some new tests fixed, some skipped to get things compiling.
Rewrote test harness to use a matrix of tests and expected results.
@glennmatthews @RichWellum @robert-w-gries @chburket

robert-w-gries:
Rubocop is not passing according to Travis:

```
Offenses:

tests/test_bgp_af.rb:599:1: C: Trailing whitespace detected.
```

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -125,36 +149,62 @@ inject_map:
...
+  cli_ios_xr:
+    config_get_token_append: '/nexthop route-policy (.*)$/'
+    config_set_append: ' <state> nexthop route-policy <route_map>'
+
+next_hop_route_policy:
+  _exclude: [cli_nexus]
+  kind: string
   default_value: ""
+  config_get_token_append: '/nexthop route-policy (.*)$/'
+  config_set_append: ' <state> nexthop route-policy <route_policy>'
```
robert-w-gries: Shouldn't this property be removed since `next_hop_route_map` configures `route_policy` for XR?

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
robert-w-gries: ping @brturn

brturn:
Happy New Year!

Poking everyone about this PR again.  This is a major rewrite, collapsing most of the tests into a single harness.  We may need a sit-down code review?  I've left comments inline, and I'm sure it doesn't fit with typical Ruby code style. 

@glennmatthews @RichWellum @robert-w-gries @chburket

tests/test_bgp_af.rb
```diff
@@ -28,31 +28,254 @@ def setup
...
 
-  def get_bgp_af_cfg(asn, vrf, af)
-    afi, safi = af
-    string =
-      @device.cmd("show run bgp all | sec 'bgp #{asn}' |  sec 'vrf #{vrf}' | " \
-                  "sec 'address-family #{afi} #{safi}' | no-more")
-    string
+  # rubocop:disable Style/WordArray
+  # rubocop:disable Style/ClassVars
+  # rubocop:disable Metrics/LineLength
```
robert-w-gries, line 40: - `Style/WordDisable` - Why do you want to disable this?  The matrices look cleaner to me with the word array:

``` diff
-    # rubocop:disable Style/WordArray
-    master = [['lax', 'sfo'],
-              ['lax', 'sjc'],
-              ['nyc', 'sfo', 'copy-attributes'],
-              ['sjc', 'nyc', 'copy-attributes']]
+    master = [%w(lax sfo),
+              %w(lax sjc),
+              %w(nyc sfo copy-attributes),
+              %w(sjc nyc copy-attributes)]
```
- `Style/ClassVars` - Do the test variables need to be class variables?  I feel this style doesn't need to be disabled.
- `Metrics/LineLength` - This doesn't need to be disabled.  Instead you can increase the `Max` value allowed, which is set inside `tests/.rubocop.yml`

PSA: WordArray is an autocorrectable offense, which means you can just run `rubocop -a` to change the arrays to word arrays.

brturn, line 40: - Style/WordDisable -  I didn't write the inject_map code, so I don't know why the author disabled that one.  Looks about the same to me either way.
- ClassVars - I copied the matrix setup from test_bgp_neighbor_af.rb, which uses class vars.  When RuboCop complained I made them instance vars and then got lots of odd errors when running the tests.  Decided it wasn't worth the trouble and just added them back to being class vars.  Suggestions welcome.

lib/cisco_node_utils/bgp_af.rb
glennmatthews, line 115: Might be more readable if you change 'reason' to 'property'? Nitpick.

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -142,7 +170,6 @@ def next_hop_route_map
     end
 
     def next_hop_route_map=(route_map)
-      route_map.strip!
```
glennmatthews, line 145: What was the reason behind removing these? Are they provably unnecessary?

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -602,11 +635,15 @@ def networks=(should_list)
...
           state = (action == :add) ? '' : 'no'
           network = Utils.process_network_mask(network)
-          route_map = "route-map #{route_map}" unless route_map.nil?
-          set_args_keys(state: state, network: network, route_map: route_map)
+          unless route_map_policy.nil?
+            route_map = "route-map #{route_map_policy}"
+            route_policy = "route-policy #{route_map_policy}"
+          end
+          set_args_keys(state: state, network: network, route_map: route_map,
+                        route_policy: route_policy)
```
glennmatthews, line 636: The whole route-map vs route-policy behavior (where sometimes we treat them as the same thing and other times we treat them as completely separate) still smells wrong to me. Not blocking for this PR but I'll open a user story to revisit this in general.

tests/test_bgp_af.rb
glennmatthews, line 33: I'd rather leave it as is but am OK either way.

tests/test_bgp_af.rb
glennmatthews, line 40: Do you still need to disable the WordArray rule? Looks like you're using word arrays below now.

glennmatthews, line 40: I would not favor changing Metrics/LineLength in .rubocop.yml - let's keep the current 80 character limit and disable on a case-by-case basis where it makes sense (like below)

tests/test_bgp_af.rb
```diff
@@ -28,31 +28,254 @@ def setup
...
+  # Value-based properties
+  @@tvalues = [
+    [:default_information_originate,  [:toggle]],
+    [:client_to_client,               [:toggle]],
+    [:additional_paths_send,          [:toggle]],
+    [:additional_paths_receive,       [:toggle]],
+    [:additional_paths_install,       [:toggle]],
+    [:advertise_l2vpn_evpn,           [:toggle]],
+
+    # TODO: To be added in next iteration
```
glennmatthews: Do we have a user story for these TODOs?

lib/cisco_node_utils/bgp_af.rb
RichWellum, line 636: My initial idea was to delineate in the setter - create a key string where x = route-policy or route-map and pass x to the yaml. That was an idea from how we'd implemented something similar. The above seems odd to me too.

tests/test_bgp_af.rb
```diff
@@ -28,31 +28,254 @@ def setup
...
+                            "Assert 'cli error' failed for: #{test}, #{asn}, #{vrf}, #{af}") do
+                bgp_af.send("#{test}=", 'foo') # *** BT: Don't like this ***
+              end
+
+            elsif expect == :unsupported
+              puts '         Unsupported'
+
+              # Getter should return nil when unsupported?
+              # BT: does not seem to work in general ..?
+              #            assert_nil(initial, "Assert 'nil' inital value failed for: #{test} #{asn} #{vrf} #{af}")
```
glennmatthews: This should work - what value are you seeing for 'initial'?

Actually, you should maybe be testing `assert_equal(default, initial)` here just as you are doing in the 'else' case below.

tests/test_bgp_af.rb
```diff
@@ -106,407 +329,55 @@ def test_collection_not_empty
...
-  # rubocop:disable Metrics/MethodLength
   def test_dampening
     asn = '101'
-
-    ############################################
-    # Set and verify 'dampening' with defaults #
-    ############################################
-    vrf = 'orange'
+    vrf = 'default'
+    vrf = 'orange' if platform == :nexus
```
glennmatthews: Suggestion:

``` ruby
# this property is only applicable to the default vrf on XR
vrf = (platform == :ios_xr ? 'default' : 'orange')
```

Or alternately:

``` ruby
af = %w(ipv4 unicast)
if platform == :nexus
  vrf = 'orange'
elsif platform == :ios_xr
  vrf = 'default'
  config_ios_xr_dependencies(asn, vrf)
end
```

tests/test_bgp_af.rb
```diff
@@ -106,407 +329,55 @@ def test_collection_not_empty
...
 
-    pattern = /^ *dampening$/
-
+    ############################################
+    # Set and verify 'dampening' with defaults #
+    ############################################
     bgp_af.dampening = []
+    assert_equal(bgp_af.default_dampening,
+                 bgp_af.dampening) if platform == :nexus
+    assert_equal('', bgp_af.dampening) if platform == :ios_xr
```
glennmatthews, line 383: is [] not the default on XR? if so then I'd prefer you update the `bgp_af.dampening = []` above since the comment indicates that we're setting to default.

tests/test_bgp_af.rb
```diff
@@ -563,122 +404,79 @@ def test_dampening
...
-                 bgp_af.dampening_suppress_time,
-                 'Wrong default dampening suppress_time value configured')
-    assert_equal(bgp_af.default_dampening_max_suppress_time,
-                 bgp_af.dampening_max_suppress_time,
-                 'Wrong default dampening suppress_max_time value configured')
-    assert_equal(bgp_af.default_dampening_routemap,
-                 bgp_af.dampening_routemap,
-                 'The default dampening routemap should configured')
+    # IOS XR; we skip this section because we cannot query
+    #    the default parameters
```
glennmatthews, line 447: This seems wrong to me - if this property is supported at all, it must have default values. Fix the provider instead of disabling the test?

tests/test_bgp_af.rb
```diff
@@ -745,6 +544,7 @@ def default_metric(asn, vrf, af)
   def test_default_metric
     afs = [%w(ipv4 unicast), %w(ipv6 unicast)]
     afs.each do |af|
+      config_ios_xr_dependencies('55') if platform == :ios_xr
```
glennmatthews, line 553: Seems a little odd to pass `'55'` instead of `55` - is there a problem with the function definition?

tests/test_bgp_af.rb
```diff
@@ -794,96 +594,17 @@ def inject_map_tester(bgp_af, should, test_id)
   end
 
   def test_inject_map
+    # Not sure why it's broken; to be fixed in next iteration
```
glennmatthews: next iteration == another user story?

Please mark these with `# TODO:` to make them easier to find in future.

tests/test_bgp_af.rb
```diff
@@ -1122,6 +879,7 @@ def test_utils_delta_add_remove
 
   # test route_target
   def test_route_target
+    skip('Currently broken on IOS XR: US59615') if platform == :ios_xr
```
glennmatthews, line 888: please add a TODO comment here - thanks.

tests/test_bgp_af.rb
```diff
@@ -28,31 +28,254 @@ def setup
...
+  # Address Families to test:
+  @@tafs = [
+    #   afi  safi
+    %w(ipv4 unicast),
+    %w(ipv6 unicast),
+    %w(ipv4 multicast),
+    %w(ipv6 multicast),
+    %w(l2vpn evpn),
+
+    # BT: These are additional address families/modifiers reported by XR, should they be tested also?
```
glennmatthews: can you change all of these `BT:` to `TODO:`?

tests/test_bgp_af.rb
glennmatthews, line 40: It looks like the class variables are constants (they're defined once and never changed). Unless I'm missing something, try changing them to constants like `T_AFS`, `T_ASNS`, `T_VRFS`, etc. and see if the code works and rubocop is happier.

tests/test_bgp_af.rb
```diff
@@ -28,31 +28,254 @@ def setup
...
+        @@tafs.each do |af|
+          puts '**************************************'
+          @@tvalues.each do |test, test_values|
+            puts "******** #{test}, #{asn}, #{vrf}, #{af}"
+
+            # Override expectation for some specific cases..
+            expect = check_test_exceptions(test, platform, vrf, af)
+
+            # Setup
+            bgp_af = RouterBgpAF.new(asn, vrf, af)
```
glennmatthews: Just wondering - is it necessary to create/destroy the AF instance for each value under test? Or can this be moved outside the `@@tvalues.each` loop to improve performance of the test?

tests/test_bgp_af.rb
```diff
@@ -28,31 +28,254 @@ def setup
...
+              # Do nothing..
+              puts '         skip'
+
+            elsif expect == :CliError
+              puts '         CliError'
+
+              # This set of parameters produces a CLI error
+              assert_raises(Cisco::CliError,
+                            "Assert 'cli error' failed for: #{test}, #{asn}, #{vrf}, #{af}") do
+                bgp_af.send("#{test}=", 'foo') # *** BT: Don't like this ***
```
glennmatthews: Agreed. Would it make more sense to try one (or all) of the values in `test_values` instead?

tests/test_bgp_af.rb
```diff
@@ -893,13 +614,21 @@ def test_network
...
   end
 
   def network_cmd(af, dbg)
+    if platform == :ios_xr
+      %w(rtmap1 rtmap2 rtmap3 rtmap5 rtmap6 rtmap7).each do |policy|
+        config("route-policy #{policy}", 'end')
+      end
+    end
     # Initial 'should' state
     if /ipv6/.match(dbg)
```
glennmatthews, line 637: I know this is base code but since you're rewriting big chunks anyway - why are we regexping against the dbg string instead of checking `af.afi == 'ipv6'`?

tests/test_bgp_af.rb
```diff
@@ -981,13 +724,22 @@ def redistribute_cmd(af, dbg)
     # rubocop:disable Style/WordArray
     # Initial 'should' state
     ospf = (dbg.include? 'ipv6') ? 'ospfv3 3' : 'ospf 3'
```
glennmatthews, line 730: same question here about checking dbg instead of af.afi

tests/test_bgp_af.rb
```diff
@@ -981,13 +724,22 @@ def redistribute_cmd(af, dbg)
...
+                ['isis 2',  'rm_isis'],
+                [ospf,      'rm_ospf'],
+                ['rip 4',   'rm_rip']]
+    elsif platform == :ios_xr
+      config('route-policy my_policy', 'end-policy')
+      master = [['connected', 'my_policy'],
+                ['eigrp 1',   'my_policy'],
+                [ospf,        'my_policy'],
+                ['static',    'my_policy']]
+      master.push(['isis abc', 'my_policy']) if dbg.include? 'default'
```
glennmatthews, line 745: `if af.vrf == 'default'`?

tests/test_bgp_af.rb
robert-w-gries, line 40: @glennmatthews Good call regarding Metrics/LineLength.  I was confusing it with the Metrics/MethodLength cop

lib/cisco_node_utils/bgp_af.rb
robert-w-gries, line 145: I suggested removing the `route_map.strip!` and adding `kind: string` to the YAML entry because `config_get` would automatically strip the white space.

glennmatthews, line 145: Not sure I follow - this is on the setter side, not the getter side.

robert-w-gries, line 145: Ah my mistake. @brturn can you add those two lines back in?

tests/test_bgp_af.rb
brturn, line 447: Charles an I looked at this one before I understood the default_values / auto_default semantics.  I think I could fix it now, but would prefer to push to US59615.

glennmatthews:
I like the method_missing idea (though you also need to add `respond_to?`) but it's more than a bit of scope creep with this PR. :)

Can you provide latest minitest results against both XR and Nexus? If they look good, :+1: to merge this and move on.

tests/test_bgp_af.rb
brturn, line 553: Passing the ASN as a string is very common in this file, not sure which is the preferred method?  Should it always be passed as a number instead?

tests/test_bgp_af.rb
brturn, line 637: Would prefer to push new work to US59615.

tests/test_bgp_af.rb
glennmatthews, line 553: Ideally both should work (string is sometimes needed to support ASDOT notation) but it's not critical especially for this PR. Just seemed inconsistent that this function takes `'55'` while the function below takes `55`. Not blocking.

lib/cisco_node_utils/bgp_af.rb
brturn, line 636: I've been backing out the policy/map split slowly as I make edits.  I think Rich's proposal will work fine (just pass policy/map as a string to the setter).  I would prefer to push additional work to US59615.

brturn:
Minitest for Nexus and XR passes with some skips.

glennmatthews:
Are all of the skips explicitly triggered by your test code? (I.e., have you picked up the changes from #159 already)?

chburket:
I ran Bryan's latest code against my updated bgp_af beaker tests, and it passed on XR.  Running into a few issues on nexus involving route_target properties, but it looks like those are due to my changes, since the old beaker tests still run fine.

I'm good with merging and then fixing anything that pops up later.

:+1:

glennmatthews:
@brturn I'll merge as soon as I get a response on my last question - thanks for working through all this!

brturn:
I have not merged #159, do I need to rebase to the latest feature/grpc for that?

Currently, the skips come from two sources; an explicit 'skip' command in the test_bgp_af test function, or from executing a command for which _exclude: [cli_ios_xr] has been specified in the YAML.

glennmatthews:
With the change made in #159, only the first source (explicit `skip()`) is valid - the second case (trying to execute an unsupported feature) will be reported as an error and you need to update your test to catch it.

However, I'm going to go ahead and merge this PR as-is and you can fix any errored test cases as part of your next user story.

PR #139, opened 2015-12-17
BUGFIX: Ace.aces hash afi key was missing

chrisvanheuveln:
The Acl.acls code has the same logic but was already correct.

Minitest: 2 runs, 28 assertions, 0 failures, 0 errors, 0 skips

### DIFF
--- a/tests/test_ace.rb
+++ b/tests/test_ace.rb
@@ -16,8 +16,8 @@ require_relative 'ciscotest'
 require_relative '../lib/cisco_node_utils/acl'
 require_relative '../lib/cisco_node_utils/ace'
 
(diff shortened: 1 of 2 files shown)


tests/test_ace.rb
```diff
@@ -108,7 +108,7 @@ def create_destroy_ace(afi, entry)
     afi_cli = Acl.afi_cli(afi)
     all_aces = Ace.aces
     found = false
-    all_aces[acl_name].each do |seqno, _inst|
+    all_aces[afi][acl_name].each do |seqno, _inst|
```
RichWellum, line 111: :+1:

PR #146, opened 2015-12-21
Add command line options to update-hooks

jamtaylo:
The previous behavior of this script displayed, via git diff, all
updates to hooks installed by the script.  This requires user
interaction to page through the diff.  This was especially cumbersome on
the initial run, when the full contents of all hooks were displayed,
requiring the user to <spacebar> through pages of text.

This commit adds simple command line option parsing and recognizes the
following options:

```
--help         displays usage text

-v, --verbose  verbose output, including diffs of updated files
               (the diffs are no longer displayed by default)

--dry-run      Do not make any updates (but print the same messages
               that are normally printed to notify the user of which
               files are to be updated)
```

### DIFF
--- a/bin/git/update-hooks
+++ b/bin/git/update-hooks
@@ -42,6 +42,7 @@ while [ "$#" -gt 0 ]; do
             ;;
         (*)
             echo "Unrecognized option: $1"
+            usage
             exit 1
             ;;
     esac
--- a/bin/git/update-hooks
+++ b/bin/git/update-hooks
@@ -20,7 +20,7 @@ usage() {
     echo "Options:"
     echo "  --dry-run       Do not make any updates"
     echo "  -v  --verbose   Verbose output, including diffs of updated files"
-    echo "  --help          Display this help text"
+    echo "  -h  --help      Display this help text"
 }
 
 # Flags that can be set from the command line, with default values
@@ -30,7 +30,7 @@ verbose=0
 # parse command line options
 while [ "$#" -gt 0 ]; do
     case "$1" in
-        (--help)
+        (-h|--help)
             usage
             exit 0
             ;;
--- a/bin/git/update-hooks
+++ b/bin/git/update-hooks
@@ -12,6 +12,43 @@ HOOK_DIR="$(git rev-parse --show-toplevel)/.git/hooks"
 
 . "$SRC_DIR/hook_lib"
 
+usage() {
+    echo "Usage: update-hooks [OPTION]"
+    echo "Update the hooks in the local git repository to match those"
+    echo "under version control."
+    echo
+    echo "Options:"
+    echo "  --dry-run       Do not make any updates"
+    echo "  -v  --verbose   Verbose output, including diffs of updated files"
+    echo "  --help          Display this help text"
+}
+
+# Flags that can be set from the command line, with default values
+make_updates=1  # the inverse of --dry-run
+verbose=0
+
+# parse command line options
+while [ "$#" -gt 0 ]; do
+    case "$1" in
+        (--help)
+            usage
+            exit 0
+            ;;
+        (-v|--verbose)
+            verbose=1
+            ;;
+        (--dry-run)
+            make_updates=0
+            ;;
+        (*)
+            echo "Unrecognized option: $1"
+            exit 1
+            ;;
+    esac
+    shift 1
+done
+
+
 install_hook_wrapper() {
     local hook=$1
     # Does the symlink exist?
@@ -19,11 +56,15 @@ install_hook_wrapper() {
         # If a hook already exists and is a file:
         if [ -f $HOOK_DIR/$hook ]; then
             step_name "Moving existing '$hook' script to '$hook-local'"
-            mv $HOOK_DIR/$hook $HOOK_DIR/$hook-local
+            if [ $make_updates == 1 ]; then
+                mv $HOOK_DIR/$hook $HOOK_DIR/$hook-local
+            fi
         fi
         # create the symlink to our wrapper script
         step_name "Installing hook wrapper script for '$hook'"
-        ln -s -f ./hooks-wrapper $HOOK_DIR/$hook
+        if [ $make_updates == 1 ]; then
+            ln -s -f ./hooks-wrapper $HOOK_DIR/$hook
+        fi
     fi
 }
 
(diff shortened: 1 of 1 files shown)


bin/git/update-hooks
```diff
@@ -12,6 +12,43 @@ HOOK_DIR="$(git rev-parse --show-toplevel)/.git/hooks"
...
+            exit 0
+            ;;
+        (-v|--verbose)
+            verbose=1
+            ;;
+        (--dry-run)
+            make_updates=0
+            ;;
+        (*)
+            echo "Unrecognized option: $1"
```
robert-w-gries, line 52: Can you add a call to `usage` so the user sees what options are valid?

bin/git/update-hooks
```diff
@@ -32,14 +73,22 @@ check_update() {
     local dest=$2
     if [ ! -e "$dest" ] ; then
         step_name "Creating new file at '$dest':"
-        git diff --color --no-index /dev/null $src
-        cp $src $dest
+        if [ $verbose == 1 ]; then
```
RichWellum, line 85: As the next few lines are common to both sections - maybe create a small fn to do this?

bin/git/update-hooks
```diff
@@ -19,11 +56,15 @@ install_hook_wrapper() {
         # If a hook already exists and is a file:
         if [ -f $HOOK_DIR/$hook ]; then
             step_name "Moving existing '$hook' script to '$hook-local'"
-            mv $HOOK_DIR/$hook $HOOK_DIR/$hook-local
+            if [ $make_updates == 1 ]; then
```
RichWellum, line 68: I think it's just a string, so safer to do "$x" == "1". Kind of why I prefer true/false over 1/0 - more self-documenting.

RichWellum, line 68: Ni: bash notation if [[ .. ]]; then

bin/git/update-hooks
RichWellum, line 52: Nit: prefer to just show the usage() here as this forces me to do a --help.

bin/git/update-hooks
```diff
@@ -12,6 +12,43 @@ HOOK_DIR="$(git rev-parse --show-toplevel)/.git/hooks"
...
+}
+
+# Flags that can be set from the command line, with default values
+make_updates=1  # the inverse of --dry-run
+verbose=0
+
+# parse command line options
+while [ "$#" -gt 0 ]; do
+    case "$1" in
+        (--help)
```
RichWellum: -h ?

robert-w-gries:
Suggestion:  use the `--no-pager` option to remove the user interaction while printing the diff.

``` diff
diff --git a/bin/git/update-hooks b/bin/git/update-hooks
index e782d7f..a91ea12 100755
--- a/bin/git/update-hooks
+++ b/bin/git/update-hooks
@@ -74,7 +74,7 @@ check_update() {
     if [ ! -e "$dest" ] ; then
         step_name "Creating new file at '$dest':"
         if [ $verbose == 1 ]; then
-            git diff --color --no-index /dev/null $src
+            git --no-pager diff --color --no-index /dev/null $src
         fi
         if [ $make_updates == 1 ]; then
             cp $src $dest
@@ -84,7 +84,7 @@ check_update() {
         if [ $? -ne 0 ] ; then
             step_name "File '$dest' will be updated: "
             if [ $verbose == 1 ]; then
-                git diff --color --no-index $dest $src
+                git --no-pager diff --color --no-index $dest $src
             fi
             if [ $make_updates == 1 ]; then
                 cp -f $src $dest
```

bin/git/update-hooks
jamtaylo, line 52: Yes, of course.  I had meant to do that but it didn't make it from my brain to my editor.  Fixed in  9182afd

robert-w-gries:
:+1:  from my end

bin/git/update-hooks
jamtaylo, line 85: I could, but I think doing so would end up with more lines of code total (because the helper function should bind src and dest locally for documentation purposes), and also could make it harder to determine from the diffs which parts of of the script are changing and which parts are not.

In general, I don't usually feel bad about having two small sections of code be the same.  If it were more than a few lines, or if there were more repetitions then I would definitely be on board for refactoring.  I'm ambivalent about this one - but I'll change it if that's the consensus.

bin/git/update-hooks
jamtaylo: Sure, why not.  Added in 245955d

bin/git/update-hooks
jamtaylo, line 68: I believe you are correct that bash is treating them as strings.  To compare them as numbers it would need the `-eq` operator or to use the arithmetic evaluation context `(( ... ))`.  But, the variables can only contain the literal values assigned by the script, so there's nothing unsafe about it unless someone edits the script and breaks it.  If we have an actual standard then I'll update to match.

What's the rational for preferring the `[[` compound command over the `[` or `test` builtin?  (I just matched the style already in use by the other tests in the file, which I suspect are actually buggy and will fail on files with spaces in the filename.  But that's not an issue when comparing the known values 0 and 1.)

jamtaylo:
Pager behavior can be controlled via git configuration or the GIT_PAGER environment variable.

```
GIT_PAGER
    This environment variable overrides $PAGER. If it is set to an empty string
    or to the value "cat", Git will not launch a pager. See also the core.pager option
    in git-config(1).
```

Should the update_hooks script take it upon itself to override the user's preferences?

robert-w-gries:
As a user, I would expect the script to finish unless it needed user input for a special circumstance.

Having the script stop every time it sees a new/updated git-hook is more of a nuisance than a help.  And as of now, the only way to avoid this behavior is to disable the pager in the repo's `.git/config`.

bin/git/update-hooks
RichWellum, line 68: Both of these are just bash standards I'm used to, just for safety reasons. For example [[ builtin operator additionally provides compound commands and regular expression matching - like if [[ string =~ regex ]]. I doubt we'll add much bash to these scripts  - but I like to do it as suggested because sooner or later it gets copied and then as you've pointed out it introduces bugs. No biggy.

bin/git/update-hooks
RichWellum, line 85: No that's fine if you feel it's ok. Another route would be to pass in whether it is a new file or not and switch accordingly. Again probably more work discussing this than actually doing it. :)

bin/git/update-hooks
```diff
@@ -32,14 +74,22 @@ check_update() {
...
         if [ $? -ne 0 ] ; then
             step_name "File '$dest' will be updated: "
-            git diff --color --no-index $dest $src
-            cp -f $src $dest
+            if [ $verbose == 1 ]; then
+                git diff --color --no-index $dest $src
+            fi
+            if [ $make_updates == 1 ]; then
+                cp -f $src $dest
+            fi
```
RichWellum, line 100: :+1:

jamtaylo:
@robert-w-gries : The point of the exercise was to make the default behavior not pause (by making it not show the diffs at all unless you use the `-v` flag), but it's easy enough to add an option to explicitly enable or disable the pager.  See  980116b.

PR #148, opened 2015-12-22
fix for test

andy91315:
1. enable feature bgp to get the list of vni


lib/cisco_node_utils/evpn_vni.rb
```diff
@@ -37,6 +37,7 @@ def initialize(vni, instantiate=true)
     # Creat a hash of all vni instance
     def self.vnis
       hash = {}
+      RouterBgp.enable
```
chrisvanheuveln: This is wrong. You should not be enabling a feature as part of a get-all-instances method. If the feature is disabled then the method just returns an empty hash.

PR #153, opened 2015-12-23
Error if keyword substitution yields empty result

jamtaylo:
When the command_reference config_get_token and config_set methods use
keyword-based parameter substitution, array elements that contain a
parameter not supplied by the caller will be omitted from the result
array.  This behavior is desired to support, e.g., CLI interaction
involving VRFs, where both the presence or absence of a VRF are
permissible and should result in CLI interaction either containing or
not containing a vrf statement.  However, this behavior makes it easy
for developer oversight or typo to result in an empty command line
(which is bad).

This commit updates command_reference so that if the result of
keyword-based parameter substitution is an empty array then an
ArgumentError will be raised.  It also adds new assertions to test this
behavior to the existing test_config_get_token_hash_substitution test
case, and updates the documentation and changelog.

### DIFF
--- a/lib/cisco_node_utils/command_reference.rb
+++ b/lib/cisco_node_utils/command_reference.rb
@@ -134,7 +134,7 @@ module Cisco
     def printf_substitutor(config_key, value)
       arg_c = value.join.scan(/%/).length
       lambda do |*args|
-        unless args.length == arg_count
+        unless args.length == arg_c
           fail ArgumentError,
                "Given #{args.length} args, but #{config_key} requires #{arg_c}"
         end
--- a/lib/cisco_node_utils/command_reference.rb
+++ b/lib/cisco_node_utils/command_reference.rb
@@ -107,6 +107,9 @@ module Cisco
             end
             result.push(line) unless /<\S+>/.match(line)
           end
+          if result.empty?
+            fail ArgumentError, "Arguments given to #{key} yield empty result"
+          end
           preprocess_value(result)
         end
       elsif value.any? { |item| item.is_a?(String) && /%/ =~ item }
--- a/lib/cisco_node_utils/command_reference.rb
+++ b/lib/cisco_node_utils/command_reference.rb
@@ -98,38 +98,51 @@ module Cisco
       return unless value.is_a?(Array)
       if value.any? { |item| item.is_a?(String) && /<\S+>/ =~ item }
         # Key-value substitution
-        define_singleton_method key.to_sym do |**args|
-          result = []
-          value.each do |line|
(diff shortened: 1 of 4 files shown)


tests/test_command_reference.rb
```diff
@@ -381,7 +385,15 @@ def test_config_get_token_hash_substitution
...
-    # TODO: add negative tests?
+    ref = reference.lookup('test', 'test2')
+    token = ref.config_get_token(val1: '1', val2: '2')
+    assert_equal(['abc 1 def', 'xyz 2'], token)
+    token = ref.config_get_token(val1: '1', extra_val: 'asdf')
+    assert_equal(['abc 1 def'], token)
+    token = ref.config_get_token(val2: '2')
+    assert_equal(['xyz 2'], token)
+    assert_raises(ArgumentError) { token = ref.config_get_token }
+    assert_raises(ArgumentError) { token = ref.config_get_token(extra: 'x') }
```
chrisvanheuveln, line 396: Please add error messages to the `assert_raises`

chrisvanheuveln:
Code looks good. I suspect this change might break a few providers -- the other day I noticed that one of the vxlan setters was relying on this behavior to act as a noop (I have a fix for that in review) -- so we should probably have a full regression after this change goes in. We'll need one soon anyway for the next release so I don't think we need to require it for this merge.

robert-w-gries:
Should this PR be merged into `develop` instead of `feature/grpc`?  I don't see any grpc specific code

chrisvanheuveln:
@robert-w-gries My only hesitation on a `develop` drop is the aforementioned lack of full regression, but I think the risk is low.

lib/cisco_node_utils/command_reference.rb
```diff
@@ -130,6 +109,43 @@ def define_getter(key, value)
...
+        preprocess_value(result)
+      end
+    end
+
+    # curried function to define a getter method body that performs
+    # printf-style substitution
+    def printf_substitutor(config_key, value)
+      arg_c = value.join.scan(/%/).length
+      lambda do |*args|
+        unless args.length == arg_count
```
robert-w-gries: s/arg_count/arg_c

jamtaylo: good catch; fixed in 96eaaea
(This is what happens when line-length requirements lead me to rename variables to shorter names.)

mikewiebe:
@robert-w-gries @chrisvanheuveln, I agree that we should put this directly into `develop` but I would recommend we get a full node_utils regression to make sure all is well.  Just make sure we are same as baseline.

jamtaylo:
Closing this PR in order to open one against `develop` instead.

PR #155, opened 2016-01-06
add disable_policy_batching to cisco_bgp

andy91315:
1. pass minitest
2. pass static analysis

disable_policy_batching
disable_policy_batching ipv4 prefix-list xx
disable_policy_batching ipv6 prefix-list yy

I seperate first one as boolean property second and third as an array

### DIFF
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -339,11 +339,12 @@ module Cisco
     end
 
     def disable_policy_batching_prefix=(should_list)
-      delta_hash = Utils.delta_add_remove(should_list, disable_policy_batching_prefix)
+      delta_hash = Utils.delta_add_remove(should_list,
+                                          disable_policy_batching_prefix)
       return if delta_hash.values.flatten.empty?
       [:add, :remove].each do |action|
-        CiscoLogger.debug("disable_policy_batching_prefix delta #{@get_args}\n #{action}: " \
-                          "#{delta_hash[action]}")
+        CiscoLogger.debug("disable_policy_batching_prefix delta #{@get_args}" \
+                          " #{action}: #{delta_hash[action]}")
         delta_hash[action].each do |afi, prefix_list|
           # inject & exist are mandatory, copy is optional
           state = (action == :add) ? '' : 'no'
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -330,35 +330,52 @@ module Cisco
       config_get_default('bgp', 'disable_policy_batching')
     end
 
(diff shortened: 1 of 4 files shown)


lib/cisco_node_utils/bgp.rb
```diff
@@ -313,6 +313,54 @@ def default_confederation_id
...
+    end
+
+    def default_disable_policy_batching
+      config_get_default('bgp', 'disable_policy_batching')
+    end
+
+    # disable-policy-batching
+    # disable-policy-batching <afi> prefix-list <prefix_list>
+    #
+    def disable_policy_batching_prefix
```
chrisvanheuveln: The nested array unnecessarily complicates this property. This should really be two properties:

`disable_policy_batching_ipv4` and `disable_policy_batching_ipv6`, which simply take the prefix-list name string as inputs.

This unfortunately means a rewrite of ~80% of your code but it's a better approach.

glennmatthews: Does this property only apply to ipv4/ipv6, or are we going to eventually proliferate into `disable_policy_batching_l2vpn`, `disable_policy_batching_vpnv4`, etc?

chrisvanheuveln: Good question. I've only seen AFI options for ipv4/ipv6 but I suppose it could be extended at some point. Since they're all distinct clis I don't see a problem with just adding add'l properties as they become available.

lib/cisco_node_utils/bgp.rb
```diff
@@ -313,6 +313,71 @@ def default_confederation_id
...
+      @set_args[:state] = (enable ? '' : 'no')
+      config_set('bgp', 'disable_policy_batching', @set_args)
+      set_args_keys_default
+    end
+
+    def default_disable_policy_batching
+      config_get_default('bgp', 'disable_policy_batching')
+    end
+
+    # disable-policy-batching ipv4
```
chrisvanheuveln: This comment does not apply, just remove it:

```
-    # disable-policy-batching ipv4
```

lib/cisco_node_utils/bgp.rb
```diff
@@ -313,6 +313,71 @@ def default_confederation_id
...
+      end
+      config_set('bgp', 'disable_policy_batching_ipv4', @set_args)
+      set_args_keys_default
+    end
+
+    def default_disable_policy_batching_ipv4
+      config_get_default('bgp', 'disable_policy_batching_ipv4')
+    end
+
+    # disable-policy-batching ipv6
```
chrisvanheuveln: ditto

chrisvanheuveln:
:+1 after minor comments addressed.

PR #160, opened 2016-01-07
fix "weight" attribute setting.

HelenHZhang:
fix the broken beaker tests for "weight" for nexus.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/bgp_neighbor_af.yaml
+++ b/lib/cisco_node_utils/cmd_ref/bgp_neighbor_af.yaml
@@ -221,7 +221,6 @@ unsuppress_map:
 
 weight:
   kind: int
-  auto_default: false
   config_get_token_append: '/^weight (\d+)$/'
   config_set_append: ' <state> weight <int>'
   default_value: false
--- a/tests/test_bgp_neighbor_af.rb
+++ b/tests/test_bgp_neighbor_af.rb
@@ -758,13 +758,15 @@ class TestRouterBgpNeighborAF < CiscoTestCase
   end
 
   def weight(af, dbg)
-    af.weight = 22
-    assert_equal(22, af.weight, "Test 1. #{dbg} Failed to set weight")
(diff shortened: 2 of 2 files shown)


robert-w-gries:
Should this PR be merged into `develop` instead?

chburket:
@robert-w-gries 

We're not sure if this is broken in develop.  There are other differences in the develop version of these files.  If a similar problem pops up when the full regression suite is run on develop, we can make a similar fix then.

mikewiebe:
@chburket I verified that is is broken in `develop` so I agree with @robert-w-gries that it would be better to make the fix in develop and then merge develop into `feature/grpc`.

HelenHZhang:
@robert-w-gries 
@mikewiebe 
I'll switch to develop branch, try the same fix, run all tests.
test_bgp_neighbor_af.rb is very different from develop branch and feature/grpc.

In develop branch, there is some n6k_platform, nk7_platform stuff, and no test_weight at all.
In feature/grpc branch, there is just "platform", and some xr stuff.

With just yaml file change, the beaker test passed at the test step
- TestStep :: 1.1 Default Properties :: IDEMPOTENCE" passed.

However, beaker tests failed at the test step:
- TestStep :: 5.12.1 Non Default Misc Map commands Part 1 [ensure => present] :: MANIFEST"

mikewiebe:
@HelenHZhang Just fix the weight issue and we will take care of the additional beaker failures.

HelenHZhang:
forgot adding test results:
The following tests passed.
1). minitest for test_bgp_neighbor_af on nexus
2). minitest for test_bgp_neighbor_af on xr
3). beaker tests for test_bgpneighborafon nexus

glennmatthews:
@HelenHZhang Since #162 is merged, should we cancel this PR and just plan to merge latest develop code into feature/grpc?

chburket:
@glennmatthews Seems reasonable to me.  I'm sure Helen will be ok with it.

PR #161, opened 2016-01-07
change event_history

andy91315:
1. pass minitest
2. pass static analysis

munge true to default_value for each property

### DIFF
--- a/tests/test_router_bgp.rb
+++ b/tests/test_router_bgp.rb
@@ -341,37 +341,33 @@ class TestRouterBgp < CiscoTestCase
     opts = [:cli, :detail, :events, :periodic]
     opts.each do |opt|
       # Test basic true
-      bgp.send("event_history_#{opt}_set", true)
+      bgp.send("event_history_#{opt}=", 'true')
+      set = bgp.send("default_event_history_#{opt}")
       result = bgp.send("event_history_#{opt}")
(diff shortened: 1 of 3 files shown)


tests/test_router_bgp.rb
```diff
@@ -341,37 +341,33 @@ def test_event_history
...
-      result = bgp.send("event_history_#{opt}_size")
-      assert_equal('large', result,
+      bgp.send("event_history_#{opt}=", 'size_large')
+      result = bgp.send("event_history_#{opt}")
+      assert_equal('size_large', result,
                    'Failed to set True with Size')
 
       # Test false with size
-      bgp.send("event_history_#{opt}_set", false)
+      bgp.send("event_history_#{opt}=", 'false')
```
glennmatthews, line 357: How difficult would it be to support the literal value `false` as an input as well as the string `'false'`?

andy91315, line 357: It is designed as a string property. I do not think it can support both. If I set no event_history cli with false or 'false', when get it, it does not know whether to return false or 'false'. Then it does not match in the puppet.

glennmatthews, line 357: It should be possible but this is non-blocking as far as I'm concerned.

glennmatthews:
:+1:

PR #165, opened 2016-01-12
Fix feature nv overlay in tests

robert-w-gries:
Skip applicable tests when unable to set `feature nv overlay`.

All tests pass on platforms that support `nv overlay` and also on platforms that do not.

New behavior for platforms that do no support `nv overlay`:

```
~/cisco-network-node-utils$ ruby tests/test_vni.rb -v
Run options: -v --seed 37891

# Running:


Node under test:
  - name  - n3k-103
  - type  - N3K-C3048TP-1GE
  - image - 

TestVni#test_mt_full_shutdown = 2.43 s = S
TestVni#test_mt_full_vni_create_destroy = 0.97 s = S
TestVni#test_mt_lite_mapped_vlan = 0.98 s = S

Finished in 4.383573s, 0.6844 runs/s, 0.0000 assertions/s.

  1) Skipped:
TestVni#test_mt_full_shutdown [tests/test_vni.rb:27]:
Platform does not support nv overlay feature


  2) Skipped:
TestVni#test_mt_full_vni_create_destroy [tests/test_vni.rb:27]:
Platform does not support nv overlay feature


  3) Skipped:
TestVni#test_mt_lite_mapped_vlan [tests/test_vni.rb:27]:
Platform does not support nv overlay feature
```

### DIFF
--- a/lib/cisco_node_utils/bgp_af.rb
+++ b/lib/cisco_node_utils/bgp_af.rb
@@ -237,8 +237,7 @@ module Cisco
 
     # advertise_l2vpn_evpn
     def advertise_l2vpn_evpn
-      Feature.feature_nv_overlay_enable unless
-        Feature.feature_nv_overlay_enabled?
+      Feature.nv_overlay_enable unless Feature.nv_overlay_enabled?
       return false unless RouterBgpAF.feature_nv_overlay_evpn_enabled
       config_get('bgp_af', 'advertise_l2vpn_evpn', @get_args)
     end
--- a/lib/cisco_node_utils/bgp_af.rb
+++ b/lib/cisco_node_utils/bgp_af.rb
@@ -66,26 +66,6 @@ module Cisco
       config_set('bgp', 'address_family', @set_args)
     end
 
-    def feature_nv_overlay_enabled?
-      config_get('feature', 'nv_overlay')
-    rescue Cisco::CliError => e
-      # cmd will syntax when feature is not enabled.
-      raise unless e.clierror =~ /Syntax error/
-      return false
-    end
-
-    def feature_nv_overlay_enable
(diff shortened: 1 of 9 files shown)


lib/cisco_node_utils/bgp_af.rb
```diff
@@ -57,6 +57,7 @@ def self.afs
     end
 
     def create
+      feature_nv_overlay_enable unless feature_nv_overlay_enabled?
```
chrisvanheuveln: This does not feel safe. You're enabling nv overlay at bgp address-family creation time regardless of whether the manifest actually needs this feature or the platform even supports it. Won't this raise an error on some platforms?

In other words, I think this feature call should happen in the affected properties themselves.

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -66,6 +67,26 @@ def destroy
       config_set('bgp', 'address_family', @set_args)
     end
 
+    def feature_nv_overlay_enabled?
+      config_get('feature', 'nv_overlay')
```
chrisvanheuveln: I like the idea of moving these into `feature.yaml` but I don't like all the duplication in the different providers. 
I'd prefer to have a feature provider where these can all be kept centrally and then just call `Feature.nv_overlay_enabled?` instead. Is that doable?

robert-w-gries: Sounds like a good idea to me. Will start working on it

lib/cisco_node_utils/bgp_af.rb
robert-w-gries: Moved check to `advertise_l2vpn_evpn` property

PR #167, opened 2016-01-12
RD & route-target properties move to vrf

chrisvanheuveln:
- route-distinguisher is now supported in both vrf and bgp
- route-target props have moved completely into vrf_af (new provider)
- vrf.yaml now template-tized
- changed feature calls to use new Feature class
- minitests updated
- feature calls moved to feature.rb

### DIFF
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 require_relative 'node_util'
+require_relative 'feature'
 require_relative 'bgp_af'
 
 module Cisco
@@ -73,18 +74,6 @@ module Cisco
       return {}
     end
 
-    def self.enabled
-      config_get('bgp', 'feature')
-    rescue Cisco::CliError => e
-      # cmd will syntax reject when feature is not enabled
-      raise unless e.clierror =~ /Syntax error/
-      return false
-    end
-
-    def self.enable(state='')
-      config_set('bgp', 'feature', state: state)
-    end
-
     # Convert BGP ASN ASDOT+ to ASPLAIN
     def self.dot_to_big(dot_str)
       fail ArgumentError unless dot_str.is_a? String
@@ -99,17 +88,7 @@ module Cisco
       high_bits + low_bits
     end
 
-    def router_bgp(asnum, vrf, state='')
-      # Only one bgp autonomous system number is allowed
-      # Raise an error if one is already configured that
-      # differs from the one being created.
-      configured = config_get('bgp', 'router')
-      if !configured.nil? && configured.to_s != asnum.to_s
-        fail %(
-          Changing the BGP Autonomous System Number is not allowed.
-          Current BGP asn: #{configured}
-          Attempted change to asn: #{asnum})
-      end
+    def router_bgp(state='')
       @set_args[:state] = state
       if vrf == 'default'
         config_set('bgp', 'router', @set_args)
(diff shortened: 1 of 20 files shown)


lib/cisco_node_utils/feature.rb
```diff
@@ -19,24 +19,69 @@
...
+    # ---------------------------
+    def self.bgp_enable(state='')
+      config_set('feature', 'bgp', state: state)
+    end
+
+    def self.bgp_enabled?
+      config_get('feature', 'bgp')
     rescue Cisco::CliError => e
-      # cmd will syntax when feature is not enabled.
+      # cmd will syntax reject when feature is not enabled
```
glennmatthews: This seems weird - config_get for bgp is just calling 'show run | i feature' so why would we get a CliError in this case?

chrisvanheuveln: It's a cut&paste from the old code which was doing show run bgp. I'll remove it and the others where appropriate.

lib/cisco_node_utils/bgp.rb
```diff
@@ -119,32 +98,15 @@ def router_bgp(asnum, vrf, state='')
...
-    end
-
     # Create one router bgp instance
     def create
-      if RouterBgp.enabled
-        router_bgp(@asnum, @vrf)
-      else
-        enable_create_router_bgp(@asnum, @vrf)
-      end
+      Feature.bgp_enable unless Feature.bgp_enabled?
```
chrisvanheuveln: @glennmatthews Just had a thought about these  `unless _enabled?` checks. I'm thinking of just removing them from the callers and add the unless checks to the `_enable` methods themselves.  Is it worth changing?

edit: nm, changed it anyway.

lib/cisco_node_utils/bgp.rb
```diff
@@ -691,16 +653,15 @@ def default_reconnect_interval
...
     def route_distinguisher
-      return false unless RouterBgpAF.feature_nv_overlay_evpn_enabled
       config_get('bgp', 'route_distinguisher', @get_args)
     end
 
     def route_distinguisher=(rd)
-      RouterBgpAF.feature_nv_overlay_evpn_enable unless
-        RouterBgpAF.feature_nv_overlay_evpn_enabled
+      Feature.nv_overlay_enable
+      Feature.nv_overlay_evpn_enable
```
robert-w-gries, line 664: Nit: Since the `nv_overlay_evpn` feature depends on `nv_overlay` being enabled, you can move `nv_overlay_enable` to the `Feature.nv_overlay_evpn_enable` method

lib/cisco_node_utils/cmd_ref/feature.yaml
```diff
@@ -1,8 +1,32 @@
 # feature
 ---
+bgp:
+  kind: boolean
+  config_get: "show running | i feature"
+  config_get_token: '/^feature bgp$/'
+  config_set: "<state> feature bgp"
```
robert-w-gries: `<state>` isn't necessary since we don't need to disable features

lib/cisco_node_utils/cmd_ref/feature.yaml
```diff
@@ -1,8 +1,32 @@
...
+
+vn_segment_vlan_based:
+  # MT-lite only
+  /N(3|9)/:
+    kind: boolean
+    default_value: false
+    cli_nexus:
+      config_get: 'show running section feature'
+      config_get_token: '/^feature vn-segment-vlan-based$/'
+      config_set: 'feature vn-segment-vlan-based'
```
robert-w-gries: Why is `cli_nexus` needed here?

lib/cisco_node_utils/cmd_ref/feature.yaml
chrisvanheuveln: hey that was my comment on your PR! :-)

I'll fix.

lib/cisco_node_utils/cmd_ref/feature.yaml
chrisvanheuveln: I think it was a pre-emptive strike to help out the XR merge. But the `/N(3|9)/:` already keeps this cli out of XR. I'll remove it.

lib/cisco_node_utils/bgp.rb
chrisvanheuveln, line 664: Yeah, considered that. Also thought about maybe getting rid of nv_overlay_evpn_enable altogether and just making it part of nv_overlay_enable since it's not really a "feature" command (`nv overlay evpn`). I think I'd prefer to keep it like this until we can come up with a better story for it.

robert-w-gries:
:+1: but an extra pair of eyes wouldn't hurt

PR #169, opened 2016-01-13
Fix invalid 'feature fabric forwarding' command

robert-w-gries:
### Changes

`feature fabric forwarding` is not supported for 3k.

Additionally, `feature fabric forwarding` depends on `feature-set fabric` being installed.  VxlanGlobal provider now handles the case where `feature-set fabric` is not installed.
### Minitest Error

Currently only n9k fails minitest.

On my n9k-108, I get an error that says `install feature-set fabric` is an invalid command.  I thought that meant `fabric forwarding` is unsupported on 9k, but I found out I was still able to enable `feature fabric forwarding` while `feature-set fabric` was uninstalled.  Can anyone reproduce this on their n9k?

```
switch# show feature-set 
Feature Set Name      ID        State   
--------------------  --------  --------
fex                    3          disabled
fabric                 7          uninstalled
mpls                   4          uninstalled


switch(config)# install feature-set ?
  fex   FEX
  mpls  MPLS

switch(config)# feature fabric forwarding 
switch(config)#
```

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/feature.yaml
+++ b/lib/cisco_node_utils/cmd_ref/feature.yaml
@@ -1,6 +1,7 @@
 # feature
 ---
 fabric:
+  _exclude: [/N3/]
   config_get: "show feature-set"
   config_get_token: '/^fabric[\s\d]+(\w+)/'
   config_set: "<state> feature-set fabric"
@@ -9,7 +10,7 @@ fabric_forwarding:
   _exclude: [/N3/]
   config_get: "show running | i ^feature"
   config_get_token: '/^feature fabric forwarding$/'
-  config_set: "<state> feature fabric forwarding"
+  config_set: "feature fabric forwarding"
 
 nv_overlay:
   _exclude: [/N(5|6)/]
--- a/lib/cisco_node_utils/cmd_ref/feature.yaml
+++ b/lib/cisco_node_utils/cmd_ref/feature.yaml
@@ -1,7 +1,7 @@
 # feature
 ---
 fabric:
-  _exclude: [/N3/]
+  _exclude: [/N(3|9)/]
   config_get: "show feature-set"
   config_get_token: '/^fabric[\s\d]+(\w+)/'
   config_set: "<state> feature-set fabric"
--- a/lib/cisco_node_utils/cmd_ref/feature.yaml
+++ b/lib/cisco_node_utils/cmd_ref/feature.yaml
@@ -1,5 +1,10 @@
 # feature
 ---
+fabric:
+  config_get: "show feature-set"
+  config_get_token: '/^fabric[\s\d]+(\w+)/'
+  config_set: "<state> feature-set fabric"
+
 fabric_forwarding:
   _exclude: [/N3/]
   config_get: "show running | i ^feature"
--- a/lib/cisco_node_utils/cmd_ref/feature.yaml
+++ b/lib/cisco_node_utils/cmd_ref/feature.yaml
@@ -1,5 +1,11 @@
 # feature
 ---
+fabric_forwarding:
+  _exclude: [/N3/]
+  config_get: "show running | i ^feature"
+  config_get_token: '/^feature fabric forwarding$/'
+  config_set: "<state> feature fabric forwarding"
+
 nv_overlay:
   _exclude: [/N(5|6)/]
   kind: boolean
--- a/lib/cisco_node_utils/feature.rb
+++ b/lib/cisco_node_utils/feature.rb
@@ -34,9 +34,13 @@ module Cisco
       config_get('feature', 'fabric') !~ /^uninstalled/
     end
 
+    def self.fabric_supported?
+      config_get('feature', 'fabric')
+    end
+
     def self.fabric_forwarding_enable
       return if fabric_forwarding_enabled?
-      Feature.fabric_enable unless node.product_id =~ /N9/
+      Feature.fabric_enable if Feature.fabric_supported?
       config_set('feature', 'fabric_forwarding')
     end
 
--- a/lib/cisco_node_utils/feature.rb
+++ b/lib/cisco_node_utils/feature.rb
@@ -21,8 +21,9 @@ module Cisco
   class Feature < NodeUtil
     def self.fabric_enable
       # install feature-set and enable it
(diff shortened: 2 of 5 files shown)


lib/cisco_node_utils/vxlan_global.rb
```diff
@@ -29,17 +30,19 @@ def initialize(instantiate=true)
     end
 
     def enable
-      config_set('vxlan_global', 'feature', state: '')
+      Feature.fabric_enable unless Feature.fabric_enabled?
```
chrisvanheuveln: I made some changes to feature.rb as part of my PR. One thing I'm doing is removing the `unless Feature._enabled?` from the callers and moving them into the feature setters; ie. above would be:

```
    Feature.fabric_enable
```

and feature.rb would have:

```
    def self.fabric_enable
      return if Feature.fabric_enabled?
      config_set('feature', 'fabric')
    end
```

lib/cisco_node_utils/vxlan_global.rb
```diff
@@ -29,17 +30,19 @@ def initialize(instantiate=true)
...
+      config_set('feature', 'fabric_forwarding', state: '')
```
chrisvanheuveln: This should be another Feature call:   `Feature.fabric_forwarding_enable`

lib/cisco_node_utils/vxlan_global.rb
```diff
@@ -29,17 +30,19 @@ def initialize(instantiate=true)
...
     end
 
     def disable
-      config_set('vxlan_global', 'feature', state: 'no')
+      config_set('feature', 'fabric_forwarding', state: 'no') if
+        Feature.fabric_enabled?
```
chrisvanheuveln: We've elected to not turn off features anymore. So just remove this 'no' call altogether.

The dup_host_mac_detection_default can stay.

lib/cisco_node_utils/vxlan_global.rb
```diff
@@ -29,17 +30,19 @@ def initialize(instantiate=true)
...
       dup_host_mac_detection_default
     end
 
     # Check current state of the configuration
     def self.enabled
```
chrisvanheuveln: This should be removed and the callers should be updated to use `Feature.fabric_forwarding` instead.

lib/cisco_node_utils/vxlan_global.rb
```diff
@@ -25,29 +26,13 @@ module Cisco
...
 
-    # Check current state of the configuration
-    def self.enabled
-      feat = config_get('vxlan_global', 'feature')
-      return !(feat.nil? || feat.empty?)
-    rescue Cisco::CliError => e
-      # This cmd will syntax reject if feature is not
-      # enabled. Just catch the reject and return false.
-      return false if e.clierror =~ /Syntax error/
-      raise
```
robert-w-gries, line 48: I tried reproducing this syntax error, but I could not.  Anyone have confirmation that this is a necessary rescue?

robert-w-gries:
Refactored Feature provider and fixed vxlan_global for 9k.

lib/cisco_node_utils/cmd_ref/feature.yaml
```diff
@@ -1,5 +1,17 @@
 # feature
 ---
+fabric:
+  _exclude: [/N3/]
+  config_get: "show feature-set"
+  config_get_token: '/^fabric[\s\d]+(\w+)/'
+  config_set: "<state> feature-set fabric"
```
chrisvanheuveln, line 13: Hardcode `<state>` as `install`, since we only install/enable, never uninstall/disable

lib/cisco_node_utils/feature.rb
```diff
@@ -19,6 +19,31 @@
 module Cisco
   # Feature - node util class for managing common features
   class Feature < NodeUtil
+    def self.fabric_enable
+      # install feature-set and enable it
+      return if fabric_enabled?
+      config_set('feature', 'fabric', state: 'install') unless fabric_installed?
+      config_set('feature', 'fabric', state: '')
```
chrisvanheuveln, line 40: remvoe the `state: ''` references from these two lines

lib/cisco_node_utils/feature.rb
```diff
@@ -19,6 +19,31 @@
...
+      config_get('feature', 'fabric') =~ /^enabled/
+    end
+
+    def self.fabric_installed?
+      config_get('feature', 'fabric') !~ /^uninstalled/
+    end
+
+    def self.fabric_forwarding_enable
+      return if fabric_forwarding_enabled?
+      Feature.fabric_enable unless node.product_id =~ /N9/
```
chrisvanheuveln: Ugh. So N9 supports `feature fabric forwarding` but not `feature-set fabric` ? blech.

I don't like platform checks in the code itself unless it can't be helped. Consider doing:

```
fabric:
  _exclude: [/N(3|9)/]
  config_get: "show feature-set"
  config_get_token: '/^fabric[\s\d]+(\w+)/'
  config_set: "<state> feature-set fabric"
```

```
  def self.fabric_supported?
    config_get('feature', 'fabric')
  end

  def self.fabric_forwarding_enable
    return if fabric_forwarding_enabled?
    Feature.fabric_enable if Feature.fabric_supported?
     config_set('feature', 'fabric_forwarding')
  end
```

lib/cisco_node_utils/vxlan_global.rb
chrisvanheuveln, line 48: I think it was copied over from bgp, which originally did a 'show run bgp', which fails when the feature is off. If the yaml just checks 'show running | section feature' then it won't fail.
So yeah just delete it.

lib/cisco_node_utils/cmd_ref/feature.yaml
robert-w-gries, line 13: `<state>` is required here because the process of enabling the `fabric` feature-set is:

```
install feature-set fabric
feature-set fabric
```

If `install` was hardcoded, the feature-set wouldn't be enabled.

robert-w-gries:
Replaced the platform specific code

chrisvanheuveln:
:+1:   after you fix the conflicts

PR #170, opened 2016-01-13
Tweaks to support XR for bgp-af.

chburket:


lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -207,43 +207,50 @@ redistribute_policy:
     config_set_append: ' <state> redistribute <protocol> route-policy <policy>'
 
 #config in vrf context
+# TODO: fix these for XR
 route_target_both_auto:
```
robert-w-gries, line 211: If this property is broken for XR, you should add `_exclude: [cli_ios_xr]`

Also, why is `config_get` listed underneath `cli_nexus`?  If `cli_ios_xr` is excluded, you don't need the `cli_nexus` specification.

chburket, line 211: This was just an easy way to "comment them out" without making other changes.  Marking them excluded for XR makes the getters return nil and causes a nil error in some of the code.  I didn't want to start changing code that is currently being worked on in US59615, so I just used this work-around.

glennmatthews, line 211: It doesn't hurt to explicitly specify something for Nexus, even if xr is excluded.

robert-w-gries, line 211: @chburket ack

chrisvanheuveln, line 211: fyi, the rd / route-target props are in the process of moving from bgp_af to vrf_af in the develop branch (PR #167) so you may want to just leave a breadcrumb on them for now. We plan to pull a release branch next tuesday and then you can start the fun of merging with develop.

PR #159, opened 2016-01-07
Remove 'skip' logic in minitest and fix tests

glennmatthews:
Remove the logic in our minitest infra that was added early in gRPC development to convert an uncaught Cisco::UnsupportedException to a 'skipped' result. Update the minitests that are currently working on XR so that they do not have any such uncaught exceptions.

The `Node` extended functionality exercised in `test_node_ext` was not fully implemented for XR before, so I finished implementing the parts that are applicable to XR and added appropriate negative tests for the ones that are unsupported.

Tests run successfully against XR:
- test_interface, test_interface_svi, test_interface_switchport
- test_command_reference, test_command_config
- test_router_bgp, test_bgp_neighbor, test_bgp_neighbor_af (test_bgp_af is blocked pending the merge of #136)

PR #172, opened 2016-01-14
Explicitly exclude platforms and skip minitest

robert-w-gries:
TestVxlanVtep and TestVxlanVtepVni skip for 3k, 5k, and 6k.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
+++ b/lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
@@ -1,5 +1,7 @@
 # vxlan_vtep
 ---
+_exclude: [/N(3|5|6)/]
+
 _template:
   config_get: "show running-config interface all | section 'interface nve'"
   config_get_token: '/^interface <name>$/i'
--- a/lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
+++ b/lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
@@ -29,7 +29,7 @@ mt_full_support:
 mt_lite_support:
   # This is only used for determining support for Multi-Tenancy Lite
   kind: boolean
-  /N(3|9)/:
+  /N9/:
     default_only: true
   else:
     # this feature is always off on these platforms and cannot be changed
--- a/tests/test_vxlan_vtep_vni.rb
+++ b/tests/test_vxlan_vtep_vni.rb
@@ -49,6 +49,8 @@ class TestVxlanVtepVni < CiscoTestCase
   end
 
   def test_vnis
+    skip('Platform does not support vnis') if VxlanVtepVni.vnis['nve1'].nil?
+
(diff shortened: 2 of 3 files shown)


lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
```diff
@@ -1,5 +1,7 @@
 # vxlan_vtep
 ---
+_exclude: [/N(3|5|6)/]
```
mikewiebe, line 3: You should be able to remove the n3k reference here now right?

https://github.com/cisco/cisco-network-node-utils/blob/feature/vxlan-vtep-vni-test-fix/lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml#L32

tests/test_vxlan_vtep_vni.rb
```diff
@@ -49,6 +49,8 @@ def test_create_with_existing
   end
 
   def test_vnis
+    skip('Platform does not support vnis') if VxlanVtepVni.vnis['nve1'].nil?
```
mikewiebe, line 52: Why are you only skipping this one test case.  Shouldn't the other test cases be skipped for unsupported platforms?

lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
robert-w-gries, line 3: Yep.  Removed the reference now.

tests/test_vxlan_vtep_vni.rb
robert-w-gries, line 52: The other testcases automatically skip because they attempt to `config_set` a property that is excluded for the platform.

This testcase assumed the config line in the setup method `config('interface nve1')` worked properly, which caused the following error for unsupported platforms:

```
  5) Failure:
TestVxlanVtepVni#test_vnis [tests/test_vxlan_vtep_vni.rb:53]:
Expected nil (NilClass) to respond to #empty?.
```

glennmatthews, line 52: Drat. I'd forgotten that we added the auto-skip logic to develop as well. If at all possible, you should port #159 (probably just the changes to basetest.rb and test_command_reference.rb) to develop and fix up the resulting failures. Otherwise, we'll be on the hook for fixing this when we do the feature/grpc branch collapse.

mikewiebe, line 52: @glennmatthews  Remind me again why we don't want to auto-skip anymore?

mikewiebe, line 52: @robert-w-gries @glennmatthews.  Spoke with Rob this morning.  The amount of work needed is beyond the scope of what we need for release 1.2 which goes live on 1/29.  I am fine with doing this in develop but we need this code to collapse today so we can pull the release branch.  Any objection to this?

PR #175, opened 2016-01-19
US59615 - Fix bgp nodeutils on eXR for new attributes (cisco_bgp_af)

brturn:
## TODO:
- Move ios_xr? & nexus? to Node.rb
- default_metric (need to mark as not applicable to XR and add negative tests)
- inject_map (need to mark as not applicable to XR and add negative tests)
- suppress_inactive (need to mark as not applicable to XR and add negative tests)
- test create/delete AFs (call BGP.afs and check the value?)
- distance  (need to add XR CLI)
- table_map (need to add XR CLI)
- test_route_target
- test_dampening - skipping defaults-setting on XR due to no default values on XR?
- deleted code covered by the missing_method() metafunction
- added replies_to? metafunction


brturn:
@glennmatthews @chburket Everything from the story is done, added in most of the comments from the last story as well, and deleted code which is covered by missing_method().

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -98,52 +98,34 @@ def set_args_keys(hash={}) # rubocop:disable Style/AccessorMethodName
     #                      PROPERTIES                      #
     ########################################################
 
-    # IOS XR?
-    def ios_xr?
-      platform == :ios_xr
+    def fail_unsupported
+      fail Cisco::UnsupportedError.new('','','')
```
glennmatthews: I don't like this. Can we provide _any_ information for troubleshooting?

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -98,52 +98,34 @@ def set_args_keys(hash={}) # rubocop:disable Style/AccessorMethodName
...
     end
 
-    # Nexus?
-    def nexus?
-      platform == :nexus
-    end
+    # def vrf_unsupported_xr?
```
glennmatthews: If these are no longer needed/used, please remove them rather than just commenting out.

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -98,52 +98,34 @@ def set_args_keys(hash={}) # rubocop:disable Style/AccessorMethodName
...
-      fail Cisco::UnsupportedError.new(
-        'bgp_af', reason, 'set',
-        "#{reason} is not configurable on a " \
-        'per-VRF basis on IOS XR') if vrf_unsupported_xr?
-    end
+    # def fail_if_vrf_unsupported_xr(reason)
+    #   fail Cisco::UnsupportedError.new(
+    #     'bgp_af', reason, 'set',
+    #     "#{reason} is not configurable on a " \
+    #     'per-VRF basis on IOS XR') if vrf_unsupported_xr?
```
glennmatthews: Curious why this is removed - are there now no VRF-exclusive properties left?

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -853,59 +588,16 @@ def method_missing(*args)
...
-    #     def suppress_inactive
-    #     def suppress_inactive=(state)
-    #     def default_suppress_inactive
-    #     def table_map
-    #     def table_map_filter
-    #     def default_table_map
-    #     def default_table_map_filter
-    ##########################################
+    # Is the given name available in the YAML?
+    def responds_to?(name)
```
glennmatthews, line 581: The relevant Ruby API is `respond_to?` not `responds_to?`. See http://technicalpickles.com/posts/using-method_missing-and-respond_to-to-create-dynamic-methods/ for a pretty good walkthrough.

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -74,13 +74,14 @@ dampen_igp_metric:
 
 dampening:
   auto_default: false
+  multiple: true
```
glennmatthews: Why this change?

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
```diff
@@ -252,20 +270,23 @@ route_target_import_evpn:
...
   default_value: ""
+  cli_nexus:
+    config_get_token_append: '/^table-map (\S+)(?: filter)?$/'
+    config_set_append: '<state> table-map <map>'
+  cli_ios_xr:
+    config_get_token_append: '/^table-policy (\S+)(?: filter)?$/'
+    config_set_append: ' <state> table-policy <map>'
 
 table_map_filter:
+  _exclude: [cli_ios_xr]
```
glennmatthews, line 287: This seems odd - from table_map above it looks like filter is supported on XR?
(edit) Ah, I see - filter is _not_ supported on XR, the regexp provided above in table_map is needlessly broad. OK.

tests/test_bgp_af.rb
```diff
@@ -132,44 +137,41 @@ def setup
...
-    [:advertise_l2vpn_evpn,          :nexus,  'default', %w(ipv6 unicast),     :CliError],
-    [:advertise_l2vpn_evpn,          :nexus,  'default', %w(ipv4 multicast),   :CliError],
-    [:advertise_l2vpn_evpn,          :nexus,  'default', %w(ipv6 multicast),   :CliError],
-    [:advertise_l2vpn_evpn,          :nexus,  'red',     %w(ipv4 multicast),   :CliError],
-    [:advertise_l2vpn_evpn,          :nexus,  'red',     %w(ipv6 multicast),   :CliError],
+    [:additional_paths_install,      :nexus,  :any,      :ipv6,                :CliError],
+    [:advertise_l2vpn_evpn,          :nexus,  'default', :any,                 :CliError],
+    [:advertise_l2vpn_evpn,          :nexus,  'red',     :multicast,           :CliError],
+    [:inject_map,                    :nexus,  :any,      :multicast,           :CliError],
+    [:route_target_both_auto,        :nexus,  :any,      :any,                 :CliError],
```
glennmatthews, line 169: So the idea with here plus the above is that the route-target commands are only supported in a VRF on Nexus?

tests/test_bgp_af.rb
```diff
@@ -132,44 +137,41 @@ def setup
...
-
-    [:maximum_paths_ibgp,            :ios_xr, :any,      %w(vpnv4 multicast),  :unsupported],
-    [:maximum_paths_ibgp,            :ios_xr, :any,      %w(vpnv6 multicast),  :unsupported],
     [:dampen_igp_metric,             :ios_xr, :any,      :any,                 :unsupported],
+    [:default_metric,                :ios_xr, :any,      :any,                 :unsupported],
+    [:route_target_both_auto,        :ios_xr, :any,      :any,                 :unsupported],
+    [:route_target_both_auto_evpn,   :ios_xr, :any,      :any,                 :unsupported],
+    [:route_target_import_evpn,      :ios_xr, :any,      :any,                 :unsupported],
+    [:route_target_export_evpn,      :ios_xr, :any,      :any,                 :unsupported],
+    [:inject_map,                    :ios_xr, :any,      :any,                 :unsupported],
```
glennmatthews: What's the ordering of these entries? Can we alphabetize them or something?

tests/test_bgp_af.rb
```diff
@@ -295,16 +316,25 @@ def test_collection_empty
...
+
+    bgp_afs = []
+    %w(default red blue orange black).each do |vrf|
+      [%w(ipv4 unicast), %w(ipv6 unicast), %w(ipv4 multicast), %w(ipv6 multicast)].each do |af|
+        config_ios_xr_dependencies(55, vrf)
+        bgp_afs.push(RouterBgpAF.new(55, vrf, af))
+      end
+    end
+
+    # config('router bgp 55',
```
glennmatthews, line 329: delete?

tests/test_bgp_af.rb
```diff
@@ -410,26 +425,47 @@ def test_dampening
...
+    #############################################
+    # Set and verify 'dampening' to defaults    #
+    #############################################
+
+    bgp_af.dampening = bgp_af.default_dampening
+
+    # Check getters
+
+    # TODO: one returns a list of integers, the other a list of numbers
+    #       thus they don't match and the assert fails... should fix
```
glennmatthews: Yup - this needs to be fixed. :-)

tests/test_bgp_af.rb
```diff
@@ -1019,35 +810,37 @@ def table_map(asn, vrf, af)
     #
     val = 'sjc'
     bgp_af.table_map_set(val)
-    assert_equal(val, bgp_af.table_map,
-                 'Error: default metric value does not match set value')
+    assert_equal(val, bgp_af.table_map)
```
glennmatthews, line 812: Why did you remove the explanatory messages in these checks?
(edit) Ah, I see - these are incorrect messages, copy-and-paste errors. Still, I'd prefer to fix them rather than removing them.

glennmatthews:
@brturn I'm going to merge part of these changes as part of the sync with latest develop content (which moves the route_target properties into Vrf). I'll work with you to address the merge conflicts tomorrow.

lib/cisco_node_utils/bgp_af.rb
brturn: Driving towards metafunctions; the vrf unsupported exceptions are more compactly represented in a table rather than individual functions with individual fail calls.  This will also allow merging the test matrix with the base code - previously the tests replicated the base code.  Now that I've drawn most of the data into the tests, the structure of the command tree can be queried and the appropriate result generated.  Right now, that means throwing a CliError for almost everything, but that would be easy to change.

lib/cisco_node_utils/bgp_af.rb
brturn, line 581: Reminds me that I need to write a test function for this!

lib/cisco_node_utils/cmd_ref/bgp_af.yaml
brturn: It's an error; thought 'multiple' meant 'returns a list', but that is not quite correct.

tests/test_bgp_af.rb
brturn, line 169: Currently, it appears that "route_target_.." works only on Nexus, in a VRF, under Unicast modes.

tests/test_bgp_af.rb
brturn: Will do.

tests/test_bgp_af.rb
brturn, line 329: @chburket - You suggested I add this function; is this what you were thinking about, or did I miss the mark?

tests/test_bgp_af.rb
brturn, line 812: Basically, all the assert messages were just typing out the same information that gets printed out when the assert fails.  This seems wasteful to me.  In some cases, the failing-assert message was incorrect and I think the original author added a specific message to correct the auto-generated messages.  Actually, it just needed the order of the parameters to be swapped to make the auto-generated message correct.  I've done this for all the asserts I've come across so far, and I prefer the auto-generated messaging since it can be internationalized more easily.

brturn:
Oops!  Forgot to run this latest one on the Nexus..  I'll start it before I leave tonight.

glennmatthews:
I resolved the merge conflicts and ran test_bp_af against XR and NX and it's passing - merged. Thanks for seeing this through to completion!

lib/cisco_node_utils/bgp_af.rb
glennmatthews, line 581: BTW, the Puppet autogen code heavily relies on calling `respond_to?` in the NU class, so getting this working properly is a pre-requisite for getting Puppet cisco_bgp_af working again. I'm fixing this up and should have a commit ready soon.

PR #176, opened 2016-01-19
Major refactor - YAML, XR CLI, and more

glennmatthews:
Major changes to the YAML keywords:

| Before | After | Comments |
| --- | --- | --- |
| config_get | get_command |  |
| config_get_token / config_get_token_append | get_context / get_value | "Context" is used to filter down to the "value". Both template and specific properties can define context and value - template is no longer 'special' |
| config_set / config_set_append | set_context / set_value | Ditto |
| test_config_get |  | (removed) |
| test_config_get_regex |  | (removed) |
| test_config_result |  | (removed) |
| cli_nexus / cli_ios_xr | nexus / ios_xr | Separate platform from data format. 'cli_nexus' was outright misleading for properties using 'structured' format |
| (new) | get_data_format | Can be 'cli' or 'nxapi_structured', with plans for 'yang' etc to be added later. Defaults to 'cli' if not specified |

Changes to client API:
- "exec" removed (it was never really used for NXAPI, and never implemented for gRPC)
- "show" renamed to "get" (less CLI-specific)
- "config" renamed to "set" (ditto)
- Filtering logic from Node class (find_ascii, find_subconfig, config_get_handle_structured) has moved into the Client namespace (new file client/utils.rb) and is invoked inside the Client.get() method instead of being applied after-the-fact by Node.config_get().

Enhancements:
- gRPC CLI config no longer cares about indentation whitespace because it concatenates 'set_context' and 'set_value' together in the XR fashion ('router bgp 100 address-family ipv4 unicast bgp router-id 1.1.1.1')
- Because we now have `get_data_format`, we no longer need to 'guess' whether a given token is for CLI or structured output. As a result, get_context and get_value (formerly config_get_token(append)) for CLI properties no longer have to be specified as regexps in the YAML - see bgp.yaml for an example.
- As a further benefit of this, get_context and set_context are often symmetric and can be defined via a single shared 'context' entry in the YAML. Again, see bgp.yaml. (I have plans to eventually extend this to get_value/set_value as well, but that's a bit trickier to implement).
- We now use Kwalify to enforce a YAML schema. Git commit hook now includes both schema validation and running of test_command_reference.rb.
- All 'context' entries are assumed to be required and will raise an error if key substitution is not possible (missing params). If you want a specific entry (e.g., 'vrf <name>' in BGP) to be optional, you must flag it as such by prepending `(?)` to the entry in the YAML.

I'm sure there's more that I've forgotten but these are the highlights. I don't intend to collapse this before we rebase feature/grpc to pick up latest code from develop, but wanted to get this out there for initial review and feedback.


docs/README-develop-best-practices.md
```diff
@@ -55,16 +55,16 @@ All attribute entries in a given YAML file must be kept in alphabetical order. A
 
 This rule is enforced by the `Cisco::CommandReference` class itself - it will raise an exception if it detects any out-of-order entries.
 
-### <a name="yaml3">Y3: Use *regexp* anchors where needed for `config_get` and `config_get_token` entries.
+### <a name="yaml3">Y3: Use *regexp* anchors where needed for CLI `get_context` and `get_value` entries.
 
-Please use *regexp* anchors `^$` to ensure you match the correct feature information in the `show` output.
+By default, CLI clients assume that `get_context` and `get_value` are to be treated as Regexps, and implicitly add regexp anchors and case-insensitivity (i.e., a `get_value` of `'router bgp 100'` becomes the regexp `/^router bgp 100$/i`). If you want to explicitly specify a regexp (perhaps because the default behavior does not meet your needs for a specific property), be sure to add the `^` and `$` anchors to ensure you match the correct feature information in the `show` output and do not unexpectedly match similar but undesired CLI strings.
```
jamtaylo, line 60: If the expected model is to use the new implicit-regex style, should some (perhaps most) of the examples in the documentation be updated to use it?

lib/cisco_node_utils/client/utils.rb
```diff
@@ -0,0 +1,155 @@
...
+  def munge_to_array(val)
+    self.class.munge_to_array(val)
+  end
+
+  # Helper function that subclasses may use with get(data_format: :cli)
+  # Method for working with hierarchical show command output such as
+  # "show running-config". Searches the given multi-line string
+  # for all matches to the given value query. If context is provided,
+  # the matches will be filtered to only those that are located "under"
+  # the given context sequence (as determined by indentation).
```
jamtaylo, line 42: Looks like `find-subconfig` could have some unexpected behavior if a context regexp does not match at the left-most column (of the currently trimmed sub-config body text).  Perhaps the documentation for the `get-context` values in the YAML files should say that matching at the beginning of a line (either by including the `^` regexp anchor or by letting the new implicit regexp behavior do it for you) is _required_, not just a best practice?

lib/cisco_node_utils/cmd_ref/README_YAML.md
```diff
@@ -154,40 +202,59 @@ instead of the more repetitive (but equally valid):
...
+    data_model: nxapi_structured
+    get_value: 'chassis_id'
+  else:
+    data_model: cli
+    get_value: '/Hardware\n  cisco (([^(\n]+|\(\d+ Slot\))+\w+)/'
+```
+
+### Platform variants
+
+Even for clients using the same data format (e.g., CLI), there may be differences between classes of Cisco platform.  Any of the attribute properties can be subdivided by platform type by using the platform type as a key. For example, interface VRF membership defaults to "" (no VRF) on both Nexus and IOS XR platforms, but the CLI is 'vrf member <vrf>' for Nexus and 'vrf <vrf>' for IOS XR. Thus, the YAML could be written as:
```
jamtaylo: The angle brackets around `<vrf>` (two occurrences) are being eaten (treated as an unknown html tag).  Are those single-quotes supposed to be back-quotes?

glennmatthews: Nice catch. Will fix.

jamtaylo:
Looks good to me (for whatever that's worth given my limited experience with this codebase).

lib/cisco_node_utils/client/grpc/client.rb
```diff
@@ -79,27 +78,54 @@ def cache_flush
...
+    context = context.join(' ')
+    unless context.empty?
+      values.map! do |cmd|
+        match = cmd[/^\s*no\s+(.*)/, 1]
+        if match
+          cmd = "no #{context} #{match}"
+        else
+          cmd = "#{context} #{cmd}"
+        end
+        cmd
```
RichWellum, line 112: This is great - much better!

:+1:

lib/cisco_node_utils/client/nxapi/client.rb
```diff
@@ -103,59 +102,56 @@ def cache_flush
...
-  # @param command [String] the exec command to execute
-  # @return [String, nil] the body of the output of the exec command
-  #   (if any)
-  def exec(command)
+  # @param data_format one of Cisco::DATA_FORMATS. Default is :cli
+  # @param context [String, Array<String>] Zero or more configuration commands
+  #   used to enter the desired CLI sub-mode
+  # @param values [String, Array<String>] One or more commands
+  #   to enter within the CLI sub-mode.
+  def set(data_format: :cli,
```
RichWellum, line 114: I still feel a prefix makes this a little clearer (cli_set and cli_get etc ) as a generic set / get isn't that obvious. But I know you removed 'config_' deliberately so not a big deal by any means.

lib/cisco_node_utils/cmd_ref/aaa_auth_login_service.yaml
```diff
@@ -1,22 +1,22 @@
 # aaa_auth_login_service
```
RichWellum, line 1: Haven't checked the impacted yaml files but all looks good and big improvement. :+1:

lib/cisco_node_utils/client/nxapi/client.rb
glennmatthews, line 114: I could certainly add data-model-specific functions (`cli_set(**args) == set(data_format: :cli, **args)`) but I'm not sure if there's a point. The goal is to make this as transparent as possible to the Node - it just gets stuff from the CmdRef and passes it through to the Client.

RichWellum, line 114: Fair enough.

lib/cisco_node_utils/client/utils.rb
glennmatthews, line 42: Not sure I follow - can you provide a more specific example of the case you're concerned about?

docs/README-develop-best-practices.md
glennmatthews, line 60: Great suggestion. Done.

lib/cisco_node_utils/client/utils.rb
jamtaylo, line 42: If I'm reading it correctly, `find-subconfig` selects all lines from the matched context key until the next line that doesn't start with spaces.  This works great for nested contexts because it also trims a block of leading spaces from the result.  But if the key itself didn't start at the leftmost column then the selected lines might include things at the same indentation level as the key (or even out-dented).

PR #179, opened 2016-01-22
Interface regressions fixes for 3k, 5k, 6k, 7k, and 9k

robert-w-gries:
### Fixed errors
- Remove `RuntimeError` anti-pattern where `Cisco::CliError` was rescued and `RuntimeError` was raised instead
- Add another CliError message to `speed_change_disallowed` skip pattern
- Handle uppercase letter in `encapsulation dot1Q` for 5k and 6k
- Change max valid interface description size to `80` to fix error for 5k and 6k
- Handle `switchport mode` being enabled before `mtu` tests
- Remove `all` keyword from `show_cmd` for 7k because it is unsupported for that platform
### N3k Minitest Results

```
  1) Skipped:
TestInterface#test_negotiate_auto_ethernet [/home/robgrie/cisco-network-node-utils/tests/basetest.rb:103]:
Feature 'interface', attribute 'negotiate_auto_ethernet', operation 'config_set' is unsupported on this node


  2) Skipped:
TestInterface#test_interface_duplex_change [tests/test_interface.rb:87]:
Skip test: Interface type does not allow config change


  3) Skipped:
TestInterface#test_interface_speed_valid [tests/test_interface.rb:87]:
Skip test: Interface type does not allow config change


  4) Skipped:
TestInterface#test_interface_duplex_invalid [tests/test_interface.rb:87]:
Skip test: Interface type does not allow config change


  5) Skipped:
TestInterface#test_interface_duplex_valid [tests/test_interface.rb:87]:
Skip test: Interface type does not allow config change


  6) Skipped:
TestInterface#test_interface_speed_change [tests/test_interface.rb:87]:
Skip test: Interface type does not allow config change

47 runs, 336 assertions, 0 failures, 0 errors, 6 skips
```
### N5k Minitest Results

```
  1) Skipped:
TestInterface#test_interface_speed_change [tests/test_interface.rb:87]:
Skip test: Interface type does not allow config change

47 runs, 383 assertions, 0 failures, 0 errors, 1 skips
```
### N6k Minitest Results

```
  1) Skipped:
TestInterface#test_interface_speed_change [tests/test_interface.rb:87]:
Skip test: Interface type does not allow config change

47 runs, 383 assertions, 0 failures, 0 errors, 1 skips
```
### N7k Minitest Results

```
  1) Skipped:
TestInterface#test_interface_speed_valid [tests/test_interface.rb:87]:
Skip test: Interface type does not allow config change


  2) Skipped:
TestInterface#test_negotiate_auto_portchannel [/home/robgrie/cisco-network-node-utils/tests/basetest.rb:103]:
Feature 'interface', attribute 'negotiate_auto_portchannel', operation 'config_set' is unsupported on this node


  3) Skipped:
TestInterface#test_interface_duplex_valid [tests/test_interface.rb:87]:
Skip test: Interface type does not allow config change


  4) Skipped:
TestInterface#test_negotiate_auto_ethernet [/home/robgrie/cisco-network-node-utils/tests/basetest.rb:103]:
Feature 'interface', attribute 'negotiate_auto_ethernet', operation 'config_set' is unsupported on this node


  5) Skipped:
TestInterface#test_interface_speed_change [tests/test_interface.rb:87]:
Skip test: Interface type does not allow config change


  6) Skipped:
TestInterface#test_interface_duplex_change [tests/test_interface.rb:87]:
Skip test: Interface type does not allow config change


  7) Skipped:
TestInterface#test_interface_duplex_invalid [tests/test_interface.rb:87]:
Skip test: Interface type does not allow config change

47 runs, 323 assertions, 0 failures, 0 errors, 7 skips
```
### N9k Minitest Results

```
  1) Skipped:
TestInterface#test_interface_speed_change [tests/test_interface.rb:87]:
Skip test: Interface type does not allow config change


  2) Skipped:
TestInterface#test_negotiate_auto_ethernet [tests/test_interface.rb:724]:
Skip test: Interface type does not allow config change

47 runs, 341 assertions, 0 failures, 0 errors, 2 skips
```

### DIFF
--- a/lib/cisco_node_utils/interface.rb
+++ b/lib/cisco_node_utils/interface.rb
@@ -475,11 +475,7 @@ module Cisco
     def negotiate_auto=(negotiate_auto)
       lookup = negotiate_auto_lookup_string
       no_cmd = (negotiate_auto ? '' : 'no')
-      begin
-        config_set('interface', lookup, @name, no_cmd)
-      rescue Cisco::CliError => e
-        raise "[#{@name}] '#{e.command}' : #{e.clierror}"
-      end
+      config_set('interface', lookup, @name, no_cmd)
     end
 
     def default_negotiate_auto
--- a/lib/cisco_node_utils/interface.rb
+++ b/lib/cisco_node_utils/interface.rb
@@ -74,8 +74,6 @@ module Cisco
 
     def access_vlan=(vlan)
       config_set('interface', 'access_vlan', @name, vlan)
-    rescue Cisco::CliError => e
-      raise "[#{@name}] '#{e.command}' : #{e.clierror}"
     end
 
     def ipv4_acl_in
@@ -182,8 +180,6 @@ module Cisco
       end
       config_set('interface',
                  'channel_group', @name, state, val, force)
-    rescue Cisco::CliError => e
-      raise "[#{@name}] '#{e.command}' : #{e.clierror}"
     end
 
     def default_channel_group
@@ -201,8 +197,6 @@ module Cisco
       else
         config_set('interface', 'description', @name, '', desc)
       end
-    rescue Cisco::CliError => e
-      raise "[#{@name}] '#{e.command}' : #{e.clierror}"
     end
 
     def default_description
@@ -219,8 +213,6 @@ module Cisco
       else
         config_set('interface', 'encapsulation_dot1q', @name, '', val)
       end
-    rescue Cisco::CliError => e
-      raise "[#{@name}] '#{e.command}' : #{e.clierror}"
     end
 
     def default_encapsulation_dot1q
@@ -258,8 +250,6 @@ module Cisco
         config_set('fex', 'feature', 'no') if curr == :enabled
         config_set('fex', 'feature_install', 'no')
       end
-    rescue Cisco::CliError => e
-      raise "[#{@name}] '#{e.command}' : #{e.clierror}"
     end
 
     def ipv4_addr_mask_set(addr, mask, secondary=false)
@@ -278,8 +268,6 @@ module Cisco
         am = "#{addr}/#{mask}"
       end
       config_set('interface', 'ipv4_addr_mask', @name, state, am, sec)
-    rescue Cisco::CliError => e
-      raise "[#{@name}] '#{e.command}' : #{e.clierror}"
     end
 
     def ipv4_addr_mask
@@ -371,8 +359,6 @@ module Cisco
       Pim.feature_enable unless Pim.feature_enabled
       config_set('interface', 'ipv4_pim_sparse_mode', @name,
                  state ? '' : 'no')
-    rescue Cisco::CliError => e
-      raise "[#{@name}] '#{e.command}' : #{e.clierror}"
     end
 
     def default_ipv4_pim_sparse_mode
@@ -432,8 +418,6 @@ module Cisco
     def mtu=(val)
       check_switchport_disabled
       config_set('interface', 'mtu', @name, '', val)
-    rescue Cisco::CliError => e
-      raise "[#{@name}] '#{e.command}' : #{e.clierror}"
     end
 
     def default_mtu
@@ -449,8 +433,6 @@ module Cisco
         fail 'Changing interface speed is not permitted on this platform'
       end
       config_set('interface', 'speed', @name, val)
-    rescue Cisco::CliError => e
-      raise "[#{@name}] '#{e.command}' : #{e.clierror}"
     end
 
     def default_speed
@@ -466,8 +448,6 @@ module Cisco
         fail 'Changing interface duplex is not permitted on this platform'
       end
       config_set('interface', 'duplex', @name, val)
-    rescue Cisco::CliError => e
-      raise "[#{@name}] '#{e.command}' : #{e.clierror}"
     end
 
     def default_duplex
@@ -506,8 +486,6 @@ module Cisco
     def shutdown=(state)
       no_cmd = (state ? '' : 'no')
       config_set('interface', 'shutdown', @name, no_cmd)
-    rescue Cisco::CliError => e
-      raise "[#{@name}] '#{e.command}' : #{e.clierror}"
     end
 
     def default_shutdown
@@ -633,9 +611,6 @@ module Cisco
       else
         switchport_enable_and_mode(mode_set)
       end # case
-
-    rescue Cisco::CliError => e
-      raise "[#{@name}] '#{e.command}' : #{e.clierror}"
     end
 
     def default_switchport_mode
@@ -656,8 +631,6 @@ module Cisco
         config_set(
           'interface', 'switchport_trunk_allowed_vlan', @name, '', val)
       end
-    rescue Cisco::CliError => e
-      raise "[#{@name}] '#{e.command}' : #{e.clierror}"
     end
 
     def default_switchport_trunk_allowed_vlan
@@ -676,8 +649,6 @@ module Cisco
         config_set(
           'interface', 'switchport_trunk_native_vlan', @name, '', val)
       end
-    rescue Cisco::CliError => e
-      raise "[#{@name}] '#{e.command}' : #{e.clierror}"
     end
 
     # vlan_mapping & vlan_mapping_enable
@@ -724,8 +695,6 @@ module Cisco
                      state, original, translated)
         end
       end
-    rescue Cisco::CliError => e
-      raise "[#{@name}] '#{e.command}' : #{e.clierror}"
     end
 
     # cli: switchport vlan mapping enable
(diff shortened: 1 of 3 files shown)


tests/test_interface.rb
```diff
@@ -40,7 +40,8 @@ class TestInterface < CiscoTestCase
   DEFAULT_IF_IP_PROXY_ARP = false
   DEFAULT_IF_IP_REDIRECTS = true
   DEFAULT_IF_VRF = ''
-  IF_DESCRIPTION_SIZE = 243 # SIZE = VSH Max 255 - "description " keyword
+  IF_DESCRIPTION_VALID_SIZE = 68 # SIZE = VSH Max 80 - "description " keyword
```
mikewiebe, line 43: Can you add a bit more context for why this change is needed?  I assume it's a difference across platforms?

tests/test_interface.rb
```diff
@@ -78,11 +80,12 @@ def interface_count
 
   # Helper to check for misc speed change disallowed error messages.
   def speed_change_disallowed(message)
-    pattern = Regexp.new('(port doesn t support this speed|' \
-                         'Changing interface speed is not permitted|' \
-                         'requested config change not allowed)')
+    patterns = %w(port doesn t support this speed
```
mikewiebe: Looks odd to me.  Did you intent to make each of these words a separate item in the patterns array?

Update: Never mind. I see the union below.

mikewiebe:
Minor comment, but also please close this and re-open against the `release_1.2.0` branch.  We will be merging `master` back into `develop` once the release code is in `master`.

lib/cisco_node_utils/interface.rb
```diff
@@ -74,8 +74,6 @@ def access_vlan
 
     def access_vlan=(vlan)
       config_set('interface', 'access_vlan', @name, vlan)
-    rescue Cisco::CliError => e
-      raise "[#{@name}] '#{e.command}' : #{e.clierror}"
```
chrisvanheuveln, line 78: I saw your comment in the PR description but I don't quite understand why you're removing these completely? 
These were originally added because we had a hard time determining which interface was failing when it hit an error. 
Is the interface name preserved and still displayed somewhere when we hit a cli error?

tests/test_interface.rb
chrisvanheuveln: Even with the union this can't be right -- you'll match against any single word in `patterns`. Please fix.

lib/cisco_node_utils/interface.rb
glennmatthews, line 78: IMHO, it's annoying to have this pattern copy-and-pasted all over the place, and I don't like the approach of "catch one type of error and throw a new type of error with less detail" (we're discarding `e.previous`, for one). If anything, we should prefer to enhance the CliError class itself to provide the necessary context.

tests/test_interface.rb
```diff
@@ -40,7 +40,8 @@ class TestInterface < CiscoTestCase
   DEFAULT_IF_IP_PROXY_ARP = false
   DEFAULT_IF_IP_REDIRECTS = true
   DEFAULT_IF_VRF = ''
-  IF_DESCRIPTION_SIZE = 243 # SIZE = VSH Max 255 - "description " keyword
+  IF_DESCRIPTION_VALID_SIZE = 68 # SIZE = VSH Max 80 - "description " keyword
+  IF_DESCRIPTION_INVALID_SIZE = 255
```
glennmatthews, line 44: Also it doesn't make sense to me that there's this huge gap between VALID_SIZE and INVALID_SIZE - what falls in the middle?

tests/test_interface.rb
robert-w-gries: Nice catch.  The word array literal wasn't the right choice.  Now it's changed to a normal array of strings.

I also altered one of the patterns to match an error that was being skipped by the former, improper regexp: `/does not match the (transceiver speed|port capability)/`

```
puts "union=#{Regexp.union(patterns)}"

union=(?-mix:port\ doesn\ t\ support\ this\ speed|Changing\ interface\ speed\ is\ not\ permitted|requested\ config\ change\ not\ allowed|(?-mix:does not match the (transceiver speed|port capability)))
TestInterface#test_interface_speed_change = 1.24 s = S
```

tests/test_interface.rb
robert-w-gries, line 44: @mikewiebe @glennmatthews 

The 5k and 6k platforms failed the `test_interface_description_valid` test with the following error:

```
  1) Error:
TestInterface#test_interface_description_valid:
Cisco::CliError: CliError: ' description bcdefghijklmnopqrstuvwxyz 0123456789abcdefghijklmnopqrstuvwxyz 0123456789abcdefghijklmnopqrstuvwxyz 0123456789abcdefghijklmnopqrstuvwxyz 0123456789abcdefghijklmnopqrstuvwxyz 0123456789abcdefghijklmnopqrstuvwxyz 0123456789abcdefghijklmnopqrstuv' rejected with message:
'ERROR: error: name too long. max size allowed is 80'
```

I made `VALID_SIZE` small enough that the valid test would pass on all platforms, and I made `INVALID_SIZE` large enough that it would raise an error for all platforms.  I'm not sure what you mean by "what falls in the middle?"

glennmatthews, line 44: So even though it says 'max size allowed is 80', it actually means 'max size allowed is 63'? That seems odd.

If you can just add some comments explaining why these numbers are selected, I'd be happy - it's just that with the current lack of comments, it's not at all obvious why INVALID_SIZE != VALID_SIZE + 1.

robert-w-gries, line 44: Ah I see what you mean now.  The valid test fills the description command with garbage characters up to the accepted limit.  Since the keyword `description` is included in the command, VALID_SIZE is equal to `80 - length("description ")`.  This is sort of explained in [this comment](https://github.com/cisco/cisco-network-node-utils/pull/179/files#diff-cf2422f8b2cb8aa033a2f17f39d7354fR43).

I agree it's not very obvious why we need the two different values.  Instead of using constants, I could add a `description_max` getter that explicitly performs some of this logic:

```
def description_max
  max_vsh = node.product_id =~ /N(5|6)/ ? 80 : 255
  max_vsh - 'description '.length
end
```

glennmatthews, line 44: Just to confirm - if you enter a description where the name itself (NOT counting 'description ') is 79 or 80 characters, is it actually rejected as invalid?

I like the per-platform approach you suggested in any case - ideally then you can also use `description_max + 1` instead of having a separate INVALID_SIZE constant.

robert-w-gries, line 44: After testing, it looks like n9k is the only platform that rejects the command when `VALID_SIZE` is set to `VSH_MAX` instead of `VSH_MAX - 'description '.length`.  It's a platform-specific limit that applies to all commands

```
  1) Error:
TestInterface#test_interface_description_valid:
Cisco::CliError: CliError: ' description bcdefghijklmnopqrstuvwxyz 0123456789abcdefghijklmnopqrstuvwxyz 0123456789abcdefghijklmnopqrstuvwxyz 0123456789abcdefghijklmnopqrstuvwxyz 0123456789abcdefghijklmnopqrstuvwxyz 0123456789abcdefghijklmnopqrstuvwxyz 0123456789abcdefghijklmnopqrstuvwxyz 0123456' rejected with message:
'ERROR: PPM Command exceeded vsh command limits.Optimize the command to satisfy vsh command limits'
```

robert-w-gries:
@mikewiebe I'm going to close this PR and open a new one against the release branch once the comments are closed out

lib/cisco_node_utils/interface.rb
chrisvanheuveln, line 78: In one provider I actually had created a method for the raise instead of using the pattern on every setter. 
But I agree it would be better to enhance CliError, assuming it somehow has name as part of the context.

tests/test_interface.rb
chrisvanheuveln, line 44: Just a thought: The description_too_long and _valid/_invalid tests are carry-overs from the old OneP code. I'm of the opinion that a lot of these out-of-range tests are outside the scope of our APIs -- why should my setter care that the max char limit is 80 on one platform and 255 on another? ie. attempt to set the manifest's requested state and let the platform reject it.  

IOW, my suggestion is to just remove these tests.

robert-w-gries, line 44: That sounds reasonable to me.  Any thoughts on this @glennmatthews?

glennmatthews, line 44: We already skip the too_long test on XR because there is no upper limit in XR. I'd make sure that we have a test case that covers valid description strings (i.e., with whitespace and punctuation) without exercising the length specifically.

robert-w-gries, line 44: @glennmatthews Sounds good.  I'll open a new PR with the updated test.

lib/cisco_node_utils/interface.rb
robert-w-gries, line 78: Is this a separate story, or do I need to add anything to my PR?

chrisvanheuveln, line 78: Let's not remove them until we have equivalent or better messaging from CliError. 
IMO the test effort from changing CliError is outside the scope of this PR so just open a US to tackle this down the road.

robert-w-gries, line 78: :+1:

PR #180, opened 2016-01-23
Refactor vxlan_vtep_vni NU to have ingress_replication and multicast_…

alok-aggarwal:
…addr exclusive of each other, and other minor changes.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/vxlan_vtep_vni.yaml
+++ b/lib/cisco_node_utils/cmd_ref/vxlan_vtep_vni.yaml
@@ -16,7 +16,7 @@ all_vnis:
 ingress_replication:
   kind: string
   config_get_token_append: '/^ingress-replication protocol (\S+)$/'
-  config_set_append: '<state> ingress-replication protocol <protocol>'
+  config_set_append: '<state> ingress-replication protocol <protocol> ; end'
   default_value: ''
 
 multicast_group:
--- a/lib/cisco_node_utils/vxlan_vtep_vni.rb
+++ b/lib/cisco_node_utils/vxlan_vtep_vni.rb
@@ -29,8 +29,9 @@ module Cisco
     def initialize(name, vni, assoc_vrf=false, instantiate=true)
       @name = name
       @vni = vni
-      @assoc_vrf = assoc_vrf
+      @assoc_vrf = assoc_vrf == :true ? true : false
 
+      set_args_keys_default
       create if instantiate
       # NOTE: This can be removed after the puppet type/provider are coded
       # but a few design notes so we don't forget!.
@@ -47,7 +48,7 @@ module Cisco
         vni_list = config_get('vxlan_vtep_vni', 'all_vnis', get_args)
         next if vni_list.nil?
         vni_list.each do |vni, assoc_vrf|
-          assoc_vrf = assoc_vrf.nil? ? false : true
+          assoc_vrf = assoc_vrf.nil? ? :false : :true
           hash[name][vni] = VxlanVtepVni.new(name, vni, assoc_vrf, false)
         end
       end
@@ -118,7 +119,7 @@ module Cisco
         config_set('vxlan_vtep_vni', 'ingress_replication', @set_args)
       else
         # Sadly, the only way to change between protocols is to
-        # first remove the exisitng protocol.
+        # first remove the existing protocol.
         set_args_keys(state: 'no', protocol: ingress_replication)
         config_set('vxlan_vtep_vni', 'ingress_replication', @set_args)
         set_args_keys(state: '', protocol: protocol)
@@ -132,6 +133,12 @@ module Cisco
         config_set('vxlan_vtep_vni', 'ingress_replication', @set_args) unless
           ingress_replication == default_ingress_replication
       else
+        # Since multicast group and ingress replication are exclusive
+        # properties, flush out multicast_group first
+        unless multicast_group.empty?
+          set_args_keys(state: 'no', ip_start: '', ip_end: '')
+          config_set('vxlan_vtep_vni', 'multicast_group', @set_args)
+        end
         remove_add_ingress_replication(protocol)
       end
     end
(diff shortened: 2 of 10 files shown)


lib/cisco_node_utils/cmd_ref/vxlan_vtep_vni.yaml
```diff
@@ -16,7 +16,7 @@ all_vnis:
 ingress_replication:
   kind: string
   config_get_token_append: '/^ingress-replication protocol (\S+)$/'
-  config_set_append: '<state> ingress-replication protocol <protocol>'
+  config_set_append: '<state> ingress-replication protocol <protocol> ; end'
```
mikewiebe, line 21: Why is this change needed?

alok-aggarwal, line 21: To make sure that submode of ingress replication is not entered because there is no sub mode config being applied. This change may not be correcting anything but it keeps things clean.

lib/cisco_node_utils/vxlan_vtep_vni.rb
```diff
@@ -29,8 +29,9 @@ class VxlanVtepVni < NodeUtil
     def initialize(name, vni, assoc_vrf=false, instantiate=true)
       @name = name
       @vni = vni
-      @assoc_vrf = assoc_vrf
+      @assoc_vrf = assoc_vrf == :true ? true : false
```
mikewiebe: This won't work if `assoc_vrf` is a `TrueClass` object.  While I understand that puppet is passing this as a `Symbol` other orchestration frameworks that use this may not pass it as a symbol.  I would prefer that we leave this as `@assoc_vrf = assoc_vrf` and just make sure to pass the appropriate thing in puppet.  (https://github.com/cisco/cisco-network-puppet-module/pull/142/files#diff-1323926772cf8643b976e839a2d0f0fcR167)

```
assoc_vrf = @resource[:assoc_vrf] == :true ? true : false
@vtep_vni = Cisco::VxlanVtepVni.new(@resource[:interface], @resource[:vni], assoc_vrf)
```

lib/cisco_node_utils/vxlan_vtep_vni.rb
```diff
@@ -154,11 +161,23 @@ def remove_add_multicast_group(ip_start, ip_end)
 
     def multicast_group=(range)
       if range == default_multicast_group
-        set_args_keys(state: 'no', ip_start: '', ip_end: '')
-        config_set('vxlan_vtep_vni', 'multicast_group', @set_args)
+        # Due to CSCux78514, trying to remove multicast-group in CLI
+        # when ingress replication is configured results in removing
+        # ingress replication from nvgen. So be careful and negate
+        # Multicast-group only is it is configured.
+        unless multicast_group.empty?
```
mikewiebe, line 168: Does ingress replication always get removed .. even if multicast_group is configured?

lib/cisco_node_utils/vxlan_vtep_vni.rb
```diff
@@ -154,11 +161,23 @@ def remove_add_multicast_group(ip_start, ip_end)
...
+          set_args_keys(state: 'no', ip_start: '', ip_end: '')
+          config_set('vxlan_vtep_vni', 'multicast_group', @set_args)
+        end
       else
         ip_start, ip_end = range.split(' ')
         ip_end = '' if ip_end.nil?
+        # Since multicast group and ingress replication are exclusive
+        # properties, flush out ingress replication first
```
mikewiebe: Instead of `flush out`, I would use `first remove ingress replication`

lib/cisco_node_utils/vxlan_vtep_vni.rb
```diff
@@ -195,10 +214,16 @@ def suppress_arp
...
-      set_args_keys(state: (state ? '' : 'no'))
-      config_set('vxlan_vtep_vni', 'suppress_arp', @set_args)
+      if state
+        set_args_keys(state: '')
+        # Host reachability must be enabled for this property
+        VxlanVtep.new(@name).host_reachability = 'evpn'
+        config_set('vxlan_vtep_vni', 'suppress_arp', @set_args)
+      else
+        set_args_keys(state: 'no')
+        # Negate suppress-arp only if it is configured
```
mikewiebe, line 224: Why are you adding this check here?  Does this also result in bad things happening if you attempt to remove it when it's not configured?

tests/test_vxlan_vtep_vni.rb
```diff
@@ -34,7 +34,7 @@ def teardown
 
   def test_create_with_existing
     VxlanVtep.new('nve1').host_reachability = 'evpn'
-    associate_vrf = true
+    associate_vrf = :true
```
mikewiebe: Change back to `true`.

tests/test_vxlan_vtep_vni.rb
```diff
@@ -55,13 +55,14 @@ def test_vnis
     # Host reachablity must be set to evpn for associate_vrf
     # testing.
     VxlanVtep.new('nve1').host_reachability = 'evpn'
-    associate_vrf = true
+    associate_vrf = :true
```
mikewiebe: Change back to `true`.

mikewiebe:
@alok-aggarwal Please re-open this against the `release_1.2.0` branch.

tests/test_vxlan_vtep_vni.rb
alok-aggarwal: Done.

tests/test_vxlan_vtep_vni.rb
alok-aggarwal: Done.

lib/cisco_node_utils/vxlan_vtep_vni.rb
alok-aggarwal, line 224: This is to protect customers who haven't reloaded the box after configuring the TCAM regions. negation will send a 'no suppress arp' which causes an error. The default it 'false' which means 'no suppress-arp' which means error :)

lib/cisco_node_utils/vxlan_vtep_vni.rb
alok-aggarwal, line 168: Ingress_rep can't be configured unless we remove mcast_group. But if we just say no mcast-group even if its not configured, it removes ingress-rep as well which is a defect.

lib/cisco_node_utils/vxlan_vtep_vni.rb
alok-aggarwal: done.

lib/cisco_node_utils/vxlan_vtep_vni.rb
alok-aggarwal: Done.

PR #181, opened 2016-01-23
Initial commit interface_channel_group provider

chrisvanheuveln:
- moved channel_group property out of interface provider, into new interface_channel_group provider
  - channel-group causes problems in nxos when it is present on an interface; e.g. you cannot change switchport modes, etc
  - moving it into its own provider allows the manifest to explicitly remove it while allowing new (non-channel-group) properties to be added to the interface

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,12 +16,15 @@ Changelog
 * Feature
   * feature (@robert-w-gries)
 * Interface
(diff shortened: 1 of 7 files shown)


mikewiebe:
Looks good. :+1:

mikewiebe:
Sorry.. clicked the close button by accident :)

PR #182, opened 2016-01-25
fix node util for n7k

andy91315:
1 do not need feature overlay
2. n7k do not have default value for properties

### DIFF
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -360,9 +360,7 @@ module Cisco
     # Nvgen as True With optional 'size <size>
     def event_history_cli
       match = config_get('bgp', 'event_history_cli', @get_args)
-      if match == default_event_history_cli
-        return default_event_history_cli
-      end   
+      return default_event_history_cli if match == 'size_small'
       return 'false' if match[0] == 'no '
       return 'size_' + match[1] if match[1]
       default_event_history_cli
@@ -405,9 +403,7 @@ module Cisco
     # Nvgen as True With optional 'size <size>
     def event_history_events
       match = config_get('bgp', 'event_history_events', @get_args)
-      if match == default_event_history_events
-        return default_event_history_events
-      end
+      return default_event_history_events if match == 'size_small'
       return 'false' if match[0] == 'no '
       return 'size_' + match[1] if match[1]
       default_event_history_events
@@ -429,9 +425,7 @@ module Cisco
     # Nvgen as True With optional 'size <size>
     def event_history_periodic
       match = config_get('bgp', 'event_history_periodic', @get_args)
-      if match == default_event_history_periodic
-        return default_event_history_periodic
-      end
+      return default_event_history_periodic if match == 'size_small'
       return 'false' if match[0] == 'no '
       return 'size_' + match[1] if match[1]
       default_event_history_periodic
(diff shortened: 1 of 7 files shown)


chrisvanheuveln:
1. Please close this PR and re-open for merge with `release_1.2.0`.
2. Fix rubocop errors

lib/cisco_node_utils/bgp.rb
```diff
@@ -360,6 +360,9 @@ def default_enforce_first_as
     # Nvgen as True With optional 'size <size>
     def event_history_cli
       match = config_get('bgp', 'event_history_cli', @get_args)
+      if match == default_event_history_cli
```
chrisvanheuveln: This method already returns the default value at the end. Can you explain why this is needed here?

edit: I see this issue. N7k doesn't show defaults; so config_get gets nil, returns the default value which is a string, so `match[1]` is `i` (first index in the string `size_small`).

edit 2: Playing around with this it looks like you don't need the second `default_event_history_cli` call on line 368; nor the one on line 437; however, you still need the one on line 413 because of the wacky behavior on the 7k.

tests/test_bgp_neighbor.rb
```diff
@@ -215,6 +215,7 @@ def test_bgpneighbor_set_get_log_neighbor_changes
       neighbor = RouterBgpNeighbor.new(@@asn, vrf, @@addr)
       check = [:enable, :disable, :inherit, 'enable', 'disable', 'inherit',
                neighbor.default_log_neighbor_changes]
+      end
```
chrisvanheuveln: Why is this `end` here? I can't imagine that the test file still works after this?

mikewiebe:
@andy91315 This needs to be opened against the release branch.

tests/test_router_bgp.rb
```diff
@@ -777,7 +779,7 @@ def test_routerbgp_set_get_reconnect_interval
   def test_routerbgp_get_reconnect_interval_default
     asnum = 55
     bgp = RouterBgp.new(asnum)
-    assert_equal(60, bgp.reconnect_interval,
+    assert_equal(bgp.default_reconnect_interval, bgp.reconnect_interval,
                  "reconnect_interval should be set to default value of '60'")
```
chrisvanheuveln, line 783:
```
-                  "reconnect_interval should be set to default value of '60'")
+                  "reconnect_interval should be set to default value")
```

lib/cisco_node_utils/bgp.rb
```diff
@@ -660,7 +663,6 @@ def route_distinguisher
     end
 
     def route_distinguisher=(rd)
-      Feature.nv_overlay_enable
```
mikewiebe, line 663: Is removing `Feature.nv_overlay_enable` ok for non-n7k platforms?

lib/cisco_node_utils/bgp.rb
```diff
@@ -685,10 +687,15 @@ def router_id
     def router_id=(id)
       # In order to remove a bgp router-id you cannot simply issue
```
mikewiebe: Remove the comment.

lib/cisco_node_utils/bgp.rb
```diff
@@ -685,10 +687,15 @@ def router_id
...
       # 'no bgp router-id'. Dummy-id specified to work around this.
-      dummy_id = '1.2.3.4'
+      # not working on N7K need a valid router_id
```
mikewiebe, line 690: Suggest the following instead.

```
    def router_id=(id)
      if id == default_router_id
        # Nothing to do if router_id is already set to default.
        return if router_id == default_router_id
        @set_args[:state] = 'no'
        @set_args[:id] = router_id
      else
        @set_args[:state] = ''
         @set_args[:id] = id
      end
      config_set('bgp', 'router_id', @set_args)
      set_args_keys_default
    end
```

PR #183, opened 2016-01-25
Added IPv6 support for radius_server

jonnytdevops:
Folks, it would be appreciated if you could review this for IPv6 support, thanks

### DIFF
--- a/lib/cisco_node_utils/radius_server.rb
+++ b/lib/cisco_node_utils/radius_server.rb
@@ -17,6 +17,7 @@
 # limitations under the License.
 
 require File.join(File.dirname(__FILE__), 'node_util')
+require 'ipaddr'
 
 module Cisco
   # RadiusServer - node utility class for
@@ -25,8 +26,16 @@ module Cisco
     attr_reader :name
 
     def initialize(name, instantiate=true)
-      unless name[/^(?:[0-9]{1,3}\.){3}[0-9]{1,3}$/]
-        fail ArgumentError, 'Invalid value (Name is not an IP address)'
+      unless name =~ /^[a-zA-Z0-9\.\:]*$/
+        fail ArgumentError,
+             'Invalid value (IPv4/IPv6 address contains invalid characters)'
+      end
+
+      begin
+        IPAddr.new(name)
+      rescue
+        raise ArgumentError,
+              'Invalid value (Name is not a valid single IPv4/IPv6 address)'
       end
       @name = name
 
@@ -37,8 +46,7 @@ module Cisco
       hash = {}
 
       radiusservers_list = config_get('radius_server', 'hosts')
-      return hash if radiusservers_list.nil?
-
+      return hash if radiusservers_list.empty?
       radiusservers_list.each do |id|
         hash[id] = RadiusServer.new(id, false)
       end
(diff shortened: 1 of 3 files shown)


jonnytdevops:
TestRadiusServer#test_create_destroy_single_ipv6 = 14.73 s = .
TestRadiusServer#test_create_destroy_single = 14.00 s = .
TestRadiusServer#test_radiusserver_create_destroy_multiple = 16.23 s = .

Finished in 44.970864s, 0.0667 runs/s, 1.8901 assertions/s.

3 runs, 85 assertions, 0 failures, 0 errors, 0 skips

jonnytdevops:
Note: the reason why we have 2 validation steps for the IP address, is because we don't want to allow subnets, which unfortunately ruby's ipaddr library accepts and returns true for the ipv4? and ipv6? methods.

lib/cisco_node_utils/radius_server.rb
```diff
@@ -25,8 +26,16 @@ class RadiusServer < NodeUtil
     attr_reader :name
 
     def initialize(name, instantiate=true)
-      unless name[/^(?:[0-9]{1,3}\.){3}[0-9]{1,3}$/]
-        fail ArgumentError, 'Invalid value (Name is not an IP address)'
+      unless name =~ /^[a-zA-Z0-9\.\:]*$/
+        fail ArgumentError,
+             'Invalid value (IP address contains invalid characters)'
```
chrisvanheuveln: `~IP~IPv4/IPv6~`

Line 38 as well.

tests/test_radius_server.rb
```diff
@@ -90,9 +91,60 @@ def test_radiusserver_create_destroy_single
     refute_includes(Cisco::RadiusServer.radiusservers, id)
   end
 
+  def test_create_destroy_single_ipv6
+    id = '2004::3'
+    refute_includes(Cisco::RadiusServer.radiusservers, id)
+
+    server = Cisco::RadiusServer.new(id, true)
+    assert_includes(Cisco::RadiusServer.radiusservers, id)
+    assert_equal(Cisco::RadiusServer.radiusservers[id], server)
```
chrisvanheuveln: You've got the params reversed on these `assert_equal` calls - should be `assert_equal(expected, actual)`

When they fail they'll display the wrong values for `expected:` and `actual:`

jonnytdevops:
@chrisvanheuveln Code updated. Can this be merged now please? Thanks

jonnytdevops:
TestRadiusServer#test_create_destroy_single = 14.58 s = .
TestRadiusServer#test_create_destroy_single_ipv6 = 13.83 s = .
TestRadiusServer#test_radiusserver_create_destroy_multiple = 15.82 s = .

Finished in 44.226840s, 0.0678 runs/s, 1.9219 assertions/s.

3 runs, 85 assertions, 0 failures, 0 errors, 0 skips

jonnytdevops:
@chrisvanheuveln just realised that I've swapped everything when it didn't need to be swapped. Just re-looking over this, thanks.

jonnytdevops:
@chrisvanheuveln actually this looks good to me now! Please merge when ready, unless you see any other issues :)

BTW, assert_includes was always correct. The function fails the other way round :)

Chris: yep, I was only concerned about the assert_equal checks :-)

PR #186, opened 2016-01-25
Adding support for self.pims needed for beaker

smigopal:
# Description:

Adding support for self.pims needed for beaker
# Rubocop:

rtpfe1@agent-lab11-ws:~/trial_sync/cisco-network-node-utils$ rubocop lib/cisco_node_utils/pim.rb
warning: parser/current is loading parser/ruby22, which recognizes
warning: 2.2.3-compliant syntax, but you are running 2.2.1.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
Inspecting 1 file
.

1 file inspected, no offenses detected
rtpfe1@agent-lab11-ws:~/trial_sync/cisco-network-node-utils$ rubocop tests/test_pim.rb
warning: parser/current is loading parser/ruby22, which recognizes
warning: 2.2.3-compliant syntax, but you are running 2.2.1.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
Inspecting 1 file
.

1 file inspected, no offenses detected
rtpfe1@agent-lab11-ws:~/trial_sync/cisco-network-node-utils$
# Minitests:

rtpfe1@agent-lab11-ws:~/trial_sync/cisco-network-node-utils$ ruby tests/test_pim.rb -v -- 10.122.197.125 admin admin
Run options: -v -- --seed 23965

 Running:

Node under test:
- name  - agent-lab11-nx
- type  - N9K-NXOSV

TestPim#test_single_invalid_ssm_range_single_vrf = 2.46 s = .
TestPim#test_multiple_ssm_range_overwrite_multiple_vrfs = 2.31 s = .
TestPim#test_single_ssm_range_none_single_vrf = 1.73 s = .
TestPim#test_multiple_ssm_range_multiple_vrfs = 1.91 s = .
TestPim#test_single_ssm_range_single_vrf = 1.87 s = .

Finished in 10.282460s, 0.4863 runs/s, 0.8753 assertions/s.

5 runs, 9 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for MiniTest to /home/rtpfe1/trial_sync/cisco-network-node-utils/coverage. 388 / 520 LOC (74.62%) covered.
rtpfe1@agent-lab11-ws:~/trial_sync/cisco-network-node-utils$

### DIFF
--- a/lib/cisco_node_utils/pim.rb
+++ b/lib/cisco_node_utils/pim.rb
@@ -47,35 +47,26 @@ module Cisco
       config_set('pim', 'feature')
     end
 
+    # self.pims returns a hash of all current pim objects.
+    # There will be one object for each vrf. This method has slightly different
+    # behavior than most of the other "get all objects" methods because the
+    # pim properties are standalone and do not reside in a "pim" container;
+    # therefore, this method simply returns a pim object for each vrf
+    # regardless of whether there are any specific pim properties currently
+    # defined for the context.
     def self.pims
-      afis = %w(ipv4) # Add ipv6 later
+      afis = %w(ipv4) # TBD: No support for ipv6 at this time
       hash_final = {}
       afis.each do |afi|
         hash_final[afi] = {}
         default_vrf = 'default'
-        ranges = config_get('pim', 'all_ssm_ranges', afi: Pim.afi_cli(afi))
-        unless ranges.nil?
-          ranges.each do |range|
-            # Get the RPs under default VRF
-            hash_final[afi][default_vrf] =
-                      Pim.new(afi, default_vrf, false)
-            hash_final[afi][default_vrf].ssm_range = (range)
-          end
-        end
-        # Getting all custom vrfs rp_Addrs"
+        hash_final[afi][default_vrf] = Pim.new(afi, default_vrf, false)
+
+        # Now the vrf's
         vrf_ids = config_get('vrf', 'all_vrfs')
         vrf_ids.delete_if { |vrf_id| vrf_id == 'management' }
         vrf_ids.each do |vrf|
-          get_args = { vrf: @vrf, afi: Pim.afi_cli(afi) }
-          get_args[:vrf] = vrf
-          ranges = config_get('pim', 'all_ssm_ranges', get_args)
-          next if ranges.nil?
-          ranges.each do |addr|
-            hash_final[afi] ||= {}
-            hash_final[afi][vrf] =
-                      Pim.new(afi, vrf, false)
-            hash_final[afi][vrf].ssm_range = (addr)
-          end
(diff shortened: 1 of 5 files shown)


lib/cisco_node_utils/pim.rb
```diff
@@ -47,6 +47,44 @@ def self.feature_enable
...
+          end
+        end
+      end
+      hash_final
+    rescue Cisco::CliError => e
+      # cmd will syntax reject when feature is not enabled
+      raise unless e.clierror =~ /Syntax error/
+      return {}
+    end
+
```
chrisvanheuveln, line 78: Sorry for the delay on this review. I noticed earlier that this change was a problem but I didn't get the chance to come up with a codefix for you until now.

The `self.pims` method above can't be used as is - the problem is that you've tied it too strictly to 'ssm-range', which means that if I have a different property like 'anycast-rp' then I won't get an object unless I've also got an 'ssm-range' configured.

But that's okay, you were on the right track with this. In this case all you need to do is create an object for each vrf and you're done. I drafted up some test code that works -- it returns what I expect to see and passes your minitest anyway:

```
    # self.pims returns a hash of all current pim objects.
    # There will be one object for each vrf. This method has slightly different
    # behavior than most of the other "get all objects" methods because the
    # pim properties are standalone and do not reside in a "pim" container;
    # therefore, this method simply returns a pim object for each vrf
    # regardless of whether there are any specific pim properties currently
    # defined for the context.
    def self.pims
      afis = %w(ipv4) # TBD: No support for ipv6 at this time
      hash_final = {}
      afis.each do |afi|
        hash_final[afi] = {}
        vrf = 'default'
        hash_final[afi][vrf] = Pim.new(afi, vrf, false)

        # Now the vrf's
        vrf_ids = config_get('vrf', 'all_vrfs')
        vrf_ids.delete_if { |vrf_id| vrf_id == 'management' }
        vrf_ids.each do |vrf|
          hash_final[afi][vrf] = Pim.new(afi, vrf, false)
        end
      end
      hash_final
    rescue Cisco::CliError => e
      # cmd will syntax reject when feature is not enabled
      raise unless e.clierror =~ /Syntax error/
      return {}
    end
```

Also, you can remove the `all_ssm_ranges` entry in the yaml:

```
-all_ssm_ranges:
-  multiple:
-  config_get_token_append : '/^<afi> pim ssm range (.*)$/'
-
```

The rest of your diff looks fine.

smigopal, line 78: Hey Chris

Initially I had coded it without the ssm_range itself.
But since I had an unused variable addr, I added it into the hash.
But, you're right it makes sense to uncouple that property from the hash.
I think I have the old version of the code which worked, will use that and update the comment.

Thanks
Smitha

PR #189, opened 2016-01-25
Change YAML product regexps to standardized keys

glennmatthews:
(Opening this as a PR against develop because it's more applicable to Nexus than to XR at present, but perhaps too risky for the 1.2 release.)

Basically this gets rid of the special-case handling in our YAML for product-id regexps (`/N7/`, `/N(5|6|7)/`, etc.) in favor of having standardized YAML keys that represent known platform categories. This makes it that much easier to define a proper YAML schema (as was recently committed to feature/grpc), provides a centralized location for determining what defines a product category, and lets us make our YAML parser just a little bit less 'clever'.

```
$ ruby tests/test_command_reference.rb 
Run options: --seed 55599

# Running:

........................

Finished in 0.959710s, 25.0075 runs/s, 85.4425 assertions/s.

24 runs, 82 assertions, 0 failures, 0 errors, 0 skips
```


lib/cisco_node_utils/cmd_ref/feature.yaml
```diff
@@ -35,9 +35,10 @@ nv_overlay_evpn:
...
 vn_segment_vlan_based:
   # MT-lite only
-  /N(3|9)/:
+  N3k: &vn_segment_vlan_based_N3k
     kind: boolean
     config_get: 'show running section feature'
     config_get_token: '/^feature vn-segment-vlan-based$/'
     config_set: 'feature vn-segment-vlan-based'
     default_value: false
+  N9k: *vn_segment_vlan_based_N3k
```
chrisvanheuveln: I'm not familiar with this new syntax:

```
  N3k: &vn_segment_vlan_based_N3k
```

```
  N9k: *vn_segment_vlan_based_N3k
```

Can you provide some context or a readme link? Just trying to understand how these are used.

glennmatthews: Yeah, sorry, I had meant to do so. This is a YAML feature (http://www.yaml.org/spec/1.2/spec.html#id2785586) - basically think of it as aliases or pointers. `&label` provides an anchor to link against, while `*label` means 'this is a reference to that label'. So by doing it this way I'm basically saying 'the N9k values are identical to the N3k values' without needing to repeat myself and without needing to introduce special syntax for saying 'these are the values for N3k + N9k'. Clear as mud? :)

chrisvanheuveln: Got it. I was concerned that these would somehow drive changes to the NU APIs themselves but these anchors & aliases are only meaningful inside the yaml file itself.

Consider adding a note or reference in README_YAML.md since this is bound to come up again (that's the first place I looked).

:+1: on the diffset

glennmatthews: Great suggestion. Updated both README_YAML and develop-best-practices with this information. Mind taking a look?

docs/README-develop-best-practices.md
```diff
@@ -206,6 +207,42 @@ The following commands should be preferred over `show [feature]` commands since
...
+```yaml
+  vn_segment_vlan_based:
+   # MT-lite only
+   N3k: &vn_segment_vlan_based_N3k
+     kind: boolean
+     config_get: 'show running section feature'
+     config_get_token: '/^feature vn-segment-vlan-based$/'
+     config_set: 'feature vn-segment-vlan-based'
+     default_value: false
+   N9k: *vn_segment_vlan_based_N3k
```
chrisvanheuveln: I like the readme writeups. 

Just one last suggestion on the anchor ref names themselves: it might be better (at least in the MT cases) to use an abstract tag instead of referencing the plat type, e.g:

```
N3k: &mt_lite_vn_segment_vlan_based
 ...
N9k: *mt_lite_vn_segment_vlan_based

```

```
N3k: &mt_lite_mapped_vlan
 ...
N9k: *mt_lite_mapped_vlan
```

```
N3k: &mt_lite_destroy
 ...
N9k: *mt_lite_destroy
```

(or `&destroy_mt_lite`, etc)

mikewiebe: +1 on this suggestion, but maybe add `common` to the anchor name?

```
N5k: 
 ...
N3k: &mt_lite_destroy_common
 ...
N9k: *mt_lite_destroy_common
...
N7k: *mt_lite_destroy_common
```

chrisvanheuveln: or even:

`N9k: *mt_lite_destroy_commonly_unique_to_nexus_multi_tenancy_lite_supported_platforms_mostly`

:-)

lib/cisco_node_utils/command_reference.rb
```diff
@@ -339,12 +339,34 @@ def debug(text)
...
+        /N3/
+      when 'N5k'
+        /N5/
+      when 'N6k'
+        /N6/
+      when 'N7k'
+        /N7/
+      when 'N9k'
+        /N9/
+      when 'XRv9k'
```
mikewiebe, line 358: Add NX-OSv?  I return `n9k` for both NX-OSv and the hardware platform in the beaker utilitylib api.

https://github.com/cisco/cisco-network-puppet-module/blob/develop/tests/beaker_tests/lib/utilitylib.rb#L568

docs/README-develop-best-practices.md
mikewiebe: **YES**

lib/cisco_node_utils/command_reference.rb
glennmatthews, line 358: This is just mapping the key in the YAML to the corresponding regexp. I didn't have any examples of NX-OSv specific keys in the YAML so I didn't add it yet. We can add later if needed.

docs/README-develop-best-practices.md
glennmatthews: I renamed them to one of the more sane suggestions. :-)

glennmatthews:
I went ahead and ported over the commit from feature/grpc that defines a schema and uses Kwalify to validate the YAML against this schema as part of the git pre-commit hooks and part of the travis validation. The schema isn't exactly the same as in that branch (since develop doesn't have my latest YAML refactorings) but this version of the schema is correct for the code currently in develop.

mikewiebe:
:+1:

PR #193, opened 2016-01-26
Change NU dependency to nxapi 1.0.1

chrisvanheuveln:
I ran into a null hash problem with one of the new providers; the RC was traced back to an nxapi problem that was fixed with 1.0.1: https://github.com/cisco/cisco-nxapi/pull/2/files


mikewiebe:
:+1:

PR #194, opened 2016-01-27
US63892: added a XR specific attribute (bgp nsr) in Ruby API object.

sushirol:


sushirol:
$ ruby test_router_bgp.rb -v -n test_nsr -- agent-lab6-nx admin admin 
Run options: -v -n test_nsr -- --seed 33036 

# Running:

Node under test:                                                                                                         
- name  - agent-lab6-nx  
- type  - N9K-NXOSV  
  TestRouterBgp#test_nsr = 11.30 s = .  

Finished in 11.301495s, 0.0885 runs/s, 0.1770 assertions/s.                                                              

1 runs, 2 assertions, 0 failures, 0 errors, 0 skips

Coverage report generated for MiniTest to /home/rtpfe1/cisco-network-node-utils/tests/coverage. 0.0 / 0.0 LOC (100.0%) covered.

$ ruby test_router_bgp.rb -v -n test_nsr -- agent-lab22-xr:57777 root lab
Run options: -v -n test_nsr -- --seed 56429

# Running:

Node under test:
- name  - agent-lab22-xr
- type  - R-IOSXRV9000-CH
  TestRouterBgp#test_nsr = 8.92 s = .

Finished in 8.926540s, 0.1120 runs/s, 0.3361 assertions/s.

1 runs, 3 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for MiniTest to /home/rtpfe1/cisco-network-node-utils/tests/coverage. 0.0 / 0.0 LOC (100.0%) covered.

lib/cisco_node_utils/bgp.rb
```diff
@@ -152,6 +158,22 @@ def bestpath_med_non_deterministic
       config_get('bgp', 'bestpath_med_non_deterministic', @get_args)
     end
 
+    # nsr Setter
+    def nsr=(enable)
+      if platform == :nexus
+        fail Cisco::UnsupportedError.new('bgp', 'nsr', 'set', 'nsr is not ' \
+                  'configurable on NEXUS')
```
glennmatthews: I'd prefer 'Nexus' or 'NX-OS'. Better still would be to remove this check and let the default handling of an excluded property take place instead - compare against the implementation of `graceful_restart=`

lib/cisco_node_utils/bgp.rb
```diff
@@ -152,6 +158,22 @@ def bestpath_med_non_deterministic
...
+      end
+      if @vrf != 'default'
+        fail Cisco::UnsupportedError.new('bgp', 'nsr', 'set', 'nsr is not ' \
+                  'configurable with verf != default')
```
glennmatthews: 'vrf' not 'verf' - but better still to copy the failure message style used elsewhere in this class (see `graceful_restart=` as an example).

tests/test_router_bgp.rb
```diff
@@ -229,6 +229,27 @@ def test_get_asnum
...
+    nsr(setup_default)
+    nsr(setup_vrf)
+  end
+
+  def nsr(bgp)
+    if platform == :nexus || !@vrf[/default/]
+      assert_raises(Cisco::UnsupportedError) do
+        bgp.nsr = true
+      end
+    else
```
glennmatthews, line 254: Add a check to make sure the default is correct before you set anything directly:

``` ruby
else
  assert_equal(bgp.nsr, bgp.default_nsr)
  bgp.nsr = true
  ...
```

tests/test_router_bgp.rb
```diff
@@ -229,6 +229,27 @@ def test_get_asnum
...
   end
 
+  def test_nsr
+    nsr(setup_default)
+    nsr(setup_vrf)
+  end
+
+  def nsr(bgp)
+    if platform == :nexus || !@vrf[/default/]
+      assert_raises(Cisco::UnsupportedError) do
```
glennmatthews, line 251: For unsupported cases, you should also check `bgp.nsr == false` and `bgp.default_nsr == false`.

glennmatthews, line 251: Not all inside the assert_raises block though :-)

```
  bgp.nsr = true
end
assert_equal(bgp.nsr, bgp.default_nsr)
assert_nil(bgp.default_nsr)
```

lib/cisco_node_utils/bgp.rb
```diff
@@ -152,6 +158,22 @@ def bestpath_med_non_deterministic
       config_get('bgp', 'bestpath_med_non_deterministic', @get_args)
     end
 
+    # nsr Setter
```
glennmatthews, line 161: This could still stand to be refactored to be more like `graceful_restart=`

lib/cisco_node_utils/bgp.rb
```diff
@@ -122,6 +122,12 @@ def set_args_keys_default
 
     # Attributes:
 
+    # nsr Getter
+    def nsr
+      result = config_get('bgp', 'nsr', @get_args)
```
glennmatthews: In order to keep Puppet happy, you will need to return 'nil' for cases where it's unsupported. You get this for free with Nexus thanks to the `_exclude` in the YAML, but for XR VRF you need an explicit check:

``` ruby
def nsr
  return nil if platform == :ios_xr && @vrf != 'default'
  config_get('bgp', 'nsr', @get_args)
end
```

tests/test_router_bgp.rb
```diff
@@ -229,6 +229,35 @@ def test_get_asnum
...
   end
 
+  def test_nsr
+    nsr(setup_default)
+    nsr(setup_vrf)
+  end
+
+  def nsr(bgp)
+    if platform == :nexus || !@vrf[/default/]
+      refute(bgp.default_nsr,
```
glennmatthews: Better to `assert_nil` instead of `refute` as `refute` will accept both `nil` (desired) and `false` (wrong).

tests/test_router_bgp.rb
```diff
@@ -229,6 +229,35 @@ def test_get_asnum
...
+             'bgp nsr should be disabled by default')
+      assert_raises(Cisco::UnsupportedError) do
+        bgp.nsr = true
+        assert(bgp.nsr,
```
glennmatthews: You don't need the rest of these lines in the `assert_raises` block - `bgp.nsr = true` should raise the desired error immediately.

tests/test_router_bgp.rb
chburket: Yes, I think you want to assert_nil for bgp.nsr and bgp.default_nsr outside of the assert_raises block.

glennmatthews:
:+1:

lib/cisco_node_utils/bgp.rb
```diff
@@ -152,6 +158,17 @@ def bestpath_med_non_deterministic
       config_get('bgp', 'bestpath_med_non_deterministic', @get_args)
     end
 
+    # nsr Setter
+    def nsr=(enable)
+      if platform == :nexus || @vrf != 'default'
+        fail Cisco::UnsupportedError.new('bgp', 'nsr', 'set', 'nsr is not ' \
```
chburket, line 164: Could you make the error text a little more descriptive?  Something like this:
"nsr is not configurable ' \
'on a per-VRF basis on Nexus')

(I used other msgs in bgp.rb as an example)

chburket:
Other than my last comment, looks good!

lib/cisco_node_utils/bgp.rb
glennmatthews, line 164: To do this (which I'm in favor of, btw) you'll have to remove the nexus check (which is fine, as you've excluded Nexus in the YAML so the config_set() below will automatically raise an UnsupportedError.

```
if platform == :ios_xr && @vrf != 'default'
  fail Cisco::UnsupportedError.new(...
```

sushirol, line 164: It makes sense to remove the nexus check. Also it will help to keep code consistent with the rest.

chburket, line 164: Oh, good point.  Then the message should be simply:

"nsr is not configurable ' \
'on a per-VRF basis"

Or will that be annoying to the end-user?  Imagine on nexus, they first try to configure nsr on a non-default vrf and get the "nsr is not configurable on a per-VRF basis" message.  They say "oh, I'll just set it for global BGP" and then get the normal unsupported error.  "Why didn't you just tell me that to begin with?"

This is really nitpicking, so it's fine whichever way you want to do it.

PR #195, opened 2016-01-27
fix for test on n7k

andy91315:
1. unset router-id only when it is set
2. n7k does not have default values
3. feature nv overlay does not needed

### DIFF
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -689,14 +689,13 @@ module Cisco
 
     def router_id=(id)
       # In order to remove a bgp router-id you cannot simply issue
-      # 'no bgp router-id'. Dummy-id specified to work around this.
-      # not working on N7K need a valid router_id
-      dummy_id = router_id
+      # 'no bgp router-id'. On some platforms you can specify a dummy
+      # value, but on N7K at least you need the current router_id.
       if id == default_router_id
         # Nothing to do if router_id is already set to default.
         return if router_id == default_router_id
         @set_args[:state] = 'no'
-        @set_args[:id] = dummy_id
+        @set_args[:id] = router_id
       else
         @set_args[:state] = ''
         @set_args[:id] = id
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -360,9 +360,10 @@ module Cisco
     # Nvgen as True With optional 'size <size>
     def event_history_cli
       match = config_get('bgp', 'event_history_cli', @get_args)
-      return default_event_history_cli if match == 'size_small'
-      return 'false' if match[0] == 'no '
-      return 'size_' + match[1] if match[1]
+      if match.is_a?(Array)
+        return 'false' if match[0] == 'no '
+        return 'size_' + match[1] if match[1]
+      end
       default_event_history_cli
     end
 
@@ -403,9 +404,10 @@ module Cisco
     # Nvgen as True With optional 'size <size>
     def event_history_events
       match = config_get('bgp', 'event_history_events', @get_args)
-      return default_event_history_events if match == 'size_small'
-      return 'false' if match[0] == 'no '
-      return 'size_' + match[1] if match[1]
+      if match.is_a?(Array)
+        return 'false' if match[0] == 'no '
+        return 'size_' + match[1] if match[1]
+      end
       default_event_history_events
     end
 
@@ -425,9 +427,10 @@ module Cisco
     # Nvgen as True With optional 'size <size>
     def event_history_periodic
       match = config_get('bgp', 'event_history_periodic', @get_args)
-      return default_event_history_periodic if match == 'size_small'
-      return 'false' if match[0] == 'no '
-      return 'size_' + match[1] if match[1]
+      if match.is_a?(Array)
+        return 'false' if match[0] == 'no '
+        return 'size_' + match[1] if match[1]
+      end
       default_event_history_periodic
     end
 
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -690,12 +690,10 @@ module Cisco
       # not working on N7K need a valid router_id
       dummy_id = router_id
       if id == default_router_id
-        if dummy_id.empty?
-          return
-        else
-          @set_args[:state] = 'no'
-          @set_args[:id] = dummy_id
-        end
+        # Nothing to do if router_id is already set to default.
+        return if router_id == default_router_id
+        @set_args[:state] = 'no'
+        @set_args[:id] = dummy_id
       else
         @set_args[:state] = ''
         @set_args[:id] = id
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -360,9 +360,7 @@ module Cisco
     # Nvgen as True With optional 'size <size>
     def event_history_cli
       match = config_get('bgp', 'event_history_cli', @get_args)
-      if match == default_event_history_cli
-        return default_event_history_cli
-      end   
+      return default_event_history_cli if match == 'size_small'
       return 'false' if match[0] == 'no '
       return 'size_' + match[1] if match[1]
       default_event_history_cli
@@ -405,9 +403,7 @@ module Cisco
     # Nvgen as True With optional 'size <size>
     def event_history_events
       match = config_get('bgp', 'event_history_events', @get_args)
-      if match == default_event_history_events
-        return default_event_history_events
-      end
+      return default_event_history_events if match == 'size_small'
       return 'false' if match[0] == 'no '
       return 'size_' + match[1] if match[1]
       default_event_history_events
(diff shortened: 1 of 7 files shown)


chrisvanheuveln:
If this PR is replacing #182 then please close the earlier PR.

lib/cisco_node_utils/bgp.rb
```diff
@@ -360,6 +360,7 @@ def default_enforce_first_as
     # Nvgen as True With optional 'size <size>
     def event_history_cli
       match = config_get('bgp', 'event_history_cli', @get_args)
+      return default_event_history_cli if match == 'size_small'
```
chrisvanheuveln: You're hard-coding the default value here. Any reason you can't use `default _event_history_cli` instead of `'size_small'`?

andy91315: because that will violate that 80 characters per line.

lib/cisco_node_utils/bgp.rb
```diff
@@ -685,10 +687,15 @@ def router_id
     def router_id=(id)
       # In order to remove a bgp router-id you cannot simply issue
       # 'no bgp router-id'. Dummy-id specified to work around this.
-      dummy_id = '1.2.3.4'
+      # not working on N7K need a valid router_id
+      dummy_id = router_id
       if id == default_router_id
-        @set_args[:state] = 'no'
```
mikewiebe, line 690: Suggest the following instead.

```
    def router_id=(id)
      if id == default_router_id
        # Nothing to do if router_id is already set to default.
        return if router_id == default_router_id
        @set_args[:state] = 'no'
        @set_args[:id] = router_id
      else
        @set_args[:state] = ''
         @set_args[:id] = id
      end
      config_set('bgp', 'router_id', @set_args)
      set_args_keys_default
    end
```

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -228,7 +228,6 @@ def advertise_l2vpn_evpn
     end
 
     def advertise_l2vpn_evpn=(state)
-      Feature.nv_overlay_enable
```
mikewiebe, line 231: Is it ok to remove `Feature.nv_overlay_enable` for non-n7k platforms?

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -244,7 +243,11 @@ def default_advertise_l2vpn_evpn
 
     # dampen_igp_metric
     def dampen_igp_metric
-      config_get('bgp_af', 'dampen_igp_metric', @get_args)
+      match = config_get('bgp_af', 'dampen_igp_metric', @get_args)
```
mikewiebe, line 246: Why is this change needed?

andy91315, line 246: because no default vaule for dampen_igp_metric on n7k.

tests/test_router_bgp.rb
```diff
@@ -777,7 +779,7 @@ def test_routerbgp_set_get_reconnect_interval
   def test_routerbgp_get_reconnect_interval_default
     asnum = 55
     bgp = RouterBgp.new(asnum)
-    assert_equal(60, bgp.reconnect_interval,
+    assert_equal(bgp.default_reconnect_interval, bgp.reconnect_interval,
                  "reconnect_interval should be set to default value of '60'")
```
mikewiebe: Remove the `60` from the test message.

lib/cisco_node_utils/bgp_af.rb
andy91315, line 231: I have tried on n9k it works.

mikewiebe, line 231: @chrisvanheuveln Can you comment on why/if `Feature.nv_overlay_enable` is still needed here?

tests/test_router_bgp.rb
andy91315: It should be 60 when it is supported and nil when it is not supported on the platform.

lib/cisco_node_utils/bgp.rb
glennmatthews: This doesn't make any sense to me as a change. 'size_small' == default_event_history_cli, so this line essentially means:

`return default_event_history_cli if match == default_event_history_cli`

which seems really unnecessary. What problems were you seeing without this check?

(edit:) Additionally, match should be an array of two substrings, not a single string, so I'm not sure this check is ever actually passing.

lib/cisco_node_utils/bgp_af.rb
glennmatthews, line 246: Clearly there is some default value (since you are returning it on the next line!) What is the default on N7k? Why can't you just provide a different default_value in the YAML for N7k rather than making this change?

tests/test_router_bgp.rb
glennmatthews: Right, so the message should be `"reconnect_interval should be set to default value"`

lib/cisco_node_utils/bgp_af.rb
andy91315, line 246: I mean n7k does not show dampen_igp_metric 60 in show run bgp all when nothing configured.
So when nothing we need to return default. And we need to match no dampen_igp_metric. And return nil.

lib/cisco_node_utils/bgp.rb
chrisvanheuveln: @glennmatthews From prev PR #182:

The problem is that the N7k show run all doesn't show defaults; so config_get gets nil, returns the default value (from yaml) which is a string, so match[1] is i (first index in the string size_small).

andy91315: If it does not match, it will return the default value which is a string.

chrisvanheuveln: Right, which makes this code kind of ugly and a bit redundant where it returns the default in two places, however I wasn't able to find a better solution.

lib/cisco_node_utils/bgp_af.rb
chrisvanheuveln, line 246: @glennmatthews I can give you a little more context. Let's discuss this in the office tomorrow.

lib/cisco_node_utils/bgp.rb
andy91315: nothing and event history cli   with no size specific both need to return the default value.
n7k that branch works wierd.

glennmatthews: Chris and I talked through it and I understand the issue now. I think a better fix would be:

1) Add `auto_default: false` to this property in the YAML:

``` diff
 event_history_cli:
+  auto_default: false
   config_get_token_append: '/^(no )?event-history cli(?: size (\S+))?$/'
   config_set_append: '<state> event-history cli <size>'
   default_value: 'size_small'
```

This will cause `config_get` to return `nil` if there is no match instead of returning the string `'size_small'`.

2) Change the getter as follows:

``` diff
     # Nvgen as True With optional 'size <size>
     def event_history_cli
-      return 'false' if match[0] == 'no '
-      return 'size_' + match[1] if match[1]
+      if match.is_a?(Array)
+        return 'false' if match[0] == 'no '
+        return 'size_' + match[1] if match[1]
+      end
       default_event_history_cli
```

lib/cisco_node_utils/bgp_af.rb
glennmatthews, line 246: Seems like the same issue as the discussion above. The correct fix is probably to restore the 'auto_default: false' and handle the case where config_get returns nil.

andy91315:
Thanks Glenn for the fix suggestion. I have changed the code. The auto_default:false makes no difference here. Either way it will return the default value in the get function in .rb file.

lib/cisco_node_utils/bgp.rb
```diff
@@ -685,8 +690,11 @@ def router_id
     def router_id=(id)
       # In order to remove a bgp router-id you cannot simply issue
       # 'no bgp router-id'. Dummy-id specified to work around this.
-      dummy_id = '1.2.3.4'
+      # not working on N7K need a valid router_id
```
glennmatthews: Having comments that contradict one another is not ideal. Would it be better to rework this whole function as:

``` ruby
      def router_id=(id)
       # In order to remove a bgp router-id you cannot simply issue
       # 'no bgp router-id'. On some platforms you can specify a dummy
       # value, but on N7K at least you need the current router_id.
       if id == default_router_id
         # Nothing to do if router_id is already set to default.
         return if router_id == default_router_id
         @set_args[:state] = 'no'
         @set_args[:id] = router_id
       else
         @set_args[:state] = ''
         @set_args[:id] = id
       end
       config_set('bgp', 'router_id', @set_args)
       set_args_keys_default
     end
```

glennmatthews:
:+1:

lib/cisco_node_utils/bgp.rb
mikewiebe: Second this.  This is what I had suggested to you before as well.
https://github.com/cisco/cisco-network-node-utils/pull/195#discussion_r51036558

PR #199, opened 2016-01-28
ipv4_addr_mask_set unset idempotence

glennmatthews:
In doing some testing in the gRPC branch I found that ipv4_addr_mask_set is not safe when attempting to unset an IP address that is already unset - you get a CliError because it's trying to do something like 'no ip address /'. This looks to also be an issue in the release branch so I'm opening this PR to double-commit it to release_1.2.0.

```
Run options: -n test_interface_ipv4_address -- --seed 45791

# Running:


Node under test:
  - name  - n9k-108
  - type  - N9K-C9396PX

.

Finished in 12.178028s, 0.0821 runs/s, 1.3960 assertions/s.

1 runs, 17 assertions, 0 failures, 0 errors, 0 skips
```

### DIFF
--- a/lib/cisco_node_utils/interface.rb
+++ b/lib/cisco_node_utils/interface.rb
@@ -238,9 +238,11 @@ module Cisco
       if addr.nil? || addr == default_ipv4_address
         state = 'no'
         if secondary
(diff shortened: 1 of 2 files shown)


mikewiebe:
:+1:

PR #200, opened 2016-01-28
Feature/stp global

saichint:
stp design assumptions:
There are lot of range based commands in spanning-tree and these ranges dynamically change based on the input. 
To simplify this, we assume: 
- All the multiple ranges given in the manifest will not overlap each other for any given attribute.

For ex:

vlan_hello_time => [['5,7,13,15-42', '10'], ['83-92,100-230', '5']] -----> correct

vlan_hello_time => [['5,7,13,15-42', '10'], ['6-100', '5']] -----> not correct
- If there are multiple ranges, they will be given in sorted order in the manifest (only the ranges and not their associated values or keys). 

For ex: 

vlan_hello_time => [['2,7,13,15-42', '10'], ['3,10-11,100-230', '5']] -----> correct

vlan_hello_time => [['3,10-11,100-230', '5'], ['2,7,13,15-42', '10']] -----> not correct
- The ranges which are not given in the manifest will be set to default.

For ex:
vlan_hello_time => [['2,7,13,15-42', '10'], ['3,10-11,100-230', '5']]
In this case, the vlan_hello_time for [1,4-6,8-9,12,14,43-99,231-3967] will be set to default.
If the manifest is changed to:
vlan_hello_time => [['10-86,99,200-350', '11'], ['3,400-3000', '6']]
then 
vlan_hello_time for ['1-2,4-9,87-98'100-199,351-399,3001-3967] will be set to default.
Basically, the only ranges for which a change is needed should be in the manifest, the rest will be set to default. So if the user needs a particular range to go to default, make sure that the range does not exist in the manifest.

The minitest results are attched below:
There are some nxos issues on n9k and n6k.
- On n6k, the vlan hello-time, the 'no' cmd cannot be used due to an issue so a workaround is put in place.
- On n9k and n6k, "mst root priority" is not being shown properly in the running config unless "mst designated priority" is set first. There is no real dependency between these attributes but this is due to a bug. The work around is to set "mst designated priority" first and this was done in the test_stp_global.rb

All the bugs needed are filed on n9k and n6k already.

Note: bridge_domain related attributes are available only on n7k.
[stp_mini_test.txt](https://github.com/cisco/cisco-network-node-utils/files/112803/stp_mini_test.txt)


saichint:
I will resubmit this shortly. The per interface stp works on n9k and n6k so I will redo the testing just to be sure.

saichint:
There is a design issue pending to see if we need to purge the configuration or not and I am closing this PR until we have resolution on that

PR #201, opened 2016-01-29
Build minitest interface list using `Interface.interfaces`.

mikewiebe:
This update modifies our `ciscotest.rb` `interfaces` api to build the list of available interfaces dynamically rather then rely on the hardcoded interface list in `tests/platform_info.yaml`.

I scrubbed the list of minitests that were using interfaces and only found a couple of cases where the interface names were hard-coded in the test.  I fixed those.
# Testing
- `test_interface.rb` and `test_interface_switchport.rb` pass  or same as baseline on `n3k`, `n9k`, `n5k`, `n6k`, `n7k`.
- `n3k` interface range `Ethernet 1/1 - 1/52`
- `n9k` interface range `Ethernet 1/1 - 1/48`
- `n5k` interface range `Ethernet 1/1 - 1/52`
- `n6k` interface range `Ethernet 1/1 - 1/48`
- `n7k-1` interface range `Ethernet 3/1 - 3/32`
- `n7k-2` No Interfaces.
## Test on `n7k` without interfaces.

```
rtp-ads-432:/nobackup/mwiebe/REPOS_NODE_UTILS/release120/cisco-network-node-utils> ruby tests/test_interface.rb -v -- n7k-9 admin admin
Run options: -v -- --seed 26525

# Running:


Node under test:
  - name  - n7k-9
  - type  - N7K-C7010
  - image - bootflash:///n7000-s2-dk9.7.3.0.D1.0.160.bin

No suitable interfaces found on n7k-9 for this test
Coverage report generated for MiniTest to /nobackup/mwiebe/REPOS_NODE_UTILS/release120/cisco-network-node-utils/coverage. 562 / 1201 LOC (46.79%) covered.
rtp-ads-432:/nobackup/mwiebe/REPOS_NODE_UTILS/release120/cisco-network-node-utils> 

```
## Test on `n7k` without interfaces, but test does not require interfaces.

```
Node under test:
  - name  - n7k-9
  - type  - N7K-C7010
  - image - bootflash:///n7000-s2-dk9.7.3.0.D1.0.160.bin

TestRouterOspf#test_routerospf_collection_empty = 2.91 s = .
TestRouterOspf#test_routerospf_create_valid_multiple = 5.53 s = .
TestRouterOspf#test_routerospf_get_name = 3.88 s = .
TestRouterOspf#test_routerospf_create_valid_no_feature = 4.03 s = .
TestRouterOspf#test_routerospf_destroy = 3.92 s = .
TestRouterOspf#test_routerospf_collection_not_empty = 5.32 s = .
TestRouterOspf#test_routerospf_create_name_zero_length = 0.89 s = .
TestRouterOspf#test_routerospf_create_valid_multiple_delete_one = 6.10 s = .
TestRouterOspf#test_routerospf_create_valid = 3.94 s = .

Finished in 36.519269s, 0.2464 runs/s, 1.1775 assertions/s.

9 runs, 43 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for MiniTest to /nobackup/mwiebe/REPOS_NODE_UTILS/release120/cisco-network-node-utils/coverage. 591 / 1195 LOC (49.46%) covered.
```


tests/ciscotest.rb
```diff
@@ -66,10 +67,15 @@ def interfaces
     unless @@interfaces
       # Build the platform_info, used for interface lookup
       # rubocop:disable Style/ClassVars
-      platform_info = PlatformInfo.new(node.host_name, :nexus)
-      @@interfaces = platform_info.get_value_from_key('interfaces')
+      @@interfaces = []
+      Interface.interfaces.each do |int, _obj|
```
chrisvanheuveln: nit: equivalent is:

`Interface.interfaces.keys.each do |int|`

Fine as is though.

chrisvanheuveln:
:+1:

glennmatthews:
My main concern on this is that it provides no way for us to say "Don't touch this interface!". On Nexus, fine, you can implicitly ignore 'mgmt0' and assume that all Ethernet interfaces are fair game (or can you?), but on XR, depending on the device, you may need to ignore MgmtEthernet or maybe you need to ignore one of the GigabitEthernet devices. Unless we're going to explicitly specify that the UUT must be managed by the mgmt interface only and all other interfaces are presumed fair game, this seems risky.

chrisvanheuveln:
@glennmatthews This still seems better than assuming that all testbeds have `Ethernet1/1`. 
The updated method returns a dynamic list of interfaces that we can filter depending on our needs. Is there some middle ground for this solution that fits XR requirements?

glennmatthews:
Would it make sense to only select interfaces that are admin down, or have no IP address, or criteria like that? Basically, if the interface already has non-default configuration, assume no touching it?

mikewiebe:
Today with our hard coded solution we make no such distinctions right?  We assume the 3 interfaces defined in the yaml are fair game.  The interface tests themselves are quite destructive :).

glennmatthews:
Right, yes - but in the YAML I can specify interfaces that I know are safe to destroy.

mikewiebe:
The non-interface minitests are also inherently destructive.  Nuke bgp, ospf and vlans for example... not sure protecting the interfaces make much of a difference.  How do you know that eth1/1 - 1/3 are safe to destroy?  Are you suggesting users will update the yaml before running the test?

mikewiebe:
@glennmatthews @chrisvanheuveln Had a quick syncup with Glenn and agreed that I will only add ethernet interfaces to the list that don't have ip's configured.  This way we don't shoot ourselves in the foot and blow away the interface we are using to run the minitests.

Thanks for the feedback guys!

chrisvanheuveln:
Per our secondary sync up:
- the non-ip'd interfaces approach is problematic when re-running a crashed test
- a better fix is to just skip the interface that is using the ip address that was specified on the minitest cmdline call
  - If minitest is called with a dns name then do an nslookup
  - if nslookup fails then fail test and force them to run it with a static ip address

mikewiebe:
@glennmatthews @chrisvanheuveln Updated per our conversation.  I ended up using ruby's Resolv class for the dnslookup (http://ruby-doc.org/stdlib-2.0.0/libdoc/resolv/rdoc/Resolv.html).

tests/ciscotest.rb
```diff
@@ -62,14 +65,40 @@ def config(*args)
...
+  def convert_dns_name(ip)
+    ip_address?(ip) ? ip : Resolv.getaddress(ip)
+  rescue Resolv::ResolvError
+    raise "Unable to resolve name #{ip}. Use static ip to connect instead!"
+  end
+
+  def address_match?(int_ip)
+    # Compare the interface address with the current session address.
+    # and return true if they match.
+    return false if int_ip.nil?
```
chrisvanheuveln, line 83: You don't need this line. You'll fail the next check if int_ip is nil so you return false regardless.

The rest looks good.

mikewiebe, line 83: I agree, but I intentionally left this line because we should return early and not even do the dns conversion, address check etc when nil.  No point.

The last `false` just just a fail safe.

chrisvanheuveln, line 83: ack

tests/ciscotest.rb
```diff
@@ -62,14 +65,40 @@ def config(*args)
     result
   end
 
+  def ip_address?(ip)
+    return true if IPAddr.new(ip).ipv4?
```
glennmatthews: Does this properly return false for IPv6? I would have expected this to be written as `return IPAddr.new(ip).ipv4?`

tests/ciscotest.rb
```diff
@@ -62,14 +65,40 @@ def config(*args)
...
+  rescue Resolv::ResolvError
+    raise "Unable to resolve name #{ip}. Use static ip to connect instead!"
+  end
+
+  def address_match?(int_ip)
+    # Compare the interface address with the current session address.
+    # and return true if they match.
+    return false if int_ip.nil?
+    return true if int_ip == convert_dns_name(address)
+    false
```
glennmatthews: How about:
`return int_ip == convert_dns_name(address)`

glennmatthews:
Couple of nits but looks good - and feels a lot less risky to me. Thanks for reworking this. :+1:

tests/ciscotest.rb
mikewiebe: Done

tests/ciscotest.rb
mikewiebe: argg.. what, you don't like `return true if true`? :)

mikewiebe: Yes for the IPv6 case.

```
2.2.1 :006 > IPAddr.new('2040::1').ipv4?
 => false 
2.2.1 :007 > 
```

PR #205, opened 2016-01-29
Document process for adding XR support

chburket:
I used bgp_neighbor as an example for the documentation.  While writing the doc, I found inconsistencies and a few errors, so I fixed them in bgp_neighbor.

### DIFF
--- a/docs/README-node-utils-APIs-xr.md
+++ b/docs/README-node-utils-APIs-xr.md
@@ -1,4 +1,4 @@
-# How To Create New node_utils APIs
+# How To Update node_utils APIs to Support XR
 
 #### Table of Contents
 
@@ -122,7 +122,7 @@ First, we exclude the attribute in bgp_neighbor.yaml:
     default_value: false
 ```
 
-This will cause any calls to config_set for this attribute to raise a Cisco::UnsupportedError, and  calls to config_get and config_get_default will return nil.  Because of this, we must update the test_bgp_neighbor.rb minitest to expect these conditions on XR (see [Development Best Practices: MT6](./README-develop-best-practices.md#mt6)):
+This will cause any calls to config_set for this attribute to raise a Cisco::UnsupportedError, and calls to config_get and config_get_default will return nil.  Because of this, we must update the test_bgp_neighbor.rb minitest to expect these conditions on XR (see [Development Best Practices: MT6](./README-develop-best-practices.md#mt6)):
 
 ```ruby
   def test_low_memory_exempt
@@ -169,7 +169,7 @@ RP/0/RP0/CPU0:agent-lab20-xr(config)#router bgp 55 neighbor 1.1.1.1  update-sour
 % Invalid input detected at '^' marker.
 ```
 
-Although the formatting makes it difficult to tell, the '^' marker is pointing to "Ethernet1/1" in the above example.  Instead of using hardcoded interface names, use the interfaces[] array (see [Development Best Practices: MT3](./README-develop-best-practices.md#mt3)).
+As you can see in CLI output, the '^' marker is indicating that "Ethernet1/1" is invalid.  Instead of using hardcoded interface names, use the interfaces[] array (see [Development Best Practices: MT3](./README-develop-best-practices.md#mt3)).
 
 ```ruby test_bgp_router.rb
   def test_update_source
@@ -189,7 +189,7 @@ Although the formatting makes it difficult to tell, the '^' marker is pointing t
 
 #### <a name="setup">Setup/teardown differences</a>
 
-Your minitest's setup method is run before each test, and the teardown method is run after each test.  These methods often contain calls to the TestCase.config method to execute commands to clear and setup existing configuration on the device, and often, the commands will differ between platforms. Currently, the config method does not display the output from any executed commands, even when debug is enabled, so any errors that occur will be masked.
+Your minitest's setup method is run before each test, and the teardown method is run after each test.  These methods often contain calls to the TestCase.config method to execute commands to clear existing configuration on the device, and often, the commands will differ between platforms. Currently, the config method does not display the output from executed commands, even when debug is enabled, so any errors that occur will be masked.
 
 You should verify that any existing config commands are valid for XR (either by sight, or by executing them manually on the XR CLI).  The following changes were needed for test_bgp_neighbor.rb:
 
@@ -216,9 +216,9 @@ You should verify that any existing config commands are valid for XR (either by 
 
 #### <a name="dependency">Dependency differences</a>
 
-While enhancing the Cisco BGP features to support IOS XR, we found that configuring some attributes would require that other attributes be configured in a certain way, first (sometimes in the same sub-mode, sometimes in a parent mode).  Usually, we were able to solve this by creating helper methods in the minitest to set up any dependencies.
+While enhancing the Cisco BGP features to support IOS XR, we found that configuring some attributes would require that other attributes be configured in a certain way, first (sometimes in the same sub-mode, sometimes in a parent mode).  Usually, we were able to solve this by creating helper methods in the minitests to set up any dependencies.
 
-As an example, XR requires the remote-as attribute to be set before any other attribute can be set.  Attempting to set the description on a neighbor without first setting the remove-as results in the following error:
+As an example, XR requires the bgp neighbor remote-as attribute to be set before any other neighbor attribute can be set.  Attempting to set the description on a neighbor without first setting the remote-as results in the following error:
 
 ```bash
   1) Error:
@@ -268,7 +268,7 @@ Other examples of BGP dependency issues we ran into on XR that did not exist on 
 Each feature will be different, so it there will almost certainly be unique changes needed to support XR for a particular feature.  For bgp neighbor, that included the following:
 
  - XR does not support bgp neighbor addresses in the prefix/len format ("2.2.2.2/24"), so we had to skip those in the minitest on XR.
- - XR only supports a single encryption type for the password attribute (simply specified as "encrypted", which maps to "md5").  We updated the ruby API to fail with a Cisco::UnsupportedError if an encryption type other than md5 is specified on XR (and updated the minitest to match).
+ - XR only supports a single encryption type for the password attribute (simply specified as "encrypted", which is actually "md5").  We updated the node_utils API to fail with a Cisco::UnsupportedError if an encryption type other than md5 is specified on XR (and updated the minitest to match).
  - XR supports three transport connection modes (active-only, passive-only, both), while Nexus only supports two (passive-only, both) through the transport_passive_only attribute.  We added a new transport_passive_mode attribute that accepts all three types.
 
 
--- a/docs/README-node-utils-APIs-xr.md
+++ b/docs/README-node-utils-APIs-xr.md
@@ -1,0 +1,84 @@
+# How To Create New node_utils APIs
+
+#### Table of Contents
+
+* [Overview](#overview)
+* [Before You Begin](#prerequisites)
+* [Start here: Fork and Clone the Repo](#clone)
+* [Example: router eigrp](#complex)
+ * [Step 1. YAML Definitions: router eigrp](#comp_yaml)
+ * [Step 2. Create the node_utils API: router eigrp](#comp_api)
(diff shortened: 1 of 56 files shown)


docs/README-node-utils-APIs-xr.md
```diff
@@ -0,0 +1,277 @@
...
+
+Now run the same tests against XR, to see what is broken and to get an idea of how much work will need to be done.
+
+```bash
+% ruby test_bgp_neighbor.rb -v -- 192.168.0.2:57777 admin admin
+```
+
+## <a name="fix-errors">Fix errors</a>
+
+We'll now go through the types of differences that will cause errors/failures, and how to fix them.  When you've fixed a particular error, run minitest again on that single test until it is working as you intend, then re-run the same test against Nexus to make sure it still works.
```
glennmatthews: Suggestion:

...run minitest again on that single test (`ruby test_bgp_neighbor.rb -n test_case_name_goes_here ...`) until it is working...

docs/README-node-utils-APIs-xr.md
```diff
@@ -0,0 +1,277 @@
...
+
+#### <a name="get_command">get_command differences</a>
+
+The base get_command parameter will often be different on XR than Nexus.  Usually, the get_command is common across many attributes of a given feature, and therefore will be defined in the _template section.  In the case of the bgp neighbor feature, the get_command parameter to retrieve information about neighbors is "show running bgp all" on Nexus, while it is "show running router bgp" on XR.  This difference caused most/all of the test_bgp_neighbor.rb tests to error when run against XR, which looked something like this:
```
glennmatthews: General nitpick - please format literal strings, filenames, and YAML entries in `monospace`. This would include `get_command`, `_template`, `"show running bgp all"`, `test_bgp_neighbor.rb`. Thanks.

docs/README-node-utils-APIs-xr.md
```diff
@@ -0,0 +1,277 @@
...
+      - "router bgp <asnum>"
+      - "(?)vrf <vrf>"
+      - "(?)neighbor <nbr>"
+```
+
+Here, the "global" get_command parameter was converted into two platform-specific parameters, while the context remained the same.
+
+#### <a name="get_value">get\_value/set\_value differences</a>
+
+Similar to the get_command differences above, the feature might have an attribute that is functionally equivalent on Nexus and XR, but have small syntactic differences.  An example of this for bgp neighbor is the "connected-check" attribute which produced the following error on XR:
```
glennmatthews: nit: connected_check not connected-check, since we're talking about the attribute rather than the CLI.

docs/README-node-utils-APIs-xr.md
```diff
@@ -0,0 +1,277 @@
...
+
+>>>show-cmd: show running bgp all <<<
+show running bgp all
+              ^
+% Invalid input detected at '^' marker.'
+```
+
+To correct this problem, the following changes were made in bgp_neighbor.yaml:
+
+```
```
glennmatthews: Please mark YAML blocks with ````yaml` so we get the nice syntax highlighting on GitHub. :)
(Edit: Or ````diff` if you prefer diff highlighting rather than YAML highlighting - your call)

docs/README-node-utils-APIs-xr.md
```diff
@@ -0,0 +1,277 @@
...
+  low_memory_exempt:
++   _exclude: [ios_xr]
+    kind: boolean
+    get_value: '/^low-memory exempt$/'
+    set_value: '<state> low-memory exempt'
+    default_value: false
+```
+
+This will cause any calls to config_set for this attribute to raise a Cisco::UnsupportedError, and calls to config_get and config_get_default will return nil.  Because of this, we must update the test_bgp_neighbor.rb minitest to expect these conditions on XR (see [Development Best Practices: MT6](./README-develop-best-practices.md#mt6)):
```
glennmatthews: `config_set`, `config_get`, `config_get_default`, `Cisco::UnsupportedError` please. :-)

docs/README-node-utils-APIs-xr.md
```diff
@@ -0,0 +1,277 @@
...
+
+```ruby
+  def test_low_memory_exempt
+    %w(default test_vrf).each do |vrf|
+      neighbor = create_neighbor(vrf)
++     if platform == :ios_xr
```
glennmatthews: Out of scope for this PR perhaps - but it should be really trivial for us to create a helper function in `ciscotest.rb` for this pattern - something like:

``` ruby
assert_unsupported(object, property_name)
  assert_nil(object.send(property_name))
  assert_nil(object.send("default_#{property_name}"))
  assert_raises(Cisco::UnsupportedError) { object.send("#{property_name}=", true) }
end
```

docs/README-node-utils-APIs-xr.md
```diff
@@ -0,0 +1,277 @@
...
+RP/0/RP0/CPU0:agent-lab20-xr#conf
+Fri Jan 29 14:02:50.566 UTC
+RP/0/RP0/CPU0:agent-lab20-xr(config)#router bgp 55 neighbor 1.1.1.1  update-source Ethernet1/1
+                                                                                   ^
+% Invalid input detected at '^' marker.
+```
+
+As you can see in CLI output, the '^' marker is indicating that "Ethernet1/1" is invalid.  Instead of using hardcoded interface names, use the interfaces[] array (see [Development Best Practices: MT3](./README-develop-best-practices.md#mt3)).
+
+```ruby test_bgp_router.rb
```
glennmatthews: Need a newline in here:

``````
```ruby
test_bgp_router.rb
``````

Or just remove the filename - your call, but this isn't rendering properly as-is.

File: docs/README-node-utils-APIs-xr.md [hunk 8]
```diff
@@ -0,0 +1,277 @@
...
+        assert_equal(interface.downcase, neighbor.update_source)
+      end
+      neighbor.destroy
+    end
+  end
+```
+
+#### <a name="setup">Setup/teardown differences</a>
+
+Your minitest's setup method is run before each test, and the teardown method is run after each test.  These methods often contain calls to the TestCase.config method to execute commands to clear existing configuration on the device, and often, the commands will differ between platforms. Currently, the config method does not display the output from executed commands, even when debug is enabled, so any errors that occur will be masked.
```
glennmatthews: Did you get a chance to open a User Story for enhancing this method?

docs/README-node-utils-APIs-xr.md
```diff
@@ -0,0 +1,277 @@
...
+Each feature will be different, so it there will almost certainly be unique changes needed to support XR for a particular feature.  For bgp neighbor, that included the following:
+
+ - XR does not support bgp neighbor addresses in the prefix/len format ("2.2.2.2/24"), so we had to skip those in the minitest on XR.
+ - XR only supports a single encryption type for the password attribute (simply specified as "encrypted", which is actually "md5").  We updated the node_utils API to fail with a Cisco::UnsupportedError if an encryption type other than md5 is specified on XR (and updated the minitest to match).
+ - XR supports three transport connection modes (active-only, passive-only, both), while Nexus only supports two (passive-only, both) through the transport_passive_only attribute.  We added a new transport_passive_mode attribute that accepts all three types.
+
+
+## Conclusion
+
+This was hopefully a good overview for updating an existing Cisco node_utils API to support IOS XR. It is not intended to cover every possibly case, but might be updated in the future with other common cases.
```
glennmatthews: nit: s/possibly/possible/

lib/cisco_node_utils/cmd_ref/bgp_neighbor.yaml
```diff
@@ -8,7 +8,7 @@ _template:
   context:
     - "router bgp <asnum>"
     - "(?)vrf <vrf>"
-    - "neighbor <nbr>"
+    - "(?)neighbor <nbr>"
```
glennmatthews: Not in favor of this change as neighbor is _generally_ mandatory. There are only a few specific cases in this file where it is optional (`all_neighbors` is the only one I can think of offhand) and in those cases they explicitly redefine their context. For every other property in this file, it's mandatory and should be kept as such.

Same comment for the next file as well.

glennmatthews:
Nice work Charles - a few formatting nitpicks and such but the content is very readable and thorough.

glennmatthews:
Two other nits:
- Looks like we're following the convention of 'README-develop-*' for developer guides. So this file may need to be changed to 'README-develop-add-XR-support' or similar.
- Please add a link to this document under the 'Developer Guides' heading in the main README.md.

docs/README-node-utils-APIs-xr.md
chburket: agreed

docs/README-node-utils-APIs-xr.md
chburket: Yes, US67139

lib/cisco_node_utils/cmd_ref/bgp_neighbor.yaml
chburket: Both were marked as optional before your yaml refactoring, and the minitest failed with it marked mandatory.  I can double check that, though.

glennmatthews: Please do double-check and follow up with me if there are issues - would prefer to separate this change from the doc updates if possible.

chburket: I reverted the yaml.  It was only failing on my "old" version of bgp_neighbor.yaml because I had not changed the all_neighbors entry.

chburket:
I think I've fixed everything, but could certainly have missed some things that need `monospace`.

docs/REAME-develop-add-XR-support.md
```diff
@@ -0,0 +1,277 @@
...
++     get_command: "show running router bgp"
++   nexus:
++     get_command: "show running bgp all"
+    context:
+      - "router bgp <asnum>"
+      - "(?)vrf <vrf>"
+      - "(?)neighbor <nbr>"
+```
+
+Here, the "global" get_command parameter was converted into two platform-specific parameters, while the context remained the same.
```
glennmatthews: `get_command`

docs/REAME-develop-add-XR-support.md
```diff
@@ -0,0 +1,277 @@
...
+```yaml
+  low_memory_exempt:
++   _exclude: [ios_xr]
+    kind: boolean
+    get_value: '/^low-memory exempt$/'
+    set_value: '<state> low-memory exempt'
+    default_value: false
+```
+
+This will cause any calls to config_set for this attribute to raise a `Cisco::UnsupportedError`, and calls to `config_get` and `config_get_default` will return `nil`.  Because of this, we must update the `test_bgp_neighbor.rb` minitest to expect these conditions on XR (see [Development Best Practices: MT6](./README-develop-best-practices.md#mt6)):
```
glennmatthews: `config_set`

docs/REAME-develop-add-XR-support.md
```diff
@@ -0,0 +1,277 @@
...
+        assert_equal(interface.downcase, neighbor.update_source)
+      end
+      neighbor.destroy
+    end
+  end
+```
+
+#### <a name="setup">Setup/teardown differences</a>
+
+Your minitest's setup method is run before each test, and the teardown method is run after each test.  These methods often contain calls to the `TestCase.config` method to execute commands to clear existing configuration on the device, and often, the commands will differ between platforms. Currently, the `config` method does not display the output from executed commands, even when debug is enabled, so any errors that occur will be masked.
```
glennmatthews: `setup`, `teardown`

docs/REAME-develop-add-XR-support.md
```diff
@@ -0,0 +1,277 @@
...
+
+You should verify that any existing config commands are valid for XR (either by sight, or by executing them manually on the XR CLI).  The following changes were needed for `test_bgp_neighbor.rb`:
```
glennmatthews: General nit - we probably shouldn't be referring to 'XR' but should instead say 'IOS XR' throughout.

glennmatthews:
A couple more nits, and please add the README.md link. Thanks!

chburket:
Pushed latest changes.

PR #207, opened 2016-02-01
Remove bgp dot-notation handlers

chrisvanheuveln:
I discovered a few issues while rewriting the bgp global beaker test:
- Our BGP providers were normalizing the ASN to AS-PLAIN but the nxos cli doesn't support switching between AS-PLAIN and AS-DOT notations, even when the asn is equivalent; ie. once a tag syntax is used you must continue using it:

```
n7k8(config)# router bgp 1.4
n7k8(config-router)# router bgp 65540
bgp is already running; tag is 1.4.
```

Thus, I'm removing the asnum normalizer methods `process_asnum` and `dot_to_big`
- Fixed a day 1 problem with the yaml regexp which failed to match against AS-DOT notation
- Fixed event_history_detail. We use `auto_default: false` with this one because the N7k doesn't show a value with show run all, which means config_get returns [nil, nil].
- Misc minitest fixes. All of the bgp minitests are now passing.

### DIFF
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -26,7 +26,7 @@ module Cisco
     def initialize(asnum, vrf='default', instantiate=true)
       fail ArgumentError unless vrf.is_a? String
       fail ArgumentError unless vrf.length > 0
-      @asnum = RouterBgp.process_asnum(asnum)
+      @asnum = asnum.to_s
       @vrf = vrf
       if @vrf == 'default'
         @get_args = @set_args = { asnum: @asnum }
@@ -34,19 +34,6 @@ module Cisco
         @get_args = @set_args = { asnum: @asnum, vrf: @vrf }
       end
       create if instantiate
-    end
-
-    def self.process_asnum(asnum)
-      err_msg = "BGP asnum must be either a 'String' or an" \
-                " 'Integer' object"
-      fail ArgumentError, err_msg unless asnum.is_a?(Integer) ||
-                                         asnum.is_a?(String)
-      if asnum.is_a? String
-        # Match ASDOT '1.5' or ASPLAIN '55' strings
-        fail ArgumentError unless /^(\d+|\d+\.\d+)$/.match(asnum)
-        asnum = RouterBgp.dot_to_big(asnum) if /\d+\.\d+/.match(asnum)
-      end
-      asnum.to_i
     end
 
     # Create a hash of all router bgp default and non-default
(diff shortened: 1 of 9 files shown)


lib/cisco_node_utils/bgp.rb
```diff
@@ -26,7 +26,7 @@ class RouterBgp < NodeUtil
     def initialize(asnum, vrf='default', instantiate=true)
       fail ArgumentError unless vrf.is_a? String
       fail ArgumentError unless vrf.length > 0
-      @asnum = RouterBgp.process_asnum(asnum)
+      @asnum = asnum.to_s
```
robert-w-gries: Seeing this error on 3k now

```
Node under test:
  - name  - n3k-51
  - type  - N3K-C3048TP-1GE
  - image - 

TestRouterBgp#test_routerbgp_get_asnum = 10.40 s = F

Finished in 10.406674s, 0.0961 runs/s, 0.0961 assertions/s.

  1) Failure:
TestRouterBgp#test_routerbgp_get_asnum [tests/test_router_bgp.rb:154]:
Error: router asnum not correct.
Expected: 55
  Actual: "55"
```

mikewiebe: Looks good other then this.

glennmatthews:
XR CLI has the same restriction, it appears, so I'm in favor of this change:

```
router bgp 65540
!!% The instance name is used already: asn 1.4 inst-name default
```

glennmatthews:
I'm a bit surprised that this scenario wasn't covered by the testing before. Since you're removing the old functionality, does it make sense to add some negative tests around trying to change the format?

Chris: I'll add a test or two.

lib/cisco_node_utils/bgp.rb
```diff
@@ -36,19 +36,6 @@ def initialize(asnum, vrf='default', instantiate=true)
...
     end
 
-    def self.process_asnum(asnum)
-      err_msg = "BGP asnum must be either a 'String' or an" \
-                " 'Integer' object"
-      fail ArgumentError, err_msg unless asnum.is_a?(Integer) ||
-                                         asnum.is_a?(String)
-      if asnum.is_a? String
-        # Match ASDOT '1.5' or ASPLAIN '55' strings
-        fail ArgumentError unless /^(\d+|\d+\.\d+)$/.match(asnum)
```
glennmatthews, line 46: I see that you've changed the tests to look for a CliError instead of this ArgumentError but IMHO the ArgumentError is more intuitive to troubleshoot. Would it make sense to keep the validation logic (but not the conversion) in this function and rename it to `self.validate_asnum`?

lib/cisco_node_utils/bgp.rb
chrisvanheuveln: fixed

lib/cisco_node_utils/bgp.rb
chrisvanheuveln, line 46: Sure that sounds reasonable. Of course now I get to put all those calls back in :-)

mikewiebe:
:+1:

PR #209, opened 2016-02-02
BGP minitest fixes

chrisvanheuveln:
- pim all_rp_addresses had a greedy regexp
- added conditional, non config() cleanups to test_\* setup/teardowns
- added l2vpn evpn checks to neighbor af

### DIFF
--- a/tests/ciscotest.rb
+++ b/tests/ciscotest.rb
@@ -113,4 +113,26 @@ class CiscoTestCase < TestCase
     end
     @@interfaces_id
   end
+
+  # Remove all router bgps.
+  def remove_all_bgps
+    require_relative '../lib/cisco_node_utils/bgp'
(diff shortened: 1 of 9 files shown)


mikewiebe:
:+1: Pending results of 5/6k tests.

tests/ciscotest.rb
```diff
@@ -113,4 +113,26 @@ def interfaces_id
     end
     @@interfaces_id
   end
+
+  # Remove all router bgps.
```
glennmatthews, line 117: I'd prefer to have an intermediate 'bgptest.rb' class (TestBgpAF < BgpTest < CiscoTest < TestCase) rather than putting feature-specific code in ciscotest.rb - but not critical.

Alternately, would it make sense to add these methods to the NU classes bgp.rb and vrf.rb?

glennmatthews:
:+1:

PR #213, opened 2016-02-04
fix route_distinguisher

HelenHZhang:
make vrf context optional for route_distinguisher
All mini tests passed for nexus, and ios_xr

### DIFF
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -808,6 +808,7 @@ module Cisco
     end
 
     def default_route_distinguisher
+      return nil if @vrf.nil? || @vrf == 'default'
       config_get_default('bgp', 'route_distinguisher')
     end
 
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -786,11 +786,7 @@ module Cisco
     # route_distinguisher
     # Note that this property is supported by both bgp and vrf providers.
     def route_distinguisher
-      if @vrf.nil? || @vrf == 'default'
-        fail Cisco::UnsupportedError.new('bgp', 'route_distinguisher', 'get',
-                                         'route_distinguisher is not ' \
-                                         'configurable on a default VRF')
-      end
+      return nil if @vrf.nil? || @vrf == 'default'
       config_get('bgp', 'route_distinguisher', @get_args)
     end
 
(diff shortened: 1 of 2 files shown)


glennmatthews:
This change doesn't make sense to me. The route_distinguisher only makes sense in a VRF context. We should never be trying to set or get it without a VRF.

HelenHZhang:
This change is needed in order to get beaker tests running for test_bgpneighboraf.
Without it, beaker tests was blocked for nexus.
Helen

glennmatthews:
It's not the right change though. Positive tests (get/set under a VRF) should be working without this change, and negative tests (get/set in a non-VRF context) should address this in some other way. Probably the more correct approach is to fix `route_distinguisher` and `route_distinguisher=` in bgp.rb.

lib/cisco_node_utils/bgp.rb
```diff
@@ -794,6 +795,7 @@ def route_distinguisher=(rd)
         Feature.nv_overlay_enable
         Feature.nv_overlay_evpn_enable
       end
+      return nil if @vrf.nil? || @vrf == 'default'
```
glennmatthews: This should be a check _before_ the platform check (no point in enabling features if we're not going to actually use them).

Also, the convention is that this, being a setter method, should raise a Cisco::UnsupportedError, rather than being a no-op.

lib/cisco_node_utils/bgp.rb
```diff
@@ -786,6 +786,7 @@ def default_reconnect_interval
     # route_distinguisher
     # Note that this property is supported by both bgp and vrf providers.
     def route_distinguisher
+      return nil if @vrf.nil? || @vrf == 'default'
```
glennmatthews, line 789: Same logic is needed for `default_route_distinguisher`.

lib/cisco_node_utils/bgp.rb
```diff
@@ -786,10 +786,20 @@ def default_reconnect_interval
     # route_distinguisher
     # Note that this property is supported by both bgp and vrf providers.
     def route_distinguisher
+      if @vrf.nil? || @vrf == 'default'
+        fail Cisco::UnsupportedError.new('bgp', 'route_distinguisher', 'get',
```
glennmatthews: Logic was right for this method before. Getter and default should return nil if unsupported - only setter raises an exception.

HelenHZhang: will reverse the getter then.

glennmatthews: Default also please.

PR #221, opened 2016-02-10
cisco_vxlan_vtep: added property source-interface hold-down-time

alok-aggarwal:
Rubocop passes:

Minitest passes:

```
# Running:


Node under test:
  - name  - agent-lab7-nx
  - type  - N9K-NXOSV
  - image - bootflash:///nxos.7.0.3.I2.2.22.bin

TestVxlanVtep#test_host_reachability = 142.06 s = .
TestVxlanVtep#test_create_negative = 142.10 s = .
TestVxlanVtep#test_create_destroy_one = 134.90 s = .
TestVxlanVtep#test_mt_full_create_destroy_multiple = 1.58 s = S
TestVxlanVtep#test_source_interface_hold_down_time = 135.70 s = .
TestVxlanVtep#test_shutdown = 135.42 s = .
TestVxlanVtep#test_source_interface = 137.44 s = .
TestVxlanVtep#test_description = 134.57 s = .

Finished in 963.759647s, 0.0083 runs/s, 0.0259 assertions/s.

  1) Skipped:
TestVxlanVtep#test_mt_full_create_destroy_multiple [tests/test_vxlan_vtep.rb:89]:
Platform does not support MT-full
```

### DIFF
--- a/lib/cisco_node_utils/vxlan_vtep.rb
+++ b/lib/cisco_node_utils/vxlan_vtep.rb
@@ -140,9 +140,6 @@ module Cisco
     end
 
     def source_interface_hold_down_time=(time)
-      # Source interface should be configured before hold down time
-      # is configured.
-      return if source_interface == default_source_interface
       state = time == default_source_interface_hold_down_time ? 'no' : ''
       # Cli rejects removing hold-down-time without an argument, so make
       # sure it is configured before attempting to remove it
--- a/lib/cisco_node_utils/vxlan_vtep.rb
+++ b/lib/cisco_node_utils/vxlan_vtep.rb
@@ -135,6 +135,33 @@ module Cisco
       config_get_default('vxlan_vtep', 'source_intf')
     end
 
+    def source_interface_hold_down_time
+      config_get('vxlan_vtep', 'source_intf_hold_down_time', name: @name)
+    end
+
+    def source_interface_hold_down_time=(time)
+      # Source interface should be configured before hold down time
+      # is configured.
+      return if source_interface == default_source_interface
+      state = time == default_source_interface_hold_down_time ? 'no' : ''
+      # Cli rejects removing hold-down-time without an argument, so make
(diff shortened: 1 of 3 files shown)


lib/cisco_node_utils/vxlan_vtep.rb
```diff
@@ -135,6 +135,33 @@ def default_source_interface
...
     end
 
+    def source_interface_hold_down_time
+      config_get('vxlan_vtep', 'source_intf_hold_down_time', name: @name)
+    end
+
+    def source_interface_hold_down_time=(time)
+      # Source interface should be configured before hold down time
+      # is configured.
+      return if source_interface == default_source_interface
```
glennmatthews: This seems wrong - the setter should not silently be a no-op.

alok-aggarwal: I have added a validate check in the corresponding puppet file that requires interface to be set before the hold_down_time. Would that suffice?

alok-aggarwal: Or is there a better way to handle this independently in NU?

glennmatthews: What happens if you try to set the hold-down-time in the CLI without setting the interface? Does it generate a meaningful error message?

alok-aggarwal: Yes it gives a meaningful message:
ERROR: Source loopback not configured

alok-aggarwal: I think it makes sense just to remove this check, will do it.

alok-aggarwal: @glennmatthews Removed the check.

chrisvanheuveln:
`release_1.2.0` branch is closed so I'll reopen this PR against develop.

PR #234, opened 2016-02-16
Add test_feature.rb minitest

chrisvanheuveln:
- Tested affected minitests on:
  - I3 images:  n3048, N9k, N7k, N6k, N5k, N31xx
  - I2 images:  n3048

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/vlan.yaml
+++ b/lib/cisco_node_utils/cmd_ref/vlan.yaml
@@ -16,8 +16,7 @@ mapped_vni:
     kind: int
     config_get: 'show running vlan'
     config_get_token: ['/^vlan <vlan>$/', '/^vn-segment (\d+)$/']
-    # In the cli 'vn-segment' config is not applied untill the 'end'
-    # statement is issued.
+    # The 'vn-segment' config is not applied until the 'end' statement is processed
     config_set: ['vlan <vlan>', '<state> vn-segment <vni> ; end']
     default_value: ''
   N9k: *mapped_vni_n3k_n9k
--- a/lib/cisco_node_utils/vrf.rb
+++ b/lib/cisco_node_utils/vrf.rb
@@ -95,8 +95,8 @@ module Cisco
     def route_distinguisher=(rd)
       # feature bgp and nv overlay required for rd cli in NXOS
       Feature.bgp_enable
-      Feature.nv_overlay_enable
-      Feature.nv_overlay_evpn_enable
+      Feature.nv_overlay_enable      # TBD: Only req'd for n7k?
+      Feature.nv_overlay_evpn_enable # TBD: Only req'd for n7k?
       if rd == default_route_distinguisher
         state = 'no'
         rd = ''
--- a/tests/test_feature.rb
+++ b/tests/test_feature.rb
@@ -1,0 +1,75 @@
+# Copyright (c) 2016 Cisco and/or its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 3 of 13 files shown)


robert-w-gries:
Do we need to update the files from [this commit](https://github.com/cisco/cisco-network-node-utils/commit/780ab9274b4f0ef094942f21d2851962760052ac) to use your code?

chris: No. That's an old commit and a lot of those changes no longer exist.

tests/test_feature.rb
```diff
@@ -0,0 +1,75 @@
...
+  # feature test helper
+  def feature(f)
+    config("no feature #{f}") if Feature.send("#{f}_enabled?")
+    refute_show_match(pattern: /^feature #{f}$/,
+                      msg:     "Feature #{f} should not be enabled")
+
+    Feature.send("#{f}_enable")
+    # Some features (BGP on n5k!) are slow starters...
+    unless Feature.send("#{f}_enabled?")
+      sleep 1
```
robert-w-gries, line 67: Is 1 second sufficient time for features with slow startups?

tests/test_feature.rb
```diff
@@ -0,0 +1,75 @@
...
+    feature('bgp')
+  end
+
+  def test_fabric
+    feature('fabric')
+  end
+
+  def test_fabric_forwarding
+    skip('This feature is only supported on 7.0(3)I2 images') unless
+      node.os_version[/7.0\(3\)I2\(/]
```
robert-w-gries, line 86: Nit: `%r{7.0(3)I2}` seems a bit cleaner

tests/test_feature.rb
chrisvanheuveln, line 67: In my experience it should be enough. I tested it on the slowest box I have. We can always extend it later if necessary.

tests/test_feature.rb
chrisvanheuveln, line 86: Thanks - I was trying to recall the syntax for that. I'll update.

edit: no, actually this doesn't work. You still have to escape the parenthesis.

lib/cisco_node_utils/cmd_ref/vlan.yaml
```diff
@@ -16,8 +16,7 @@ mapped_vni:
     kind: int
     config_get: 'show running vlan'
     config_get_token: ['/^vlan <vlan>$/', '/^vn-segment (\d+)$/']
-    # In the cli 'vn-segment' config is not applied untill the 'end'
-    # statement is issued.
+    # The 'vn-segment' config is not applied until the 'end' statement is processed
```
glennmatthews, line 19: Since you're touching this - is it true? Does this feature stop working if you remove the literal `; end` from the YAML?

tests/test_feature.rb
```diff
@@ -0,0 +1,75 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative 'ciscotest'
+require_relative '../lib/cisco_node_utils/feature'
+
+include Cisco
+
+# TestVrf - Minitest for Vrf node utility class
```
glennmatthews: fix please

lib/cisco_node_utils/cmd_ref/vlan.yaml
chrisvanheuveln, line 19: No, and yes. :-)

From what I see, after removing the 'end'(s), all of the positive tests seem to work (vlans get created with the correct properties); however, the negative tests fail because they don't see a RuntimeError. I add the 'end' back and they get the RuntimeError again.

chrisvanheuveln:
Retested on N9k, N3k, N5k, N7k.

lib/cisco_node_utils/vrf.rb
```diff
@@ -95,8 +95,8 @@ def route_distinguisher
     def route_distinguisher=(rd)
       # feature bgp and nv overlay required for rd cli in NXOS
       Feature.bgp_enable
-      Feature.nv_overlay_enable
-      Feature.nv_overlay_evpn_enable
+      Feature.nv_overlay_enable      # TBD: Only req'd for n7k?
+      Feature.nv_overlay_evpn_enable # TBD: Only req'd for n7k?
```
glennmatthews, line 99: Still TBD?

glennmatthews:
:+1:

lib/cisco_node_utils/vrf.rb
chrisvanheuveln, line 99: Yes.

robert-w-gries:
:+1:

PR #235, opened 2016-02-16
XR route-policy and route-target enhancements

HelenHZhang:
implement US63946
- add the following for XR
  import route-policy <policy-name>
  export route-policy <policy-name>
  import route-target <community> stitching
  export route-target <community> stitching

All mini-tests passed for XR and Nexus.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/vrf.yaml
+++ b/lib/cisco_node_utils/cmd_ref/vrf.yaml
@@ -68,22 +68,25 @@ route_distinguisher:
   default_value: ''
 
 route_policy_export:
-  _exclude: [nexus]
   default_value: ~
-  get_context:
-    - '/^vrf <vrf>$/i'
-    - '/^address-family <afi> <safi>$/'
-  get_value: '/^export route-policy (\S+)$/'
-  set_value: '<state> export route-policy <policy_name>'
+  nexus:
+    get_value: '/^export map (\S+)$/'
+    set_value: '<state> export map <policy_name>'
+  ios_xr:
+    get_value: '/^export route-policy (\S+)$/'
+    set_value: '<state> export route-policy <policy_name>'
 
 route_policy_import:
-  _exclude: [nexus]
   default_value: ~
-  get_context:
-    - '/^vrf <vrf>$/i'
-    - '/^address-family <afi> <safi>$/'
-  get_value: '/^import route-policy (\S+)$/'
-  set_value: '<state> import route-policy <policy_name>'
+  nexus:
+    get_value: '/^import map (\S+)$/'
+    set_value: '<state> import map <policy_name>'
+  ios_xr:
+    get_context:
+      - '/^vrf <vrf>$/i'
+      - '/^address-family <afi> <safi>$/'
+    get_value: '/^import route-policy (\S+)$/'
+    set_value: '<state> import route-policy <policy_name>'
 
 route_target_both_auto:
   _exclude: [ios_xr]
--- a/lib/cisco_node_utils/cmd_ref/vrf.yaml
+++ b/lib/cisco_node_utils/cmd_ref/vrf.yaml
@@ -82,9 +82,6 @@ route_policy_import:
     get_value: '/^import map (\S+)$/'
     set_value: '<state> import map <policy_name>'
   ios_xr:
-    get_context:
-      - '/^vrf <vrf>$/i'
-      - '/^address-family <afi> <safi>$/'
     get_value: '/^import route-policy (\S+)$/'
     set_value: '<state> import route-policy <policy_name>'
 
--- a/lib/cisco_node_utils/cmd_ref/vrf.yaml
+++ b/lib/cisco_node_utils/cmd_ref/vrf.yaml
@@ -64,7 +64,26 @@ route_distinguisher:
   _exclude: [ios_xr]
   get_value: '/^rd (\S+)$/'
   set_value: '<state> rd <rd>'
+  context: ['vrf context <vrf>']
   default_value: ''
+
+route_policy_export:
+  _exclude: [nexus]
+  default_value: ~
+  get_context:
+    - '/^vrf <vrf>$/i'
+    - '/^address-family <afi> <safi>$/'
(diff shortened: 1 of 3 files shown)


lib/cisco_node_utils/cmd_ref/vrf.yaml
```diff
@@ -64,8 +64,27 @@ route_distinguisher:
   _exclude: [ios_xr]
   get_value: '/^rd (\S+)$/'
   set_value: '<state> rd <rd>'
+  context: ['vrf context <vrf>']
```
glennmatthews, line 67: This shouldn't be needed - it's already inherited from the _template above.

lib/cisco_node_utils/cmd_ref/vrf.yaml
```diff
@@ -64,8 +64,27 @@ route_distinguisher:
...
   default_value: ''
 
+route_policy_export:
+  _exclude: [nexus]
```
glennmatthews: FWIW it looks like the Nexus equivalent CLI is:

```
vrf context <vrf>
 address-family <afi> <safi>
  export map <policy_name>
```

Similarly for 'route-policy import' --> 'import map'

lib/cisco_node_utils/cmd_ref/vrf.yaml
glennmatthews, line 67: Ah - I see - this is not AF specific.

We really should have separate vrf.yaml and vrf_af.yaml to make the separate groupings clear and avoid having to repeat this all over the place.

lib/cisco_node_utils/vrf_af.rb
```diff
@@ -83,6 +83,40 @@ def route_target_feature_enable
     #                      PROPERTIES                      #
     ########################################################
 
+    def route_policy_export
+      config_get('vrf', 'route_policy_export', @get_args)
+    end
+
+    def route_policy_export=(name)
+      if platform == :nexus
+        fail Cisco::UnsupportedError.new('vrf_af', route_policy_export)
```
glennmatthews: Shouldn't be needed - the config_set should automatically fail and raise this error.

tests/test_vrf.rb
```diff
@@ -169,6 +169,51 @@ def test_vrf_af_create_destroy
...
+      assert_nil(v.default_route_policy_import)
+      assert_raises(Cisco::UnsupportedError) { v.route_policy_export = 'abc' }
+      assert_raises(Cisco::UnsupportedError) { v.route_policy_import = 'abc' }
+      v.destroy
+      return
+    end
+    assert_equal(nil, v.default_route_policy_import,
+                 "Test1.1 : #{af} : route_policy_import")
+    assert_equal(nil, v.default_route_policy_export,
+                 "Test1.2 : #{af} : route_policy_export")
```
glennmatthews: These should be assert_nil(...) instead of assert_equal(nil, ...)

glennmatthews:
Please update CHANGELOG and fix rubocop failure.

tests/test_vrf.rb
HelenHZhang: fixed it.

lib/cisco_node_utils/vrf_af.rb
HelenHZhang: fixed it.

lib/cisco_node_utils/cmd_ref/vrf.yaml
chburket, line 67: @glennmatthews 
While I agree that splitting the yaml would be a good idea, can we create a story for it and put it in the backlog?  I'd rather not make a change like that in the grpc branch, which will just make future merges more difficult.

lib/cisco_node_utils/cmd_ref/vrf.yaml
HelenHZhang: fixed it

lib/cisco_node_utils/cmd_ref/vrf.yaml
glennmatthews, line 67: I guess, yeah - it would be better to make that refactoring change in develop and inherit it up to feature/grpc.

HelenHZhang:
resolved all comments in the review, and pushed the new diff out. Please re-review it. Thanks.

lib/cisco_node_utils/cmd_ref/vrf.yaml
```diff
@@ -64,8 +64,30 @@ route_distinguisher:
...
+    get_value: '/^export route-policy (\S+)$/'
+    set_value: '<state> export route-policy <policy_name>'
+
+route_policy_import:
+  default_value: ~
+  nexus:
+    get_value: '/^import map (\S+)$/'
+    set_value: '<state> import map <policy_name>'
+  ios_xr:
+    get_context:
```
glennmatthews: is this needed? seems redundant with the template.

lib/cisco_node_utils/vrf_af.rb
```diff
@@ -142,6 +182,21 @@ def default_route_target_export_evpn
     end
 
     # --------------------------
+    def route_target_export_stitching
+      cmds = config_get('vrf', 'route_target_export_stitching', @get_args)
+      cmds.nil? ? nil : cmds.sort
+    end
```
glennmatthews, line 188: It's not clear to me from the code here or from the tests below how this property is structured or how it is used. Is it a list of communities that are stitching-enabled? Can you provide some examples?

How does this property interact/coexist with the `route_target_export` property?

lib/cisco_node_utils/cmd_ref/vrf.yaml
HelenHZhang: you are correct, I removed these 3 lines.

lib/cisco_node_utils/vrf_af.rb
glennmatthews, line 188: Discussed with @chrisvanheuveln and I understand the design for this property and the route_target_export_evpn property better now. This should be fine.

PR #236, opened 2016-02-16
Feature/stp global

saichint:
# Since stp is has range based attributes, the design was discussed thoroughly with Mike, Chris, Deepak and Sanjay. It would be better if Mike or Chris can do this code review since they have the most background related to this design.

stp design assumptions:
There are lot of range based commands in spanning-tree and these ranges dynamically change based on the input. 
To simplify this, we assume:

All the multiple ranges given in the manifest will not overlap each other for any given attribute. If they do, it will result in failure. This will be taken care in the puppet-module.
For ex:

vlan_hello_time => [['5,7,13,15-42', '10'], ['83-92,100-230', '5']] -----> correct

vlan_hello_time => [['5,7,13,15-42', '10'], ['6-100', '5']] -----> not correct

The ranges which are not given in the manifest will NOT be set to default. (In future, a new attribute "purge" will be added which can be used to alter this behavior)

For ex:
vlan_hello_time => [['2,7,13,15-42', '10'], ['3,10-11,100-230', '5']]
In this case, the vlan_hello_time for [1,4-6,8-9,12,14,43-99,231-3967] will be left as before.
So if the user needs a particular range to go to default, set that range to 'default' explicitly.

The minitest results are attched below:
There are some nxos issues on n9k and n6k.

On n6k, the vlan hello-time, the 'no' cmd cannot be used due to an issue so a workaround is put in place.
On n9k and n6k, "mst root priority" is not being shown properly in the running config unless "mst designated priority" is set first. There is no real dependency between these attributes but this is due to a bug. The work around is to set "mst designated priority" first and this was done in the test_stp_global.rb
All the bugs needed are filed on n9k and n6k already.

Bridge_domain related attributes are specific to n7k only so those tests will be skipped on n9k and n6k.

Test results:
[saichint@sjc-ads-6603 /ws/saichint-sjc/puppet_agent/cisco-network-node-utils/tests] $ ruby test_stp_global.rb 10.122.197.117 admin admin  
Run options: --seed 22797                                                              
# Running:

Node under test:
- name  - agent-lab13-nx
- type  - N9K-NXOSV  
  ...................................

Finished in 309.199835s, 0.1132 runs/s, 0.3364 assertions/s.

35 runs, 104 assertions, 0 failures, 0 errors, 0 skips
[saichint@sjc-ads-6603 /ws/saichint-sjc/puppet_agent/cisco-network-node-utils/tests] $ ruby test_stp_global.rb 172.29.21.200 admin C1sco123
Run options: --seed 65197
# Running:

Node under test:
- name  - N6K-6
- type  - N6K-C6001-64P
  ...................................  

Finished in 134.866288s, 0.2595 runs/s, 0.7711 assertions/s.

35 runs, 107 assertions, 0 failures, 0 errors, 0 skips
[saichint@sjc-ads-6603 /ws/saichint-sjc/puppet_agent/cisco-network-node-utils/tests] $ ruby test_stp_global.rb 172.23.145.152 admin nbv_1234
Run options: --seed 52953
# Running:

Node under test:
- name  - switch
- type  - N7K-C7009
  ...................................

Finished in 124.697445s, 0.2807 runs/s, 0.8821 assertions/s.

35 runs, 110 assertions, 0 failures, 0 errors, 0 skips


CHANGELOG.md
```diff
@@ -42,6 +42,8 @@ Changelog
   * interface_channel_group (@chrisvanheuveln)
   * interface_portchannel (@saichint)
   * portchannel_global (@saichint)
+* Spanning Tree
```
mikewiebe: This is in the wrong section.  Please move it up under the `Unreleased` section.

CHANGELOG.md
```diff
@@ -121,6 +123,8 @@ Changelog
   * `fabric_forwarding_anycast_gateway`
   * `ipv4_acl_in`, `ipv4_acl_out`, `ipv6_acl_in`, `ipv6_acl_out`
   * `ipv4_address_secondary`, `ipv4_arp_timeout`
+  * `stp_bpdufilter`, `stp_bpduguard`, `stp_cost`, `stp_guard`, `stp_link_type`, `stp_mst_cost`
```
mikewiebe: Move under the `Unreleased` section.

lib/cisco_node_utils/cmd_ref/interface.yaml
```diff
@@ -214,6 +214,77 @@ speed:
   config_set_append: "speed %s"
   default_value: "auto"
```
mikewiebe, line 249: Do these interface properties need any `_exclude:` entries?

lib/cisco_node_utils/cmd_ref/interface.yaml
```diff
@@ -214,6 +214,77 @@ speed:
...
 
+stp_bpdufilter:
+  kind: string
```
mikewiebe, line 252: Shouldn't the `kind:` field be set to `boolean` for `stp_bpdufilter:`, `stp_bpduguard:`, `stp_guard:`, `stp_port_type:`?

lib/cisco_node_utils/cmd_ref/interface.yaml
saichint, line 249: All the new stp properties per interface are available on n9k, n6k and n7k, so no_exclude is needed for them.

lib/cisco_node_utils/cmd_ref/interface.yaml
saichint, line 252: No. They all have some string enums.
please see the code on the puppet-module (which is in my fork only) for the available values.
For ex:
https://github.com/saichint/cisco-network-puppet-module/blob/feature/stp_global/lib/puppet/type/cisco_interface.rb#L521

Also I used Chris' example for channel_group for "default_value: false" (even though kind is 'int') because the entire CLIs for the attributes you mentioned here do not exist sometimes when they are in default.
https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/cmd_ref/interface_channel_group.yaml#L16

mikewiebe, line 252: Ok, that makes sense for `stp_guard`, but `stp_bpdufilter` and `stp_bpduguard` only have two values... `enable` and `disable`.  IMO, these should be treated as booleans IF there cannot be more then these two values.  

https://github.com/saichint/cisco-network-puppet-module/blob/feature/stp_global/lib/puppet/type/cisco_interface.rb#L493

Update: Ok, I see now that even `stp_bdufilter` and `stp_bpduguard` have 3 values since by default neither `enable` or `disable` are configured.

lib/cisco_node_utils/cmd_ref/interface.yaml
```diff
@@ -214,6 +214,77 @@ speed:
...
+
+stp_link_type:
+  kind: string
+  config_get_token_append: ['/^spanning-tree link-type (.*)$/']
+  config_set_append: ["spanning-tree link-type %s"]
+  default_value: 'auto'
+
+stp_mst_cost:
+  multiple:
+  config_get: "show running interface"
```
mikewiebe: Why do you need to specify a special `config_get:` for this property and others below instead of using the `_template:` `config_get:`?

tests/test_stp_global.rb
```diff
@@ -0,0 +1,533 @@
+# Copyright (c) 2014-2016 Cisco and/or its affiliates.
```
mikewiebe: Change `2014-2016` to just `2016`

tests/test_stp_global.rb
```diff
@@ -0,0 +1,533 @@
...
+# limitations under the License.
+
+require_relative 'ciscotest'
+require_relative '../lib/cisco_node_utils/stp_global'
+
+# TestX__CLASS_NAME__X - Minitest for X__CLASS_NAME__X node utility class
+class TestStpGlobal < CiscoTestCase
+  # TESTS
+
+  DEFAULT_NAME = 'default'
```
mikewiebe: This does not add any real value IMO.  Just use the string `default` directly and remove this constant.

tests/test_stp_global.rb
```diff
@@ -0,0 +1,533 @@
...
+
+  def n9k_platform?
+    /N(3|9)/ =~ node.product_id
+  end
+
+  def n6k_platform?
+    /N(5|6)/ =~ node.product_id
+  end
+
+  def create_stp_global(name=DEFAULT_NAME)
```
mikewiebe: I would remove this method as it just obfuscates things.  Just call `StpGlobal.new` instead of this method.

tests/test_stp_global.rb
```diff
@@ -0,0 +1,533 @@
...
+    StpGlobal.new(name)
+  end
+
+  def test_bd_forward_time_change
+    skip('Platform does not support this property') if n6k_platform? ||
+                                                       n9k_platform?
+    @global = create_stp_global
```
mikewiebe: Why are you using an instance variable here?  If you are not using @global outside of the testcase method you can just use a regular local scope variable.

tests/test_stp_global.rb
```diff
@@ -0,0 +1,533 @@
...
+    config 'system bridge-domain all'
+    bdft = [['2-4,6,8-12', '4'], %w(14 30)]
```
mikewiebe: Perhaps `[%w(2-4,6,8-12 4), %w(14 30)]` instead?  Just want to make sure it's consistent throughout the code.

tests/test_stp_global.rb
```diff
@@ -0,0 +1,533 @@
...
+    @global.bd_forward_time = bdft
+    assert_equal(bdft, @global.bd_forward_time)
+    bdft = [['2-42', '26'], ['83-92,1000-2300', '20'],
+            ['3000-3960', '16']]
+    @global.bd_forward_time = bdft
+    assert_equal(bdft, @global.bd_forward_time)
+    @global.bd_forward_time = @global.default_bd_forward_time
```
mikewiebe: Please add an additional default check at the beginning of the test case right after you call `StpGlobal.new`.  Do this for all applicable properties.

tests/test_stp_global.rb
```diff
@@ -0,0 +1,533 @@
...
+  def n6k_platform?
+    /N(5|6)/ =~ node.product_id
+  end
+
+  def create_stp_global(name=DEFAULT_NAME)
+    StpGlobal.new(name)
+  end
+
+  def test_bd_forward_time_change
+    skip('Platform does not support this property') if n6k_platform? ||
```
mikewiebe: **UPDATE:** Hold off on this comment for now.  We are merging code from `feature/grpc` into develop today/tonight which will render this invalid.  Our new process will be to explicitly do the following for platforms in the `_exclude` yaml entry.
- Test that calls to `config_get` and `config_set` result in a `Cisco::UnsupportedError` being raised.
- Test the calls to `config_get_default` for the property return `nil`.
- Example: https://github.com/cisco/cisco-network-node-utils/blob/feature/grpc/tests/test_router_bgp.rb#L471

---

We are moving away from using an `explicit` skip in the test case and instead allowing the `_exclude` to be the single source of truth.  Please make sure your testcases don't use an explicit `skip` and also make sure to do the following.
- Test each property like you normally would.  If the property is not supported on a platform, the following will automatically be generated.
  - https://github.com/cisco/cisco-network-node-utils/blob/develop/tests/basetest.rb#L97

```
Node under test:
  - name  - n6k-92
  - type  - N6K-C6001-64P
  - image - bootflash:///n6000-uk9-kickstart.7.3.0.N1.1.bin

TestStpGlobal#test_bd_priorities_change = 4.08 s = S

Finished in 4.085989s, 0.2447 runs/s, 0.0000 assertions/s.

  1) Skipped:
TestStpGlobal#test_bd_priorities_change [tests/test_stp_global.rb:104]:
Platform does not support this property

1 runs, 0 assertions, 0 failures, 0 errors, 1 skips
```
- Make sure to test each property that has an exclude to make sure it get's coverage, otherwise a test case might get skipped early before all of the properties are tested.

tests/test_stp_global.rb
```diff
@@ -0,0 +1,533 @@
...
+    bdft = [['2-42', '6'], ['83-92,1000-2300', '29'],
+            ['3000-3960', '13']]
+    @global.bd_max_age = bdft
+    assert_equal(bdft, @global.bd_max_age)
+    @global.bd_max_age = @global.default_bd_max_age
+    assert_equal(@global.default_bd_max_age,
+                 @global.bd_max_age)
+  end
+
+  def test_bd_priorities_change
```
mikewiebe, line 115: I just tried to run this test case on my `n7k` but I get the following error.

```
Node under test:
  - name  - n7k-j
  - type  - N7K-C7009
  - image - bootflash:///n7000-s2-dk9.7.3.0.D1.0.202.bin

TestStpGlobal#test_bd_priorities_change = 4.85 s = E

Finished in 4.851927s, 0.2061 runs/s, 0.0000 assertions/s.

  1) Error:
TestStpGlobal#test_bd_priorities_change:
Cisco::CliError: CliError: 'no spanning-tree bridge-domain 5,7,13,15-3967 priority ' rejected with message:
'invalid bridge-domain'
    /nobackup/mwiebe/REPOS_NODE_UTILS/sai/cisco-network-node-utils/lib/cisco_node_utils/node.rb:266:in `rescue in config'
    /nobackup/mwiebe/REPOS_NODE_UTILS/sai/cisco-network-node-utils/lib/cisco_node_utils/node.rb:263:in `config'
    /nobackup/mwiebe/REPOS_NODE_UTILS/sai/cisco-network-node-utils/lib/cisco_node_utils/node.rb:161:in `config_set'
    /nobackup/mwiebe/REPOS_NODE_UTILS/sai/cisco-network-node-utils/lib/cisco_node_utils/node_util.rb:58:in `config_set'
    /nobackup/mwiebe/REPOS_NODE_UTILS/sai/cisco-network-node-utils/lib/cisco_node_utils/stp_global.rb:171:in `bd_priority='
    tests/test_stp_global.rb:109:in `test_bd_priorities_change'

1 runs, 0 assertions, 0 failures, 1 errors, 0 skips
```

lib/cisco_node_utils/node.rb
```diff
@@ -432,4 +432,61 @@ def find_subconfig(body, regex_query)
     subconfig.join("\n")
   end
   module_function :find_subconfig
+
```
mikewiebe: Shared helper methods like these belong in `cisco_cmd_utils.rb`, not `node.rb`.

https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/cisco_cmn_utils.rb

tests/test_stp_global.rb
```diff
@@ -0,0 +1,533 @@
...
+    @global.vlan_designated_priority = @global.default_vlan_designated_priority
+    assert_equal(@global.default_vlan_priority,
+                 @global.vlan_priority)
+    assert_equal(@global.default_vlan_root_priority,
+                 @global.vlan_root_priority)
+    assert_equal(@global.default_vlan_designated_priority,
+                 @global.vlan_designated_priority)
+  end
+
+  def test_interface_stp_bpdufilter_change
```
mikewiebe, line 432: We have found for interface testing that it's best to make sure the interface is starting in a clean state.  You can use the following design pattern for this.

https://github.com/cisco/cisco-network-node-utils/blob/dae5fe60e6a72db50ab14dc79242f0f2b72e3c1e/tests/test_interface_channel_group.rb#L20

This will also allow you to ensure the default value is set for each property before you set a non-default value.

lib/cisco_node_utils/interface.rb
```diff
@@ -541,6 +550,234 @@ def default_shutdown
...
+      end
+      config_set('interface',
+                 'stp_bpduguard', @name, state, val)
+    end
+
+    def default_stp_bpduguard
+      config_get_default('interface', 'stp_bpduguard')
+    end
+
+    def stp_cost
```
mikewiebe, line 621: Small simplification.

```
def stp_cost
  cost = config_get('interface', 'stp_cost', @name)
  cost == 'auto' ? cost : cost.to_i
end
```

lib/cisco_node_utils/interface.rb
```diff
@@ -541,6 +550,234 @@ def default_shutdown
...
+    #        spanning-tree mst 4000-4020 port-priority 160
+    # array: [['0,2-4,6,8-12', '64'], ['4000-4020', '160']]
+    #
+    def stp_mst_port_priority
+      config_get('interface', 'stp_mst_port_priority', @name)
+    end
+
+    # need to reset range which is not being set
+    def stp_mst_port_priority=(list)
+      config_set('interface', 'stp_mst_port_priority',
```
mikewiebe, line 709: Reminder to _not_ set unspecified range values to default.  We can do this later when we have a _purge_ knob.

**Update:** Since you already have the _purge_ knob keep the support, but the default behavior should be `purge == false`.

lib/cisco_node_utils/interface.rb
```diff
@@ -541,6 +550,234 @@ def default_shutdown
...
+                 @name, 'no', Cisco.get_reset_range('0-4094', list), '')
+      list.each do |inst, pri|
```
mikewiebe: Couple of thoughts here.
- This same logic is used in several places.  Can you simplify and create a local helper method?
- The names `inst` and `pri` are not very intuitive to me.  I think `range` and `property_value` or something along those lines would be clearer.

lib/cisco_node_utils/cmd_ref/interface.yaml
saichint: For range based attributes, if we do "show running interface all" we get results for the ranges which are not set as well and this messes up the comparison. So use only "show running config interface" for them.

tests/test_stp_global.rb
saichint, line 115: Took care of this as well. The issue is that when we have a mix of vlans and bd, the reset must be done only on the bds and not on the vlans. So added a new method to get the bd range exclusive of the vlans and resetting them if necessary.

lib/cisco_node_utils/interface.rb
saichint, line 621: done

tests/test_stp_global.rb
saichint, line 432: done

mikewiebe:
@saichint The grpc code has now been merged into develop.  You will need to update your YAML entries to conform to the new standard.
- https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/cmd_ref/README_YAML.md
- https://github.com/cisco/cisco-network-node-utils/blob/develop/CHANGELOG.md#changed

tests/test_stp_global.rb
```diff
@@ -0,0 +1,525 @@
...
+  end
+
+  def n6k_platform?
+    /N(5|6)/ =~ node.product_id
+  end
+
+  def test_bd_forward_time_change
+    skip('Platform does not support this property') if n6k_platform? ||
+                                                       n9k_platform?
+    @global = StpGlobal.new('default')
```
mikewiebe: Why are you using an instance variable here? If you are not using @global outside of the testcase method you can just use a regular local scope variable.

tests/test_stp_global.rb
```diff
@@ -0,0 +1,525 @@
...
+  def n9k_platform?
+    /N(3|9)/ =~ node.product_id
+  end
+
+  def n6k_platform?
+    /N(5|6)/ =~ node.product_id
+  end
+
+  def test_bd_forward_time_change
+    skip('Platform does not support this property') if n6k_platform? ||
```
mikewiebe: Now that grpc is merged into develop let's remove the explicit skips.

 Our new process will be to explicitly do the following for platforms in the _exclude yaml entry.
- Test that calls to config_get and config_set result in a Cisco::UnsupportedError being raised.
- Test the calls to config_get_default for the property return nil.
- Example: https://github.com/cisco/cisco-network-node-utils/blob/feature/grpc/tests/test_router_bgp.rb#L471

mikewiebe:
:+1:

glennmatthews:
:-1: This probably just broke interface support on IOS XR.

mikewiebe:
:sob:  Argh.  I can take care of this, but need a quick crash course on how to validate on XR!

PR #237, opened 2016-02-22
Added EXR support for syslog_server

jonnytdevops:


tests/test_syslog_server.rb
```diff
@@ -34,9 +34,15 @@ def teardown
 
   def no_syslogserver
     # Turn the feature off for a clean test.
-    config('no logging server 1.2.3.4',
-           'no logging server 2003::2',
-           'no vrf context red')
+    if platform == :iox_xr
```
glennmatthews: s/iox/ios/g

jonnytdevops:
@glennmatthews code updated, thanks!

jonnytdevops:
@glennmatthews Code updated again with better checks for level in the mini tests. Thanks

lib/cisco_node_utils/syslog_server.rb
```diff
@@ -42,12 +52,20 @@ def initialize(name,
...
+        begin
+          IPAddr.new(id)
+        rescue
+          next
+        end
+
         level = config_get('syslog_server', 'level', id)
+        level = LEVEL_TO_NUM[level] if platform == :ios_xr
 
         vrf = config_get('syslog_server', 'vrf', id)
```
glennmatthews, line 70: Looks like this method needs to be redesigned for XR:

```
hash[id] = SyslogServer.new(id, level, vrf, false)
```

You can have the same server (id) configured in multiple VRFs in XR:

```
logging 1.2.3.4 vrf red severity info port default
logging 1.2.3.4 vrf default severity info port default
```

but you're trying to store both instances under the same id (1.2.3.4) in this method.

tests/test_syslog_server.rb
```diff
@@ -34,9 +34,15 @@ def teardown
 
   def no_syslogserver
     # Turn the feature off for a clean test.
-    config('no logging server 1.2.3.4',
-           'no logging server 2003::2',
-           'no vrf context red')
+    if platform == :ios_xr
+      config('no logging 1.2.3.4',
+             'no logging 2003::2',
```
glennmatthews, line 39: I get some failures in the minitest because these commands will remove 'logging 1.2.3.4 (vrf default)' but they do not remove 'logging 1.2.3.4 vrf red'. Because of the vrf-blindness of the key (see above comment) the lingering VRF logging entries cause failures:

```
Node under test:
  - name  - ios
  - type  - R-IOSXRV9000-CH
.F.FF

Finished in 36.918642s, 0.1354 runs/s, 1.1376 assertions/s.

  1) Failure:
TestSyslogServer#test_create_destroy_single_vrf_ipv6 [tests/test_syslog_server.rb:134]:
Expected {"2003::2"=>#<Cisco::SyslogServer:0x007f1666f6ec48 @name="2003::2", @level=5, @vrf="red">} to not include "2003::2".


  2) Failure:
TestSyslogServer#test_create_destroy_single_ipv6 [tests/test_syslog_server.rb:65]:
Expected {"2003::2"=>#<Cisco::SyslogServer:0x007f1665e27048 @name="2003::2", @level=5, @vrf="red">} to not include "2003::2".


  3) Failure:
TestSyslogServer#test_create_destroy_single_vrf_ipv4 [tests/test_syslog_server.rb:114]:
Expected {"1.2.3.4"=>#<Cisco::SyslogServer:0x007f16661e1378 @name="1.2.3.4", @level=4, @vrf="red">, "2003::2"=>#<Cisco::SyslogServer:0x007f16661ecae8 @name="2003::2", @level=5, @vrf="red">} to not include "1.2.3.4".

5 runs, 42 assertions, 3 failures, 0 errors, 0 skips
```

lib/cisco_node_utils/syslog_server.rb
jonnytdevops, line 70: @glennmatthews this is incompatible with netdev_stdlib unless using composite namevars, which I don't believe we are supporting without further design discussion. I believe this is the same issue with some other nexus providers as well (although syslog_server isn't one of them)

jonnytpuppet: @glennmatthews I do note we have this in another provider, but is it the correct solutions without using composite namevars??

https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/dns_domain.rb#L54

glennmatthews, line 70: As you pointed me to, dnsdomain.rb works around this by having the getter function take an optional VRF as parameter and only returns the matches in that VRF (https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/dns_domain.rb#L40). Can we do something similar with this function?

jonnytdevops, line 70: @glennmatthews Sure, however what would the design be in relation to the pre-existing vrf parameter? Would it be a simple "If vrf is specified, only return the hosts that are part of that vrf"?

So basically, add the following line to line 71:

```
next if !vrf_arg.nil? && vrf_arg != vrf
```

Thanks

glennmatthews, line 70: Sounds about right but not sure that's correct for the default vrf (which I think is what is implied if vrf_arg.nil?) - for that case you should only return non-vrf cases. Also, on XR it looks like the 'vrf default' NVGENs so you may have to have slightly different behavior for XR vs NX.

```
logging 1.2.3.4
logging 2.3.4.5 vrf red
```

becomes (on XR)

```
logging 1.2.3.4 vrf default severity info port default
logging 2.3.4.5 vrf red severity info port default
```

jonnytdevops, line 70: I am confused about this. If we only return non-vrf'ed hosts for vrf_arg=nil, then we are essentially unsupporting the native vrf property of the type (This is the key different between this type, and the dns_domain, which I don't believe supported vrf's at all).

My line 71 idea was basically enabling or disabling vrf filtering (i.e. vrf_arg=nil would act just like the provider currently does)

glennmatthews, line 70: Ah, I see. My apologies for not looking closer at the netdev types. So, do I understand correctly that netdev_stdlib expects that:

1) the id (IP or name) is a unique identifier
2) the VRF is a property of a server but does not distinguish between servers

If so, then I think the correct thing to do in this code to yield the expected behavior would be to:
a) have it log a warning (or raise an error) if in iterating over the servers it finds the same server in multiple VRFs
b) in create(), remove any existing server entry in a different VRF before creating the entry in the given VRF.

Does this sound more reasonable?

jonnytdevops, line 70: Correct on 1) and 2)

Yes, I should be able to implement a) and b). 

I'll work on this tomorrow, thanks!

glennmatthews:
superseded by #239

PR #240, opened 2016-02-23
Suppress and fix warnings

glennmatthews:
When running 'rake test', Ruby interpreter warnings are enabled (this can also be manually done by running 'ruby -w tests/test_name_goes_here...'). Unfortunately this results in a lot of noise from warnings in the core grpc gem. Use a bit of Ruby magic to suppress warnings when calling out to the grpc gem.

Once the grpc noise is suppressed, there are two valid Ruby warnings:

1) redefinition of `seqno` getter method in ace.rb (we define it as an attr_reader then follow up by declaring it as a separate function)
2) circular dependency between bgp.rb and bgp_af.rb - this was added when bgp was relying on bgp_af's feature logic; now that that has been moved into feature.rb, the dependency is no longer needed.

### DIFF
--- a/lib/.rubocop.yml
+++ b/lib/.rubocop.yml
@@ -3,7 +3,7 @@ inherit_from: ../.rubocop.yml
 # Baseline code complexity metrics for the lib/ subdirectory:
 
 Metrics/AbcSize:
-  Max: 48
+  Max: 49
 
 Metrics/CyclomaticComplexity:
   Max: 23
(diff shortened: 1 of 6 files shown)


chburket:
I assume this will not suppress valid warnings, right?  Otherwise :+1:

glennmatthews:
The goal is to suppress warnings (valid or otherwise) in code that we don't own (the grpc gem) so we can run with warnings enabled and only see the warnings that we actually have some control over.

mikewiebe:
:+1:

PR #241, opened 2016-02-24
Added code to exclude syslog_setting for IOS XR

jonnytdevops:


glennmatthews:
As mentioned in email - please also update the test_syslog_settings test as well. Since there's only one test case, it should be easy enough to add - add an 'if platform==:ios_xr' check in which you perform negative testing  (make sure `timestamp` and `default_timestamp` are nil and `timestamp=` raises an UnsupportedError). See https://github.com/cisco/cisco-network-node-utils/blob/develop/tests/test_router_bgp.rb#L441 for an example.

jonnytdevops:
@glennmatthews code updated

glennmatthews:
Interesting - the XR test actually uncovered a bug in the baseline code!

```
  1) Failure:
TestSyslogSettings#test_syslogsettings_create [tests/test_syslog_settings.rb:49]:
[Cisco::UnsupportedError] exception expected, not
Class: <TypeError>
Message: <"TypeError">
---Backtrace---
lib/cisco_node_utils/syslog_settings.rb:51:in `timestamp='
```

Looks like the below fix is needed:

``` diff
diff --git a/lib/cisco_node_utils/syslog_settings.rb b/lib/cisco_node_utils/syslog_settings.rb
index cf38e7d..7c0ddda 100644
--- a/lib/cisco_node_utils/syslog_settings.rb
+++ b/lib/cisco_node_utils/syslog_settings.rb
@@ -49,7 +49,7 @@ module Cisco
     def timestamp=(val)
       fail TypeError unless val.is_a?(String)
       fail TypeError \
-        unless %w(seconds milliseconds).include?(timestamp)
+        unless %w(seconds milliseconds).include?(val)

       # There is no unset version as timestamp has a default value
       config_set('syslog_settings',
```

With this change, the test passes on XR - :+1: from me.

jonnytdevops:
@glennmatthews Code updated. That's so strange how that hasn't been uncovered before.

tests/test_syslog_settings.rb
```diff
@@ -41,14 +43,22 @@ def no_syslogsettings
...
+    else
+      assert_includes(Cisco::SyslogSettings.syslogsettings, 'default')
+      assert_equal(Cisco::SyslogSettings.syslogsettings['default'],
+                   syslog_setting)
+
+      syslog_setting.timestamp = 'milliseconds'
+      assert_equal(Cisco::SyslogSettings.syslogsettings['default'].timestamp,
+                   'milliseconds')
+      assert_equal(syslog_setting.timestamp,
+                   'milliseconds')
```
chrisvanheuveln, line 61: I know this was in the original code but... the assert_equal args are swapped. Should be `assert_equal(expected, actual)`. Please fix that. :+1:  everything else.

jonnytdevops, line 61: @chrisvanheuveln Can we please get this merged and I'll open another PR to fix this this evening? I'm keen to get this merged for this feature, thanks!

PR #242, opened 2016-02-24
Skip entire test case if feature is excluded

glennmatthews:
Looking for feedback on this approach. It adds the following functionality:
- If you have a top-level `_exclude` statement in `foo.yaml`, and
- your minitest defines the class variable `@feature = 'foo'`, and
- you run your minitest against an excluded node

then the test will report a single Skipped result and exit. 

This doesn't let us do the more fine-grained negative testing that would be done for a 'partially supported' feature, but it lets us cleanly silence entire tests for features we haven't yet begun to work on.
# Example - test_router_ospf_vrf:
## IOS XR is excluded:

```
Node under test:
  - name  - ios
  - type  - R-IOSXRV9000-CH
TestRouterOspfVrf#all_skipped = 1.22 s = S

Fabulous run in 3.430676s, 0.2915 runs/s, 0.0000 assertions/s.

  1) Skipped:
TestRouterOspfVrf#all_skipped [/home/glmatthe/cisco-network-node-utils/tests/ciscotest.rb:55]:
Skipping TestRouterOspfVrf; feature 'ospf' is unsupported on this node

1 runs, 0 assertions, 0 failures, 0 errors, 1 skips
```
## NX-OS is not excluded and runs in full:

```
Node under test:
  - name  - n9k-108
  - type  - N9K-C9396PX
TestRouterOspfVrf#test_routerospfvrf_auto_cost = 7.63 s = .
TestRouterOspfVrf#test_routerospfvrf_router_id_multiple_vrf = 11.30 s = .
TestRouterOspfVrf#test_routerospfvrf_collection_not_empty_valid = 11.84 s = .
TestRouterOspfVrf#test_routerospfvrf_get_parent_name = 3.72 s = .
TestRouterOspfVrf#test_routerospfvrf_timer_throttle_spf_multiple_vrf = 11.09 s = .
TestRouterOspfVrf#test_routerospfvrf_router_id = 7.41 s = .
TestRouterOspfVrf#test_routerospfvrf_log_adjacency_multiple_vrf = 11.31 s = .
TestRouterOspfVrf#test_routerospfvrf_timer_throttle_lsa = 9.60 s = .
TestRouterOspfVrf#test_routerospfvrf_auto_cost_multiple_vrf = 6.07 s = .
TestRouterOspfVrf#test_routerospfvrf_collection_router_multi_vrfs = 11.89 s = .
TestRouterOspfVrf#test_routerospfvrf_log_adjacency_changes = 6.55 s = .
[...]

PR #247, opened 2016-02-25
Adding vPC+ params

dcheriancisco:
Added 3 properties to vpc_domain to enable vPC+
1. fabricpath_emulated_switch_id
2. fabricpath_multicast_load_balance
3. port_channel_limit

The params 2 and 3 are needed only on Nexus 7k (modular platforms) and is supported only on that. The first param should be supported on Nexus 5k/6k/7k . But currently 5k and 6k have a bug which puts out a prompt that cannot be bypassed. So until that bug is fixed, it will be supported only on Nexus 7k.

The changes to the test_vpc.rb file are 
1. to replace the @vpc instance variable with a local variable 'vpc'
2. skip based on platform != nexus

The changes in vpc.yaml takes care of:
1. excluding ios_xr platform
2. new format of YAML


tests/test_vpc.rb
```diff
@@ -34,20 +35,12 @@ def no_feature_vpc
     config('terminal dont-ask ; no feature vpc')
```
chrisvanheuveln: Nit: This is okay but in the future I'd like to see us avoid using config() when possible. In this case you can achieve the same result with something like this:

```
domain = Vpc.domains
domain.destroy unless domain.empty?
```

tests/test_vpc.rb
```diff
@@ -56,247 +49,251 @@ def test_vpc_create
 
   def test_vpc_destroy
     # create and test again
-    @vpc = Vpc.new(100)
-    @vpc.destroy
+    vpc = Vpc.new(100)
+    vpc.destroy
     refute(Vpc.enabled, 'VPC feature should have been disabled')
```
chrisvanheuveln, line 55: Nit: I realize you're just removing the class var here but note that we should not be disabling features anymore. I'm not suggesting you change the api right now but please be aware that there are other ways to achieve a destroy.

chrisvanheuveln:
:+1:  Just nits. Please wait for an XR reviewer before merging.

lib/cisco_node_utils/cmd_ref/vpc.yaml
```diff
@@ -55,20 +57,19 @@ dual_active_exclude_interface_vlan_bridge_domain:
...
   get_value: '/^fabricpath switch\-id\s+(\d+)/'
-  set_value: "<state> fabricpath switch-id <swid>"
+  set_value: '<state> fabricpath switch-id <swid>'
   default_value: false
 
-#END vPC+ CLIs
+fabricpath_multicast_load_balance: 
+  _exclude: [N3k, N5k, N6k, N9k]
+  kind: boolean
+  get_value: '/^fabricpath multicast load\-balance/'
```
glennmatthews: Is the backslash escape actually needed/useful?

Should this regexp be anchored at the end ($/) too?

lib/cisco_node_utils/vpc.rb
```diff
@@ -62,6 +66,14 @@ def enable
       config_set('vpc', 'feature', state: '')
     end
 
+    def fabricpath_feature
```
glennmatthews: Since you're only using these helper functions once below, my personal preference would be not to add these and just to call the FabricpathGlobal methods directly:

``` ruby
      FabricpathGlobal.fabricpath_feature_set(:enabled) if
        switch_id && :enabled != FabricpathGlobal.fabricpath_feature
```

tests/test_vpc.rb
```diff
@@ -21,12 +21,13 @@
 # TestVpc - Minitest for Vpc node utility class
 class TestVpc < CiscoTestCase
   def setup
+    skip("Test not supported on #{node.product_id}") if platform != :nexus
```
glennmatthews: Better approach (now that #242 is merged) is to use the `@skip_unless_supported` property:

``` diff
 class TestVpc < CiscoTestCase
+  @skip_unless_supported = 'vpc'
   def setup
```

tests/test_vpc.rb
```diff
@@ -21,12 +21,13 @@
...
     super
     no_feature_vpc
   end
 
   def teardown
-    no_feature_vpc
+    no_feature_vpc if platform == :nexus
```
glennmatthews: this change won't be needed if you use `@skip_unless_supported` as suggested above

lib/cisco_node_utils/cmd_ref/vpc.yaml
dcheriancisco: Since in standard regex, '-' is a special character, I tend to escape it. 
For the $ anchor, sometimes there are trailing spaces in these CLI, so instead of writing '\s*$' at the end, I leave them out. Is it really necessary to anchor?

lib/cisco_node_utils/vpc.rb
dcheriancisco: ok sure, I will take those methods out

tests/test_vpc.rb
dcheriancisco: ok will do this

tests/test_vpc.rb
dcheriancisco: ok, but does the skip in setup prevent the call to teardown? I think I still saw the teardown being called

lib/cisco_node_utils/cmd_ref/vpc.yaml
glennmatthews: Anchoring it protects us against any unexpected additions to the CLI (say 'fabric-path multicast load-balance _disable_') - if there are cases where we _know_ there are trailing spaces, we can explicitly handle them in the regex, but in general this shouldn't be needed.

In Ruby regexp `-` is only significant inside a `[...]` block - so this isn't necessary, although harmless.

tests/test_vpc.rb
glennmatthews: FWIW I'm also adding the skip_unless_supported as part of #250 - so whichever of us commits first will win here ;)

tests/test_vpc.rb
glennmatthews: This is the advantage of skip_unless_supported - it prevents both setup and teardown from being run.

tests/test_vpc.rb
dcheriancisco: ok let me change it now

tests/test_vpc.rb
dcheriancisco, line 55: disabling the feature may be needed to clear the configuration easily. Also, doesn't the resource 'ensure = absent' rely on the destroy method to clear the config?

chrisvanheuveln, line 55: The problem is that we may have other providers that may need a given feature to be enabled; ie. if I did 'no feature pim' to clean up the global config it would also clear out any interface pim configs along with any other pim provider configs.

Yes, 'no feature' does make config cleanup easy but we can do the same thing in this case by calling each setter with it's default value, ie:

```
  def self.destroy
    foo=(default_foo)
    bar=(default_bar)
  end
```

I realize this approach may not work in all cases. Where it does work, we should use it. 
Again, my comment was just an fyi for this review.

lib/cisco_node_utils/vpc.rb
chrisvanheuveln: Even better, move them into the Feature class. I know, I know, beyond the scope of this story... :-)

tests/test_vpc.rb
dcheriancisco: ok cool

dcheriancisco:
Added the changes after review and re-ran minitests successfully on n5k/n6k/n7k/n9k

tests/test_vpc.rb
```diff
@@ -21,33 +21,27 @@
 # TestVpc - Minitest for Vpc node utility class
 class TestVpc < CiscoTestCase
   def setup
+    @skip_unless_supported = 'vpc'
```
glennmatthews: `@skip_unless_supported` is a class instance variable, so this is incorrect - this line belongs at the class level, not inside the `setup` method:

``` ruby
class TestVpc < CiscoTestCase
  @skip_unless_supported = 'vpc'

  def setup
    super
    no_feature_vpc
  end

  def teardown
    no_feature_vpc
    super
  end
```

dcheriancisco: done. pls merge if no other comments

PR #248, opened 2016-02-26
Refactor test_ace to handle platform differences

chrisvanheuveln:
- This test did not handle platform support differences and because it tried to do everything in a single method it was difficult to troubleshoot.
- I refactored the test so that it tests each property on its own, which allows for easy skipping when they aren't supported on a given platform.
- N5k/6k zero-pad the nvgen'd ipv6 addresses so I added a change to normalize the addrs without the padding.
- This test bypasses XR completely.
- Tested on N(3|5|6|7|9)k.

### DIFF
--- a/tests/test_ace.rb
+++ b/tests/test_ace.rb
@@ -18,251 +18,211 @@ require_relative '../lib/cisco_node_utils/ace'
 
 # TestAce - Minitest for Ace node utility class
 class TestAce < CiscoTestCase
+  @skip_unless_supported = 'acl'
+  @@pre_clean_needed = true # rubocop:disable Style/ClassVars
+
   def setup
-    # setup runs at the beginning of each test
     super
-    @acl_name_v4 = 'test-foo-v4-1'
-    @acl_name_v6 = 'test-foo-v6-1'
-    @seqno = 10
-    no_access_list_foo
+    remove_all_acls if @@pre_clean_needed
+    @@pre_clean_needed = false # rubocop:disable Style/ClassVars
   end
 
   def teardown
-    # teardown runs at the end of each test
-    no_access_list_foo
     super
+    remove_all_acls
   end
 
-  def no_access_list_foo
-    # Remove the test ACLs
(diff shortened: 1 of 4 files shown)


tests/test_ace.rb
```diff
@@ -18,251 +18,211 @@
...
   end
 
-  # TESTS
-  # rubocop:disable Metrics/MethodLength
-  def test_create_destroy_ace_one
-    attr_v4_1 = {
+  # Helper to create an ACE and return the obj. The test_hash contains
+  # only the minimum properties which can be added to or overwritten as
+  # required for each test.
+  def ace_helper(afi, props)
```
mikewiebe, line 44: _nit_ Would `ace_builder` be a better name?

tests/test_ace.rb
```diff
@@ -18,251 +18,211 @@
...
-    dst = '3.0.0.0 0.0.0.4'
-    entry = { action: action, proto: proto, src_addr: src, dst_addr: dst }
-    a.ace_set(entry)
-    assert_equal(dst, a.dst_addr)
+  def test_ttl
+    val = '3'
+    %w(ipv4 ipv6).each do |afi|
+      a = ace_helper(afi, ttl: val)
+      skip("This property has a known defect on the platform itself -\n"\
+           'CSCuy47463: access-list ttl does not nvgen') if a.ttl.nil?
```
mikewiebe, line 224: I'm fine with this for now, but in general we don't want to include DDTS numbers in our code since we have to scrub them out before the release and we might forget.

mikewiebe:
@chrisvanheuveln Few nits, but other then that :+1:

tests/test_ace.rb
chrisvanheuveln, line 44: No

tests/test_ace.rb
chrisvanheuveln, line 224: This bug is in shipping images. Why the ban on DDTS references?

PR #251, opened 2016-02-29
vdc template fixes

chrisvanheuveln:
- vdc get_context should be optional
- vdc set_context needs terminal dont-ask for prompt-less mode
- test_vdc minitest now passes 100% on N7k (the only platform currently supporting the vdc provider)

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/vdc.yaml
+++ b/lib/cisco_node_utils/cmd_ref/vdc.yaml
@@ -7,8 +7,8 @@ _exclude: [N3k, N5k, N6k, N9k]
 
 _template:
   get_command: 'show run vdc all'
-  get_context: ['/^vdc <vdc>/']
-  set_context: ['vdc <vdc>']
+  get_context: ['(?)/^vdc <vdc>/']
+  set_context: ['terminal dont-ask ; vdc <vdc>']
 
 all_vdcs:
   multiple:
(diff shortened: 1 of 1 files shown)


lib/cisco_node_utils/cmd_ref/vdc.yaml
```diff
@@ -7,8 +7,8 @@ _exclude: [N3k, N5k, N6k, N9k]
 
 _template:
   get_command: 'show run vdc all'
-  get_context: ['/^vdc <vdc>/']
-  set_context: ['vdc <vdc>']
+  get_context: ['(?)/^vdc <vdc>/']
```
glennmatthews: As I stated in the email thread - for most of the properties in the feature, this is not optional - it would be preferable to explicitly override the context for properties that it doesn't apply to, rather than making this template change.

lib/cisco_node_utils/cmd_ref/vdc.yaml
```diff
@@ -7,8 +7,8 @@ _exclude: [N3k, N5k, N6k, N9k]
...
+  set_context: ['terminal dont-ask ; vdc <vdc>']
```
glennmatthews: Does this have to be spliced with ';' to work properly, or can this be a normal multi-entry context `set_context: ['terminal dont-ask', 'vdc <vdc>']`?

chrisvanheuveln: fixed

lib/cisco_node_utils/cmd_ref/vdc.yaml
chrisvanheuveln: Agreed.

glennmatthews:
:+1:

PR #254, opened 2016-03-01
Feature/vxlan bridge domain

rkorlepa:
This is regarding the addition of bridge-domain and encapsulation node utils.
It deals with all the config commands for bridge-domain and the properties under it.
Before this bridge-domain is defined under vni node, so remove the vni.rb and related files as it not necessary and converted it to BridgeDomain and Encapsulation node utils.
Also covers the minitests for both the nodeutils


lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -0,0 +1,247 @@
...
+
+module Cisco
+  # node_utils class for bridge_domain
+  class BridgeDomain < NodeUtil
+    attr_reader :name, :bd_ids
+
+    def initialize(bds, instantiate=true)
+      @bd_ids = bds.to_s
+      fail 'bridge-domain value is empty' if @bd_ids.empty? # no empty strings
+      fail 'bridge-domain value has spaces' if @bd_ids.include?(' ') # no spaces
```
chrisvanheuveln: Why fail if spaces? Can't you just remove any spaces?

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -0,0 +1,247 @@
...
+      narray.each do |elem|
+        if elem.include?('-')
+          earray = elem.split('-')
+          earray.each do |temp_elem|
+            fail 'bridge-domain cannot be a string' unless
+                                    BridgeDomain.number?(temp_elem)
+            fail 'Invalid bridge-domain value' unless
+                                    temp_elem.to_i > 1 && temp_elem.to_i < 3968
+          end
+        else
```
chrisvanheuveln: While all of this range/type checking is nice I don't know that it's really necessary or desirable. We normally try to avoid enforcing ranges because these things can differ among platforms and releases; ie. if it's out of range the cli will reject it appropriately.

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -0,0 +1,247 @@
...
+        /(ERROR:|Warning:)/.match(result[2]['body'].to_s)
+
+      # Some test environments get result[2] as a string instead of a hash
+      fail result[2] if
+        result[2].is_a?(String) &&
+        /(ERROR:|Warning:)/.match(result[2])
+    end
+
+    def create
+      result = config_set('bridge_domain', 'create', @bd_ids, @bd_ids)
```
chrisvanheuveln: Does this work when there is existing bd configs? I seem to recall some problems changing this cli when one was already present.

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -0,0 +1,247 @@
...
+    end
+
+    def destroy
+      result = config_set('bridge_domain', 'destroy', @bd_ids, @bd_ids)
+      cli_error_check(result)
+    rescue CliError => e
+      raise "[bridge-domain #{@bd_ids}] '#{e.command}' : #{e.clierror}"
+    end
+
+    def self.allbds
```
chrisvanheuveln: Nits: Our convention has been to use camel case for these, ie. `self.all_bds` (which is also what you're using for the yaml tag). Also, we define our classes with the initialize method first, then all of the class methods, then all of the instance methods; ie. move this method up :-)

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -0,0 +1,247 @@
...
+      hash
+    end
+
+    ########################################################
+    #                      PROPERTIES                      #
+    ########################################################
+
+    # Bridge-Domain Shutdown case
+    def shutdown
+      result = config_get('bridge_domain', 'shutdown', @bd_ids)
```
chrisvanheuveln: Please recode all of your methods to use key-val instead of printf. We are deprecating the printf style yamls and don't allow any new providers to use printf.

https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/cmd_ref/README_YAML.md#key-value-wildcards

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -0,0 +1,247 @@
...
+      # Valid result is either: "active"(aka no shutdown) or "shutdown"
+      result[/DOWN/] ? true : false
+    end
+
+    def shutdown=(val)
+      no_cmd = (val) ? '' : 'no'
+      result = config_set('bridge_domain', 'shutdown', @bd_ids, no_cmd)
+      cli_error_check(result)
+    rescue CliError => e
+      raise "[bridge-domain #{@bd_ids}] '#{e.command}' : #{e.clierror}"
```
chrisvanheuveln: This is okay but consider using a helper method to do this raise instead of replicating it on every setter.

```
rescue CliError => e
  bd_raise(e)
end

def bd_raise(e)
  raise "[bridge-domain #{@bd_ids}] '#{e.command}' : #{e.clierror}"
end
```

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -0,0 +1,247 @@
...
+        result = config_set('bridge_domain', 'name', @bd_ids, '', str)
+      end
+      cli_error_check(result)
+    rescue CliError => e
+      raise "[bridge-domain #{@bd_ids}] '#{e.command}' : #{e.clierror}"
+    end
+
+    def default_name
+      sprintf('Bridge-Domain%s', @bd_ids)
+    end
```
chrisvanheuveln, line 118: What is the purpose of this method?

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -0,0 +1,247 @@
...
+      cli_error_check(result)
+    rescue CliError => e
+      raise "[bridge-domain #{@bd_ids}] '#{e.command}' : #{e.clierror}"
+    end
+
+    def default_fabric_control
+      config_get_default('bridge_domain', 'fabric_control')
+    end
+
+    # Bridge-Domain member vni assigning case
```
chrisvanheuveln, line 142: Please add some comments to this method to indicate what it needs to do.

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -0,0 +1,247 @@
...
+      return '' unless vni_list.any?
+
+      BridgeDomain.bd_list_to_string(vni_list)
+    end
+
+    def set_member_vni=(cmd, val)
+      no_cmd = (cmd) ? '' : 'no'
+      vdc = Vdc.new('default')
+      vdc.limit_resource_module_type = 'f3' unless
+        vdc.limit_resource_module_type == 'f3'
```
chrisvanheuveln: These vdc commands shouldn't be here. The user needs to specify this in their cisco_vdc manifest.

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -0,0 +1,247 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative 'node_util'
+require_relative 'feature'
+require_relative 'vdc'
```
chrisvanheuveln: Remove the vdc require.

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -0,0 +1,247 @@
...
+
+      BridgeDomain.bd_list_to_string(vni_list)
+    end
+
+    def set_member_vni=(cmd, val)
+      no_cmd = (cmd) ? '' : 'no'
+      vdc = Vdc.new('default')
+      vdc.limit_resource_module_type = 'f3' unless
+        vdc.limit_resource_module_type == 'f3'
+      Feature.vni_enable unless Feature.vni_enabled?
```
chrisvanheuveln: You don't need the unless since vni_enable already checks that.

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
```diff
@@ -0,0 +1,44 @@
+# bridge_domain
+# bridge_domain feature is available only on n7k
+---
+_exclude: [N3k, N5k, N6k, N9k]
+
+all_bds:
+  multiple: true
+  get_command: "show bridge-domain"
```
chrisvanheuveln: Is it possible to get the information you need from running-config alone? We have run into numerous problems when using non-show-run methods to get cli state, especially when trying to support new platforms. I know that bridge-domain is an unusually messy cli but please explore the running-config approach.

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
```diff
@@ -0,0 +1,44 @@
...
+  get_value: '/^Bridge-domain (\d+) /'
+
+create:
+  set_value: "system bridge-domain add %s ; bridge-domain %s"
```
chrisvanheuveln: As mentioned before, please use the key-val technique for all yaml entries. Also, don't use the semi-colon, use: `['system bridge-domain add %s', 'bridge-domain %s']

This raises another question: You're doing an add operation here instead of a replace. Is that going to cause idempotency issues (if the current state doesn't match the manifest state)?

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+destroy:
+  set_value: "no bridge-domain %s ; system bridge-domain remove %s"
+
+fabric_control:
+  get_command: "show bridge-domain summary"
+  get_value: '/^\w+ fabric-control [a-zA-Z0-9-_\/\s]+:\s+(\d+)/'
```
chrisvanheuveln: I think your regex could simplify to: ``/^\w+ fabric-control [\w\/\s]+:\s+(\d+)/'`

lib/cisco_node_utils/cmd_ref/encapsulation.yaml
```diff
@@ -0,0 +1,22 @@
+# encapsulation profile
```
chrisvanheuveln: 1. Just wondering if this should be more narrowly defined as `encapsulation_profile` or `encapsulation_profile_vni` since it's very specific to that use?

lib/cisco_node_utils/cmd_ref/encapsulation.yaml
```diff
@@ -0,0 +1,22 @@
...
+# encapsulation profile feature is available only on n7k
+---
+_exclude: [N3k, N5k, N6k, N9k]
```
chrisvanheuveln: You also need to evaluate `ios_xr` and exclude it if necessary. If you don't have access to an XR device then please ask for help from the larger team.

lib/cisco_node_utils/cmd_ref/encapsulation.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+all_encaps:
+  multiple: true
+  get_command: "show encapsulation profile"
```
chrisvanheuveln: Again, you need to have a good reason for not using the show run output to check cli state. It may be easier to get what you want from the non-show-run commands but believe me we've been burned many times on this.

lib/cisco_node_utils/encapsulation.rb
```diff
@@ -0,0 +1,100 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative 'node_util'
+require_relative 'feature'
+require_relative 'vdc'
```
chrisvanheuveln: remove vdc require

lib/cisco_node_utils/encapsulation.rb
```diff
@@ -0,0 +1,100 @@
+# Copyright (c) 2014-2015 Cisco and/or its affiliates.
```
chrisvanheuveln: update copyrights on all files

lib/cisco_node_utils/encapsulation.rb
```diff
@@ -0,0 +1,100 @@
...
+      fail result[2] if
+        result[2].is_a?(String) &&
+        /(ERROR:|Warning:)/.match(result[2])
+    end
+
+    # Enable feature and create encap instance
+    def create
+      vdc = Vdc.new('default')
+      vdc.limit_resource_module_type = 'f3' unless
+        vdc.limit_resource_module_type == 'f3'
```
chrisvanheuveln: remove vdc calls

lib/cisco_node_utils/encapsulation.rb
```diff
@@ -0,0 +1,100 @@
...
+      Feature.vni_enable unless Feature.vni_enabled?
```
chrisvanheuveln: ditto remove `unless`

lib/cisco_node_utils/cmd_ref/encapsulation.yaml
```diff
@@ -0,0 +1,22 @@
...
+  get_value: '/^encapsulation profile vni (\S+)/'
+
+create:
+  set_value: "encapsulation profile vni %s"
+
+destroy:
+  set_value: "no encapsulation profile vni %s"
+
+dot1q_map:
```
chrisvanheuveln, line 20: Will users understand what `dot1q_map` means in this case? versus `vni` or `dot1q_vni`. I'm not saying you need to change this, just asking if this is a keyword that users can relate to; in other words when you write the readme doc entry for this property will the description say 'this is the dot1q mapping between a vlan and vni' or 'this is the profile vni' or...? You should also check the yang model to see what they use for this attribute.

lib/cisco_node_utils/encapsulation.rb
```diff
@@ -0,0 +1,100 @@
...
+    # ----------
+    # PROPERTIES
+    # ----------
+
+    def dot1q_map
+      config_get('encapsulation', 'dot1q_map', @encap_name)
+    end
+
+    def set_dot1q_map=(cmd, dot1q, vni)
+      no_cmd = (cmd) ? '' : 'no'
```
chrisvanheuveln, line 70: nit: our convention has been to use `state` for the 'no' state, ie `state = state ? '' : 'no'`

tests/test_bridge_domain.rb
```diff
@@ -0,0 +1,159 @@
...
+    vlan = Vlan.new(100)
+    assert_raises(RuntimeError,
+                  'Vlan already exist did not raise RuntimeError') do
+      BridgeDomain.new(100)
+    end
+    vlan.destroy
+  end
+
+  def test_multiple_bd_create_destroy
+    create = '101-102,120'
```
chrisvanheuveln, line 74: You could also try testing out of order bds and extra whitespace:  `create = '  120,  101-102  '`

tests/test_bridge_domain.rb
```diff
@@ -0,0 +1,159 @@
...
+    bd.destroy
+  end
+
+  def test_bd_member_vni
+    bd = BridgeDomain.new(100)
+    assert_equal(bd.default_member_vni, bd.member_vni,
+                 'Error: Bridge-Domain is mapped to different vnis')
+
+    vni = '5000'
+    bd.send(:set_member_vni=, true, vni)
```
chrisvanheuveln: What's the reason for using `send` here instead of just calling `set_member_vni=(true, vni)` ?

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -0,0 +1,247 @@
...
+      hash_map = Hash[curr_bd_vni_list.zip(curr_vni_list.map)]
+      input_bds.each do |bd|
+        vni_list << hash_map[bd]
+      end
+      return '' unless vni_list.any?
+
+      BridgeDomain.bd_list_to_string(vni_list)
+    end
+
+    def set_member_vni=(cmd, val)
```
chrisvanheuveln: Why pass two args here? If you're trying to do add/remove per-vni then I'm not sure this is the right approach (I would expect this to be a replace-all) property.

Also note that you'll have issues with `=` setters when they have more than one arg (sigh, found that one out the hard way).

Single arg method:

```
def member_vni=(val)
  state = val ? '' : 'no'
```

lib/cisco_node_utils/bridge_domain.rb
glennmatthews: I've raised this objection before - I don't see the value in rescuing one type of error in order to re-raise a different error, especially as boilerplate like this. Would much prefer, if we find the information provided by CliError to be inadequate, to figure out how to capture additional information at the point where the error is raised rather than having to do this kind of thing throughout all of our various providers.

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -0,0 +1,247 @@
...
+                                    BridgeDomain.number?(elem)
+          fail 'Invalid bridge-domain value' unless
+                                    elem.to_i > 1 && elem.to_i < 3968
+        end
+      end
+
+      create if instantiate
+    end
+
+    def self.number?(string)
```
glennmatthews: I don't feel this helper method has a lot of value, to be honest - if the value is not an integer, then `to_i` will raise an appropriate error in any case.

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -0,0 +1,247 @@
...
+          list << elem.to_i
+        end
+      end
+      list.flatten
+    end
+
+    # This method will generate a batched string if a list is passed as
+    # argument
+    # Input would be as [1,2,3,4,5,10,11,12,7,30,100,31,32]
+    # output will be 1-5,10-12,7,30,100,31-32
```
glennmatthews, line 60: Is ordering significant for this property? Otherwise I would think it would be cleaner to auto-sort this to `1-5,7,10-12,30-32,100`.

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
```diff
@@ -0,0 +1,44 @@
+# bridge_domain
+# bridge_domain feature is available only on n7k
+---
+_exclude: [N3k, N5k, N6k, N9k]
```
glennmatthews: Please also add `ios_xr` to the exclude list.

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
glennmatthews: set_value only allows a single entry at present - if this is actually two config commands to set the value, as it appears to be, then this is the correct way to do it. This is the first property I've seen that actually has two set_value commands - should these actually be two separate properties?

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
glennmatthews: Can you add a comment showing the kinds of strings this needs to match?

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+fabric_control:
+  get_command: "show bridge-domain summary"
+  get_value: '/^\w+ fabric-control [a-zA-Z0-9-_\/\s]+:\s+(\d+)/'
+  set_value: "bridge-domain %s ; %s fabric-control"
+  default_value: "0" 
+
+member_vni:
+  get_command: "show running-config bridge-domain | inc 'member vni'"
+  get_value: '/^  member vni (.*)/'
```
glennmatthews, line 36: leading whitespace is probably not the right way to define this. Can you provide a get_context instead?

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
```diff
@@ -0,0 +1,44 @@
...
+  set_value: "%s vni %s ; bridge-domain %s ; %s member vni %s"
```
glennmatthews: Are all three of these really `set_value` commands? Or are some of these really just context?

lib/cisco_node_utils/encapsulation.rb
```diff
@@ -0,0 +1,100 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative 'node_util'
+require_relative 'feature'
+require_relative 'vdc'
+
+module Cisco
+  # Encapsulation - node utility class for Encapsulation config mgmt.
+  class Encapsulation < NodeUtil
```
glennmatthews, line 24: This seems like a very generic name. Is this global encapsulation properties, per-interface encapsulation, encapsulation profile definiton, or what?

tests/test_bridge_domain.rb
```diff
@@ -0,0 +1,159 @@
...
+# limitations under the License.
+
+require_relative 'ciscotest'
+require_relative '../lib/cisco_node_utils/bridge_domain'
+require_relative '../lib/cisco_node_utils/vlan'
+
+include Cisco
+
+# TestBridgeDomain - Minitest for bridge domain class.
+class TestBridgeDomain < CiscoTestCase
```
glennmatthews, line 22: Please add: `@skip_unless_supported = 'bridge_domain'` at the top of the class.

tests/test_encapsulation.rb
```diff
@@ -0,0 +1,71 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative 'ciscotest'
+require_relative '../lib/cisco_node_utils/encapsulation'
+
+# TestEncapsulation - Minitest for Encapsulation node utility class
+class TestEncapsulation < CiscoTestCase
```
glennmatthews, line 19: `@skip_unless_supported = 'encapsulation'`

tests/test_bridge_domain.rb
```diff
@@ -0,0 +1,159 @@
...
+  @@cleaned = false # rubocop:disable Style/ClassVars
+
+  def cleanup
+    BridgeDomain.allbds.each do |_bd, obj|
+      obj.destroy
+    end
+  end
```
glennmatthews, line 30: Please call `super` at the end of your `cleanup` method (same comment in other test file(s)).

lib/cisco_node_utils/bridge_domain.rb
chrisvanheuveln: I don't disagree. Do you have some ideas on how to capture the various object id's at the point of raise?

tests/test_bridge_domain.rb
chrisvanheuveln, line 30: @glennmatthews Is there a `cleanup` in the superclass? Where is that defined?

glennmatthews, line 30: Sorry, my mistake - I meant `teardown` needs to call `super`.

lib/cisco_node_utils/bridge_domain.rb
glennmatthews: Looking into it. :-)

lib/cisco_node_utils/bridge_domain.rb
rkorlepa: Done, removed the check and deleted the spaces if present.

lib/cisco_node_utils/bridge_domain.rb
rkorlepa: Yes Chris your right no need to do this, its rejected by cli and even wrote a minitest to validate it.

lib/cisco_node_utils/bridge_domain.rb
rkorlepa: Moved it above and followed the camel case convention.

lib/cisco_node_utils/bridge_domain.rb
rkorlepa: Changed all the yaml and related files to use key-value instead of printf

lib/cisco_node_utils/bridge_domain.rb
rkorlepa: Changed it with helper functions

lib/cisco_node_utils/bridge_domain.rb
rkorlepa, line 118: This is similar to the vlan name property, here even bridge-domains can be named.

lib/cisco_node_utils/bridge_domain.rb
rkorlepa, line 142: Added the comments

lib/cisco_node_utils/bridge_domain.rb
rkorlepa: Removed it as its needed to specify in manifest

lib/cisco_node_utils/bridge_domain.rb
rkorlepa: Removed.

lib/cisco_node_utils/bridge_domain.rb
rkorlepa: Yes Feature.vni_enalbe already does it so removed the unless part.

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
rkorlepa: Yes its possilbe and changed it to accommodate it. I have used "show bridge-domain" bridge-domain component is single sourced component accross nxos and its very unlikely that this syntax would be changed. Also another reason was that in show running-config there will be a huge set of configs if its scale setup so thought parsing would take a lot of time.

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
rkorlepa: Simplified the regex to /^[\S\s]+fabric-control[\S\s]+:\s+(\d+)/.
Also added the comment for the string to parse in yaml file.

lib/cisco_node_utils/cmd_ref/encapsulation.yaml
rkorlepa: Added ios_xr in exclude

lib/cisco_node_utils/cmd_ref/encapsulation.yaml
rkorlepa: Yes it can be changed as its for only mt-full vni case.

lib/cisco_node_utils/cmd_ref/encapsulation.yaml
rkorlepa: Changed it to "show running-config vni"

lib/cisco_node_utils/encapsulation.rb
rkorlepa: Removed

lib/cisco_node_utils/encapsulation.rb
rkorlepa: Updated copyrights on all the required files.

lib/cisco_node_utils/encapsulation.rb
rkorlepa: removed vdc as its defined in cisco_vdc manifest

lib/cisco_node_utils/encapsulation.rb
rkorlepa: Removed the unless as its checked in Feature.vni_enable

tests/test_bridge_domain.rb
rkorlepa, line 74: Yes added out of order testcase also

tests/test_bridge_domain.rb
rkorlepa: Changed the functionality to take only one argument as its difficult to implement 2 argument setters. So now it will use only bd.member_vni = '5000'. I had used send because if two arguments were passed to function then send was used in some test files, didn't know if this was valid statement. Thanks for the clarification.

lib/cisco_node_utils/bridge_domain.rb
rkorlepa: Yes changed it.

lib/cisco_node_utils/bridge_domain.rb
rkorlepa: Removed it as it not necessary to check these. Cli already throws an error if any failure happens. Also added a testcase to verify it.

lib/cisco_node_utils/bridge_domain.rb
rkorlepa, line 60: Yes ordering is important as this is used in member_vni getter and if the vni mapping are not in order then it would cause an issue.
Eg.
bridge-domain 100,105,108-110,150
   member vni 5000, 8000, 5008-5009, 7000, 5050

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
rkorlepa: Added

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
rkorlepa: The first one and the last one are set_value commands but the bridge-domain %s is for the context.

lib/cisco_node_utils/encapsulation.rb
rkorlepa, line 24: This is a global encap property. Can be executed only for mt-full

tests/test_bridge_domain.rb
rkorlepa, line 22: Added

tests/test_encapsulation.rb
rkorlepa, line 19: Added

tests/test_bridge_domain.rb
rkorlepa, line 30: Done fixed it

lib/cisco_node_utils/bridge_domain.rb
glennmatthews: See #266

mikewiebe:
@rkorlepa Please update your workspace to resolve the the branch conflicts and also fix the rubocop error.

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -0,0 +1,275 @@
...
+      sys_bds_array = BridgeDomain.bd_ids_to_array(system_bd_add)
+      inp_bds_array = BridgeDomain.bd_ids_to_array(@bd_ids)
+      if (inp_bds_array - sys_bds_array).any?
+        add_bds = BridgeDomain.bd_list_to_string(inp_bds_array - sys_bds_array)
+        config_set('bridge_domain', 'system_bd_add', addbd: add_bds)
+      end
+      result = config_set('bridge_domain', 'create', crbd: @bd_ids)
+      cli_error_check(result)
+    rescue CliError => e
+      bd_raise(e)
```
mikewiebe: @glennmatthews Committed a code that will alllow you to remove these `rescue` blocks that catch one kind of error and re-raise a different error.

See #266

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+create:
+  set_value: "bridge-domain <crbd>"
+
+destroy:
+  set_value: "no bridge-domain <delbd> ; system bridge-domain remove <rembd>"
+
+# Configured fabric-control Bridge-Domain/VLAN    :  100
+fabric_control:
+  get_command: "show bridge-domain summary"
```
mikewiebe, line 19: It looks like you changed some of the properties to use running-config, but not all?  Is it feasible to use the running config for all properties?

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
```diff
@@ -0,0 +1,52 @@
...
+  get_value: '/^[\S\s]+fabric-control[\S\s]+:\s+(\d+)/'
```
mikewiebe: Perhaps simplify a bit more?

`'/^Configured fabric-control.*:\s+(\d+)/'`

lib/cisco_node_utils/cmd_ref/encapsulation.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+create:
+  set_value: "encapsulation profile vni <profile>"
+
+destroy:
+  set_value: "no encapsulation profile vni <profile>"
+
+dot1q_map:
+  get_command: "show encapsulation profile"
+  get_value: '/^[a-zA-Z\s]+ vni <profile>\s+dot1q ([0-9,-]+) vni ([0-9,-]+)/'
```
mikewiebe: Would be good to have sample output that this is matching as well.  I think that this could be further simplified.

mikewiebe:
Please add an entry to the CHANGELOG.

https://github.com/cisco/cisco-network-node-utils/blob/develop/CHANGELOG.md#unreleased

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
rkorlepa: Changed it

lib/cisco_node_utils/cmd_ref/encapsulation.yaml
rkorlepa: Added an example

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
rkorlepa, line 19: Its difficult to use show run for properties as bridge domain cli is a bit complicated and cannot be used to extract the properties. Have added the example of show bridge-domain cli and explained the reason as well in the yaml file

rkorlepa:
Done with all the review comments.

mikewiebe:
:+1:

mikewiebe:
@glennmatthews Do you have any more comments on this PR?

glennmatthews:
:+1: to my limited understanding of this feature.

PR #255, opened 2016-03-01
Fix casing issue for anycast gateway mac

mikewiebe:
**Issues resolved:**
- On fretta(n8k) the anycast gateway mac is displayed in uppercase instead of lowercase like it is on n9k.  This fix normalizes the getter to return the mac downcased.  This change will also be made to puppet in a subsequent commit to downcase the manifest data.
- Removes an incorrect call to Utils.zero_pad_macaddr() since the mac is already zero-padded.

N9K and N8K tests: :white_check_mark:

### DIFF
--- a/lib/cisco_node_utils/overlay_global.rb
+++ b/lib/cisco_node_utils/overlay_global.rb
@@ -127,8 +127,7 @@ module Cisco
     def anycast_gateway_mac
       return nil unless Feature.nv_overlay_evpn_enabled?
       mac = config_get('overlay_global', 'anycast_gateway_mac')
-      # This value gets 0-padded when nvgened, so we need to convert it.
-      Utils.zero_pad_macaddr(mac).nil? ? default_anycast_gateway_mac : mac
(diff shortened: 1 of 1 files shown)


lib/cisco_node_utils/overlay_global.rb
```diff
@@ -127,8 +127,7 @@ def default_dup_host_mac_detection_timeout
     def anycast_gateway_mac
       return nil unless Feature.nv_overlay_evpn_enabled?
       mac = config_get('overlay_global', 'anycast_gateway_mac')
-      # This value gets 0-padded when nvgened, so we need to convert it.
-      Utils.zero_pad_macaddr(mac).nil? ? default_anycast_gateway_mac : mac
+      mac.nil? || mac.empty? ? default_anycast_gateway_mac : mac.downcase
```
glennmatthews, line 130: :+1: Does the Puppet provider need a similar update to downcase inputs for idempotence?

mikewiebe, line 130: Yes, from the PR description:
_This change will also be made to puppet in a subsequent commit to downcase the manifest data._

PR #256, opened 2016-03-01
Exr radius server

jonnytdevops:
I still need to update the minitests for EXR, but I submitted this PR for comment, thanks

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/radius_server.yaml
+++ b/lib/cisco_node_utils/cmd_ref/radius_server.yaml
@@ -1,66 +1,99 @@
 # radius_server.yaml
 ---
-_exclude: [ios_xr]
+_template:
+  nexus:
+    get_command: "show running-config radius all"
+  ios_xr:
+    get_command: "show running-config radius-server"
+    multiple: true
 
 accounting:
-  auto_default: false
+  _exclude: [ios_xr]
   kind: boolean
   get_command: "show running-config radius"
-  get_value: '/^radius-server host %s.* (accounting) .*$/'
+  get_value: '/^radius-server host <ip>.* (accounting) .*$/'
   set_value: '<state> radius-server host <ip> accounting'
   default_value: true
 
 acct-port:
-  kind: int
-  get_command: "show running-config radius all"
-  get_value: '/^radius-server host %s.* acct-port (\d+).*$/'
-  set_value: '<state> radius-server host <ip> acct-port <port>'
-  default_value: 1813
+  get_value: '/^radius-server host <ip>.* acct-port (\d+).*$/'
+  nexus:
+    default_value: 1813
+    kind: int
+    set_value: '<state> radius-server host <ip> acct-port <port>'
+  ios_xr:
+    default_value: 1646
 
 auth-port:
-  kind: int
-  get_command: "show running-config radius all"
-  get_value: '/^radius-server host %s.* auth-port (\d+).*$/'
-  set_value: '<state> radius-server host <ip> auth-port <port>'
-  default_value: 1812
+  get_value: '/^radius-server host <ip>.* auth-port (\d+).*$/'
+  nexus:
+    default_value: 1812
+    kind: int
+    set_value: '<state> radius-server host <ip> auth-port <port>'
+  ios_xr:
+    default_value: 1645
 
 authentication:
-  auto_default: false
+  _exclude: [ios_xr]
   kind: boolean
   get_command: "show running-config radius"
-  get_value: '/^radius-server host %s.* (authentication) .*$/'
+  get_value: '/^radius-server host <ip>.* (authentication) .*$/'
   set_value: '<state> radius-server host <ip> authentication'
   default_value: true
 
-hosts:
+host_port_pairs:
+  _exclude: [nexus]
   multiple: true
-  get_command: "show running-config radius"
-  get_value: '/^radius-server host (\S+).*/'
-  set_value: '<state> radius-server host <ip>'
+  get_command: "show running-config radius-server"
+  get_value: '/^radius-server host <ip> auth-port (\d+) acct-port (\d+)/'
   default_value: ''
 
+hosts:
+  get_value: '/^radius-server host (\S+).*/'
+  default_value: ''
+  nexus:
+    multiple: true
+    get_command: "show running-config radius"
+    set_value: '<state> radius-server host <ip>'
+  ios_xr:
+    set_value: '<state> radius-server host <ip> auth-port <auth_port> acct-port <acct_port>'
+
 key:
-  get_command: "show running-config radius all"
-  get_value: '/^radius-server host %s.* key \d+\s+(\S+)/'
-  set_value: '<state> radius-server host <ip> key <key>'
   default_value: ~
+  nexus:
+    get_value: '/^radius-server host <ip>.* key \d+\s+(\S+)/'
+    set_value: '<state> radius-server host <ip> key <key>'
+  ios_xr:
+    context: ["radius-server host <ip> auth-port <auth_port> acct-port <acct_port>"]
+    get_value: '/key \d+\s+(\S+)/'
+    set_value: '<state> key <key>'
 
 key_format:
-  kind: int
-  get_command: "show running-config radius all"
-  get_value: '/^radius-server host %s.* key (\d).*$/'
   default_value: ~
+  nexus:
+    get_value: '/^radius-server host <ip>.* key (\d).*$/'
+  ios_xr:
+    context: ["radius-server host <ip> auth-port <auth_port> acct-port <acct_port>"]
+    get_value: '/key (\d).*/'
 
 retransmit:
-  kind: int
-  get_command: "show running-config radius all"
-  get_value: '/^radius-server host %s.* retransmit (\d+).*$/'
-  set_value: '<state> radius-server host <ip> retransmit <count>'
   default_value: ~
+  nexus:
+    kind: int
+    get_value: '/^radius-server host <ip>.* retransmit (\d+).*$/'
+    set_value: '<state> radius-server host <ip> retransmit <count>'
+  ios_xr:
+    context: ["radius-server host <ip> auth-port <auth_port> acct-port <acct_port>"]
+    get_value: '/retransmit (\d+).*$/'
+    set_value: '<state> retransmit <count>'
 
 timeout:
-  kind: int
-  get_command: "show running-config radius all"
-  get_value: '/^radius-server host %s.* timeout (\d+).*$/'
-  set_value: '<state> radius-server host <ip> timeout <timeout>'
   default_value: ~
+  nexus:
+    kind: int
+    get_value: '/^radius-server host <ip>.* timeout (\d+).*$/'
+    set_value: '<state> radius-server host <ip> timeout <timeout>'
+  ios_xr:
+    context: ["radius-server host <ip> auth-port <auth_port> acct-port <acct_port>"]
+    get_value: '/timeout (\d+).*$/'
+    set_value: '<state> timeout <timeout>'
--- a/lib/cisco_node_utils/radius_server.rb
+++ b/lib/cisco_node_utils/radius_server.rb
@@ -25,7 +25,7 @@ module Cisco
   class RadiusServer < NodeUtil
     attr_reader :name
 
-    def initialize(name, instantiate=true)
+    def initialize(name, instantiate=true, auth_p=nil, acct_p=nil)
       unless name =~ /^[a-zA-Z0-9\.\:]*$/
         fail ArgumentError,
              'Invalid value (IPv4/IPv6 address contains invalid characters)'
@@ -39,7 +39,37 @@ module Cisco
       end
       @name = name
 
+      if platform == :ios_xr
+        if auth_p.nil?
+          @auth_port = config_get_default('radius_server', 'auth-port')
+        else
+          fail ArgumentError, 'auth_p must be an Integer' \
+            unless auth_p.is_a?(Integer)
+          @auth_port = auth_p
+        end
+
+        if acct_p.nil?
+          @acct_port = config_get_default('radius_server', 'acct-port')
+        else
+          fail ArgumentError, 'acct_p must be an Integer' \
+            unless acct_p.is_a?(Integer)
+          @acct_port = acct_p
+        end
+      end
+
       create if instantiate
+
+      return if platform == :ios_xr
+      unless auth_p.nil?
+        fail ArgumentError, 'auth_p must be an Integer' \
+          unless auth_p.is_a?(Integer)
+        self.auth_port = auth_p
+      end
+
+      return if acct_p.nil?
+      fail ArgumentError, 'acct_p must be an Integer' \
+        unless acct_p.is_a?(Integer)
+      self.acct_port = acct_p
     end
 
     def self.radiusservers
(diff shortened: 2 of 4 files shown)


lib/cisco_node_utils/cmd_ref/radius_server.yaml
```diff
@@ -9,20 +10,27 @@ accounting:
...
-  kind: int
-  get_command: "show running-config radius all"
+  multiple: true
   get_value: '/^radius-server host %s.* acct-port (\d+).*$/'
   set_value: '<state> radius-server host <ip> acct-port <port>'
   default_value: 1813
+  nexus:
+    get_command: "show running-config radius all"
+  ios_xr:
+    get_command: "show running-config radius-server"
```
glennmatthews: Consider creating a `_template` section in this YAML so you don't have to repeat the `get_command` definition over and over.

lib/cisco_node_utils/radius_server.rb
```diff
@@ -48,24 +66,70 @@ def self.radiusservers
       radiusservers_list = config_get('radius_server', 'hosts')
       return hash if radiusservers_list.empty?
       radiusservers_list.each do |id|
-        hash[id] = RadiusServer.new(id, false)
+        if platform == :ios_xr
+          authp = config_get('radius_server', 'auth-port', id)
+          authp = authp[0] if authp.is_a?(Array)
```
glennmatthews, line 83: Since you've flagged this as `multiple` in the YAML, it should always be an Array.

lib/cisco_node_utils/cmd_ref/radius_server.yaml
```diff
@@ -9,20 +10,27 @@ accounting:
   default_value: true
 
 acct-port:
-  kind: int
-  get_command: "show running-config radius all"
+  multiple: true
```
glennmatthews: I think you could add this under the `ios_xr` header (since Nexus _doesn't_ support multiple ports for the same host). Something like:

``` yaml
acct-port:
   get_value: '/^radius-server host %s.* acct-port (\d+).*$/'
   set_value: '<state> radius-server host <ip> acct-port <port>'
   default_value: 1813
   nexus:
     kind: int
   ios_xr:
     # IOS XR can have multiple ports attached to a single host
     multiple: true
```

This way the Nexus getter logic will work as-is (without needing to de-array the value)

lib/cisco_node_utils/cmd_ref/radius_server.yaml
```diff
@@ -30,35 +38,70 @@ authentication:
   set_value: '<state> radius-server host <ip> authentication'
   default_value: true
 
+host_port_pairs:
+  _exclude: [nexus]
+  multiple: true
+  get_command: "show running-config radius-server"
+  get_value: '/^radius-server host <ip> auth-port (\d+) acct-port (\d+)/'
+  set_value: '<state> radius-server host <ip> auth-port <auth_port> acct-port <acct_port>'
```
glennmatthews: It seems like this should be the set_value for `hosts` on IOS XR (having `host_port_pairs` be a getter-only property)

lib/cisco_node_utils/cmd_ref/radius_server.yaml
```diff
@@ -9,20 +10,27 @@ accounting:
...
+    get_command: "show running-config radius all"
+  ios_xr:
+    get_command: "show running-config radius-server"
 
 auth-port:
-  kind: int
-  get_command: "show running-config radius all"
+  multiple: true
   get_value: '/^radius-server host %s.* auth-port (\d+).*$/'
   set_value: '<state> radius-server host <ip> auth-port <port>'
```
glennmatthews: This set_value is Nexus-only, right? Same comment for acct-port.

lib/cisco_node_utils/radius_server.rb
```diff
@@ -39,6 +39,24 @@ def initialize(name, instantiate=true)
       end
       @name = name
 
+      if platform == :ios_xr
```
glennmatthews, line 42: Right now the Nexus initialize ignores the provided auth_p and acct_p values. That doesn't seem quite right to me.

jonnytdevops, line 42: It does this because in Nexus, we use separate setter methods whilst in EXR, we have to use the initialize.

jonnytdevops:
@glennmatthews code updated, with tests updated as well

lib/cisco_node_utils/cmd_ref/radius_server.yaml
```diff
@@ -1,66 +1,101 @@
...
   set_value: '<state> radius-server host <ip> accounting'
   default_value: true
 
 acct-port:
-  kind: int
-  get_command: "show running-config radius all"
-  get_value: '/^radius-server host %s.* acct-port (\d+).*$/'
-  set_value: '<state> radius-server host <ip> acct-port <port>'
-  default_value: 1813
+  multiple: true
```
glennmatthews: this is only multiple for XR (covered by _template above) right?

lib/cisco_node_utils/cmd_ref/radius_server.yaml
```diff
@@ -1,66 +1,101 @@
...
+    get_command: "show running-config radius"
+    set_value: '<state> radius-server host <ip>'
+  ios_xr:
+    set_value: '<state> radius-server host <ip> auth-port <auth_port> acct-port <acct_port>'
 
 key:
-  get_command: "show running-config radius all"
-  get_value: '/^radius-server host %s.* key \d+\s+(\S+)/'
-  set_value: '<state> radius-server host <ip> key <key>'
+  get_value: '/^radius-server host <ip>.* key \d+\s+(\S+)/'
```
glennmatthews: Can you move this under the `nexus:`?

lib/cisco_node_utils/cmd_ref/radius_server.yaml
```diff
@@ -1,66 +1,101 @@
...
-  kind: int
-  get_command: "show running-config radius all"
-  get_value: '/^radius-server host %s.* acct-port (\d+).*$/'
-  set_value: '<state> radius-server host <ip> acct-port <port>'
-  default_value: 1813
+  multiple: true
+  get_value: '/^radius-server host <ip>.* acct-port (\d+).*$/'
+  nexus:
+    default_value: 1813
+    get_command: "show running-config radius all"
```
glennmatthews: Does it make sense to move this get_command into the `_template`? It's common across most of the Nexus entries.

lib/cisco_node_utils/radius_server.rb
glennmatthews, line 42: If the platform is Nexus but these values are provided in the initialize, should we log a warning or anything? Or is ignoring them silently the right thing to do?

lib/cisco_node_utils/radius_server.rb
```diff
@@ -73,7 +132,13 @@ def ==(other)
     end
 
     def auth_port
-      config_get('radius_server', 'auth-port', @name)
+      if platform == :ios_xr
+        @auth_port
+      else
+        val = config_get('radius_server', 'auth-port', ip: @name)
+        val = val[0] if val.is_a?(Array)
+        val
```
glennmatthews: since auth-port is not `multiple` on Nexus, the Array check shouldn't be needed:

``` ruby
def auth_port
  platform == :ios_xr ? @auth_port : config_get('radius_server', 'auth-port', ip: @name)
end
```

lib/cisco_node_utils/radius_server.rb
```diff
@@ -81,6 +146,7 @@ def default_auth_port
     end
 
     def auth_port=(val)
+      return if platform == :ios_xr
```
glennmatthews: Not a big fan of a setter behaving as a no-op. Would it be better to raise an error?

lib/cisco_node_utils/cmd_ref/radius_server.yaml
```diff
@@ -1,66 +1,101 @@
 # radius_server.yaml
 ---
-_exclude: [ios_xr]
+_template:
+  ios_xr:
+    get_command: "show running-config radius-server"
+    multiple: true
 
 accounting:
+  _exclude: [ios_xr]
```
glennmatthews, line 11: Would it make sense at all for this to be `default_only: true` for XR instead of excluded? Not sure myself.

tests/test_radius_server.rb
```diff
@@ -47,21 +54,37 @@ def test_create_destroy_single
...
-
-    server.accounting = true
-    assert(Cisco::RadiusServer.radiusservers[id].accounting)
-
-    server.authentication = true
-    assert(Cisco::RadiusServer.radiusservers[id].authentication)
+    if platform != :ios_xr
+      # Default checking
+      assert_equal(server.default_acct_port, server.acct_port)
+      assert_equal(server.default_auth_port, server.auth_port)
```
glennmatthews: These two checks apply to XR as well, right?

lib/cisco_node_utils/cmd_ref/radius_server.yaml
jonnytdevops, line 11: I don't think so. I don't believe the accounting and authentication bools make sense at all in the context of XR.

tests/test_radius_server.rb
jonnytdevops: Nope, as the ports as set in the constructor.

jonnytdevops:
@glennmatthews Code updated, thanks

lib/cisco_node_utils/cmd_ref/radius_server.yaml
```diff
@@ -1,66 +1,96 @@
...
-hosts:
+host_port_pairs:
+  _exclude: [nexus]
   multiple: true
-  get_command: "show running-config radius"
+  get_command: "show running-config radius-server"
+  get_value: '/^radius-server host <ip> auth-port (\d+) acct-port (\d+)/'
+  default_value: ''
+
+hosts:
```
glennmatthews, line 51: `hosts` is no longer defined as `multiple` for Nexus, if I'm reading this right. Doesn't this break `RadiusServer.radiusservers` on Nexus?

lib/cisco_node_utils/radius_server.rb
```diff
@@ -48,24 +69,65 @@ def self.radiusservers
...
     end
 
     def destroy
-      config_set('radius_server',
-                 'hosts',
-                 state: 'no',
-                 ip:    @name)
+      if platform == :ios_xr
+        # This provider only support a 1-1 mapping between host and VRF.
+        # Thus, we must remove the other entries on different VRFs.
```
glennmatthews: s/VRF/port/ ?

lib/cisco_node_utils/radius_server.rb
```diff
@@ -202,6 +298,7 @@ def default_accounting
     end
 
     def accounting=(val)
+      return if platform == :ios_xr
```
glennmatthews: Please remove this (and the corresponding line in `authentication=`. Since you have `_exclude`d the `accounting` property from ios_xr, `config_set` will raise an `UnsupportedError` which is the desired behavior.

tests/test_radius_server.rb
```diff
@@ -47,21 +54,37 @@ def test_create_destroy_single
     id = '8.8.8.8'
     refute_includes(Cisco::RadiusServer.radiusservers, id)
 
-    server = Cisco::RadiusServer.new(id, true)
+    # The 55 and 44 are the acct and auth ports respectively for EXR
+    # They should be ignored for other platforms with the values remaining
+    # as the defaults, which is part of this test case.
```
glennmatthews: I still don't see the benefit of this approach. Why not instead have Nexus also call auth_port = 44 and acct_port = 55 if the values are provided in the constructor?

lib/cisco_node_utils/radius_server.rb
```diff
@@ -48,24 +69,65 @@ def self.radiusservers
...
-                 'hosts',
-                 state: 'no',
-                 ip:    @name)
+      if platform == :ios_xr
+        # This provider only support a 1-1 mapping between host and VRF.
+        # Thus, we must remove the other entries on different VRFs.
+        all_hosts = config_get('radius_server', 'host_port_pairs', ip: @name)
+        return unless all_hosts.is_a?(Array)
+
+        warn("#{name} is configured multiple times on the device" \
```
glennmatthews, line 116: I'd recommend adding a test case to the minitest that explicitly exercises this functionality (and also verifies the behavior of the various getters/setters in the case where multiple entries exist).

jonnytdevops:
@glennmatthews Code updated, however I can't get accounting or authentication to raise an unsupported error in the minitests for some reason:

```
  1) Failure:
TestRadiusServer#test_radiusserver_create_destroy_multiple [tests/test_radius_server.rb:216]:
Expected false to be nil.


  2) Failure:
TestRadiusServer#test_create_destroy_single [tests/test_radius_server.rb:78]:
Expected false to be nil.


  3) Failure:
TestRadiusServer#test_create_destroy_single_ipv6 [tests/test_radius_server.rb:142]:
Expected false to be nil.
```

glennmatthews:
@jonnytpuppet Those failures are in the getter checks (not the setter). config_get() returns `nil` for unsupported properties, but the `accounting` getter (with your latest diffs) is explicitly munging `nil` to `false` (which I think is intended for Nexus, but not for XR).

jonnytdevops:
TestRadiusServer#test_radiusserver_create_destroy_multiple = 45.69 s = .
TestRadiusServer#test_create_destroy_single_ipv6 = 31.21 s = .
TestRadiusServer#test_create_destroy_single = 33.17 s = .

jonnytdevops:
@glennmatthews Code updated, thanks

jonnytdevops:
TestRadiusServer#test_radiusserver_create_destroy_multiple = 46.78 s = .
TestRadiusServer#test_create_destroy_single_ipv6 = 30.16 s = .
TestRadiusServer#test_create_destroy_single = 32.15 s = .
W, [2016-03-08T16:48:26.103614 #34073]  WARN -- : 8.8.8.8 is configured multiple times on the device (possibly using different ports). This is unsupported by this API and the duplicate entries are being deleted.
TestRadiusServer#test_create_destroy_single_with_duplicates = 56.50 s = .

lib/cisco_node_utils/cmd_ref/radius_server.yaml
```diff
@@ -1,66 +1,95 @@
...
+  _exclude: [ios_xr]
   kind: boolean
   get_command: "show running-config radius"
-  get_value: '/^radius-server host %s.* (accounting) .*$/'
+  get_value: '/^radius-server host <ip>.* (accounting) .*$/'
   set_value: '<state> radius-server host <ip> accounting'
   default_value: true
 
 acct-port:
-  kind: int
```
glennmatthews, line 14: Looks like you need to restore these `kind` declarations, at least for Nexus:

```
  1) Failure:
TestRadiusServer#test_create_destroy_single [tests/test_radius_server.rb:76]:
Expected: 44
  Actual: "44"


  2) Failure:
TestRadiusServer#test_create_destroy_single_with_duplicates [tests/test_radius_server.rb:139]:
Expected: 44
  Actual: "44"


  3) Failure:
TestRadiusServer#test_radiusserver_create_destroy_multiple [tests/test_radius_server.rb:271]:
Expected: 44
  Actual: "44"


  4) Failure:
TestRadiusServer#test_create_destroy_single_ipv6 [tests/test_radius_server.rb:203]:
Expected: 44
  Actual: "44"
```

tests/test_radius_server.rb
```diff
@@ -142,77 +260,108 @@ def test_radiusserver_create_destroy_multiple
...
     assert_equal(44, Cisco::RadiusServer.radiusservers[id].acct_port)
-
-    server.auth_port = 55
     assert_equal(55, Cisco::RadiusServer.radiusservers[id].auth_port)
 
+    if platform != :ios_xr
+      # Default checking
+      assert_equal(server.default_accounting, server.accounting)
+      assert_equal(server.default_authentication, server.authentication)
+      assert_equal(server2.default_acct_port, server2.acct_port)
```
glennmatthews: After fixing the above failure:

```
  1) Failure:
TestRadiusServer#test_radiusserver_create_destroy_multiple [tests/test_radius_server.rb:278]:
Expected: 1813
  Actual: 44
```

jonnytdevops:
@glennmatthews Thanks for running this on Nexus. I've updated the code that should hopefully pass on both platforms now. Thanks!

glennmatthews:
```
Node under test:
  - name  - n9k-108
  - type  - N9K-C9396PX
  - image - bootflash:///nxos.7.0.3.I2.2.118.bin

TestRadiusServer#test_create_destroy_single = 4.11 s = .
TestRadiusServer#test_create_destroy_single_ipv6 = 4.40 s = .
TestRadiusServer#test_radiusserver_create_destroy_multiple = 5.89 s = .
TestRadiusServer#test_create_destroy_single_with_duplicates = 4.25 s = .

Fabulous run in 29.552781s, 0.1354 runs/s, 3.5191 assertions/s.

4 runs, 104 assertions, 0 failures, 0 errors, 0 skips
```

:+1:

@mikewiebe @robert-w-gries @chrisvanheuveln can one of you review this as well?

jonnytdevops:
@mikewiebe @robert-w-gries @chrisvanheuveln @glennmatthews Is there an update to this please? Thanks

lib/cisco_node_utils/radius_server.rb
robert-w-gries, line 42: @jonnytpuppet Did you end up changing the method to use the provided `auth_p` and `acct_p` values on Nexus?  It looks like you have changed it, but the diffs do not make it clear

jonnytdevops, line 42: Yes. For Nexus, if the ports are specified in the constructor, the setter methods are called for each port to use the constructor-provided values.

lib/cisco_node_utils/radius_server.rb
```diff
@@ -73,7 +144,8 @@ def ==(other)
     end
 
     def auth_port
-      config_get('radius_server', 'auth-port', @name)
+      platform == :ios_xr ? @auth_port : config_get('radius_server',
+                                                    'auth-port', ip: @name)
```
robert-w-gries, line 148: Shouldn't XR be able to use `config_get` in order to grab the proper value?  Same applies for `acct_port`

lib/cisco_node_utils/radius_server.rb
```diff
@@ -102,7 +178,8 @@ def auth_port=(val)
     end
 
     def acct_port
-      config_get('radius_server', 'acct-port', @name)
+      platform == :ios_xr ? @acct_port : config_get('radius_server',
+                                                    'acct-port', ip: @name)
```
robert-w-gries, line 182: See above

lib/cisco_node_utils/radius_server.rb
jonnytdevops, line 148: No, acct_port (and auth_port) are provided in the constructor and are set as instance variables. Thus, the getter for these values are effectively the same as what :attr_reader does (creates methods that return instance variable values, that were originally set in the constructor)

The thing to understand here, is that for XR, it is a combination of ip, acct_port and auth_port that define a resources uniqueness.

robert-w-gries, line 148: Okay, sounds good

robert-w-gries:
:+1:

jonnytdevops:
@glennmatthews @robert-w-gries thanks for the thumbs up guys, may we please get this merged? Thanks

PR #261, opened 2016-03-01
Feature pim fixes (n8|9k)

mikewiebe:
- Added `pim` to various `feature` files.
- Updated `pim` objects to make use of `Feature.pim_enable`

**Tests:**
- tests/test_feature.rb  `n8k, n9k(I2 & I3)` :white_check_mark: 
- tests/test_pim*.rb `n8k, n9k(I2 & I3)` :white_check_mark: 
- Support for `n5|6|6k` is beyond the scope of this commit.
- Note: PIM is currently excluded on XR

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/feature.yaml
+++ b/lib/cisco_node_utils/cmd_ref/feature.yaml
@@ -34,6 +34,12 @@ nv_overlay_evpn:
   get_value: '/^nv overlay evpn$/'
   set_value: "nv overlay evpn"
 
+pim:
+  kind: boolean
+  get_command: 'show running | section feature'
+  get_value: '/^feature pim$/' # TBD pim6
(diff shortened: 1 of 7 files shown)


lib/cisco_node_utils/cmd_ref/feature.yaml
```diff
@@ -34,6 +34,12 @@ nv_overlay_evpn:
   get_value: '/^nv overlay evpn$/'
   set_value: "nv overlay evpn"
 
+pim:
+  kind: boolean
+  get_command: 'show running | section feature'
```
chrisvanheuveln: Nit: the other get_command's are doing `'show running | i ^feature'`.

I was also thinking we could just use a top-level get_command for most of these. Let's consider changing that on the next fix to this file.

PR #262, opened 2016-03-01
bgp minitest platform unsup checks

chrisvanheuveln:
Now runs clean on N7k.

### DIFF
--- a/tests/test_router_bgp.rb
+++ b/tests/test_router_bgp.rb
@@ -479,6 +479,10 @@ class TestRouterBgp < CiscoTestCase
       return
     end
     bgp = setup_default
+    default = bgp.default_disable_policy_batching_ipv4
+    assert_equal(default, bgp.disable_policy_batching_ipv4,
+                 'bgp disable_policy_batching_ipv4 not set to default')
+
     bgp.disable_policy_batching_ipv4 = 'xx'
     assert_equal('xx', bgp.disable_policy_batching_ipv4,
                  "bgp disable_policy_batching_ipv4 should be set to 'xx'")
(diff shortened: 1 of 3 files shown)


chburket:
:+1:

tests/test_router_bgp.rb
```diff
@@ -469,66 +469,52 @@ def test_default_disable_policy_batching
...
-                 'disable_policy_batching_ipv4 default value should be nil')
-    else
-      assert_equal(bgp.default_disable_policy_batching_ipv4,
-                   bgp.disable_policy_batching_ipv4,
-                   'disable_policy_batching_ipv4 default value should be empty')
-    end
+    bgp.disable_policy_batching_ipv4 = 'xx'
+    assert_equal('xx', bgp.disable_policy_batching_ipv4,
+                 "bgp disable_policy_batching_ipv4 should be set to 'xx'")
+    bgp.disable_policy_batching_ipv4 = \
```
mikewiebe, line 489: Here and below for `ipv6` your are explicitly setting the default value, but we should also have a test for the default value right after we create the new object.

tests/test_router_bgp.rb
```diff
@@ -972,45 +958,48 @@ def test_default_maxas_limit
...
+      bgp.neighbor_down_fib_accelerate = true
+      assert(bgp.neighbor_down_fib_accelerate,
+             "vrf #{@vrf}: bgp neighbor_down_fib_accelerate "\
+             'should be enabled')
+      bgp.neighbor_down_fib_accelerate = false
+      refute(bgp.neighbor_down_fib_accelerate,
+             "vrf #{@vrf}: bgp neighbor_down_fib_accelerate "\
+             'should be disabled')
+
+      default = bgp.default_neighbor_down_fib_accelerate
```
mikewiebe: Again here, we need a default test right after creating the object.

chrisvanheuveln: nag, nag... :-)

PR #263, opened 2016-03-02
Fixes for interface minitests

chburket:
Various fixes to the test_interface, test_interface_channel_group and test_interface_switchport minitests on both nx and xr.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/interface.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface.yaml
@@ -339,7 +339,7 @@ svi_autostate:
   kind: boolean
   get_value: '/^(?:no )?autostate$/'
   set_value: "<state> autostate"
-  default_value: true
+  default_value: false
 
 svi_management:
   _exclude: [ios_xr]
--- a/lib/cisco_node_utils/cmd_ref/interface.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface.yaml
@@ -339,7 +339,7 @@ svi_autostate:
   kind: boolean
   get_value: '/^(?:no )?autostate$/'
   set_value: "<state> autostate"
-  default_value: false
+  default_value: true
 
 svi_management:
   _exclude: [ios_xr]
--- a/lib/cisco_node_utils/interface.rb
+++ b/lib/cisco_node_utils/interface.rb
@@ -1026,11 +1026,12 @@ module Cisco
     end
 
     def switchport_vtp
-      return nil unless switchport_vtp_mode_capable?
+      return default_switchport_vtp unless switchport_vtp_mode_capable?
       config_get('interface', 'vtp', name: @name)
     end
 
     def switchport_vtp=(vtp_set)
+      # TODO: throw UnsupportedError instead of returning false?
       return false unless switchport_vtp_mode_capable?
       no_cmd = (vtp_set) ? '' : 'no'
       config_set('interface', 'vtp', name: @name, state: no_cmd)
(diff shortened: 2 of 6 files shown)


lib/cisco_node_utils/interface.rb
```diff
@@ -1026,11 +1026,12 @@ def switchport_vtp_mode_capable?
     end
 
     def switchport_vtp
-      return nil unless switchport_vtp_mode_capable?
+      return default_switchport_vtp unless switchport_vtp_mode_capable?
```
chburket: @glennmatthews 

I know you suggested changing the default_switchport_vtp method so that it returns nil when switchport_vtp_mode_capable? is false, but I'm not sure that's the right way to go.  Shouldn't I be able to tell the default by looking in the yaml?  Willing to be convinced, however.

lib/cisco_node_utils/cmd_ref/interface.yaml
```diff
@@ -339,7 +339,7 @@ svi_autostate:
   kind: boolean
   get_value: '/^(?:no )?autostate$/'
   set_value: "<state> autostate"
-  default_value: true
+  default_value: false
```
glennmatthews: This seems suspect as a global change - are you sure it's not a platform difference that needs to be accounted for?

lib/cisco_node_utils/interface.rb
glennmatthews: My argument was that `switchport_vtp` is not applicable to a non-switchport interface. We should thus report it as `nil`, just like we do for e.g. svi_autostate on non-Vlan interfaces.

Note that there's a TODO under `switchport_mode` that expresses a similar concern.

Another possibility to consider would be whether we really should be looking at splitting some of these interface types/modes into separate classes, similar to how interface_channel_group was extracted from interface to support only the subset of interface properties that are applicable to an interface in a channel-group.

glennmatthews:
Some really good catches here - thanks for the thoroughness!

lib/cisco_node_utils/cmd_ref/interface.yaml
chrisvanheuveln: This one has been difficult to pin down and I think it has changed several times now. Please test this on every supported platform.

lib/cisco_node_utils/interface.rb
chrisvanheuveln: @glennmatthews We talked about doing that early on -- that if the interface class became too big or unwieldy that we should split them up. Maybe we're at that point now.

tests/test_interface_switchport.rb
```diff
@@ -427,7 +427,7 @@ def test_switchport_autostate_enabled_access
            'switchport autostate exclude')
 
     cmd_ref = cmd_ref_switchport_autostate_exclude
-    if cmd_ref.config_set?
+    if cmd_ref.setter?
```
chrisvanheuveln, line 430: Where does `setter` come from? Shouldn't this be `set_value`?

chburket, line 430: Glenn added `setter?` to command_reference.rb as part of the yaml refactoring.

lib/cisco_node_utils/cmd_ref/interface.yaml
chburket: Reverting for now.  I only ran into trouble with this from the beaker tests, so I will revisit then.

chburket:
I added a minitest case to catch the failure (previously only caught by the beaker test).  Reworked the original fix to return nil for default_switchport_vtp when switchport_mode is disabled.

Ready for a re-review.

glennmatthews:
:+1:

PR #265, opened 2016-03-02
route_distinguisher fixes for I2 vs I3

chrisvanheuveln:
- We were forcing Feature.nv_overlay_evpn_enable for all nexus devices but this is actually only necessary on N3k/N9k platforms running the older 7.0(3)I2 images. The I3 images do not require overlay to use route_distinguisher.
- I added a check in cmn utils to help look for I2 images.
- Tested on N3,5,6,7k,N9kI2, N9kI3

### DIFF
--- a/lib/cisco_node_utils/cisco_cmn_utils.rb
+++ b/lib/cisco_node_utils/cisco_cmn_utils.rb
@@ -96,6 +96,13 @@ module Cisco
   # General utility class
   class Utils
     require 'ipaddr'
+
+    # Helper utility to check for older Nexus I2 images
+    def self.nexus_i2_image
+      require_relative 'platform'
+      true if Platform.image_version[/7.0.3.I2.*/]
+    end
+
     # Helper utility method for ip/prefix format networks.
     # For ip/prefix format '1.1.1.1/24' or '2000:123:38::34/64',
     # we need to mask the address using the prefix length so that they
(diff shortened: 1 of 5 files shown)


lib/cisco_node_utils/cisco_cmn_utils.rb
```diff
@@ -96,6 +96,13 @@ def self.generic_prop_set(klass, rlbname, props)
   # General utility class
   class Utils
     require 'ipaddr'
+
+    # Helper utility to check for older Nexus I2 images
+    def self.nexus_i2_image
+      require_relative 'platform'
+      true if Platform.image_version[/7.0.3.I2.*/]
```
mikewiebe: Do we need the `.*`?  I would prefer we remove it.

chrisvanheuveln:
vrf_af has a similar I2 problem with route_target props. I'll fix that separately.

lib/cisco_node_utils/cisco_cmn_utils.rb
chrisvanheuveln: Nope. Originally I was returning the version info to the caller instead of true so I wanted to keep it. I'll remove it.

lib/cisco_node_utils/platform.rb
```diff
@@ -19,6 +19,13 @@
 module Cisco
   # Platform - class for gathering platform hardware and software information
   class Platform < NodeUtil
+    #
+    # NX: 7.0(3)I2(3) [build 7.0(3)I2(2.118)]
+    # XR: 6.1.1.04I
+    def self.image_version
```
mikewiebe, line 25: Add tests in `tests/test_platform.rb` for this.

Test would be similar to system_image below except that XR is supported.

https://github.com/cisco/cisco-network-node-utils/blob/develop/tests/test_platform.rb#L20

PR #266, opened 2016-03-02
Context for CliErrors

glennmatthews:
The key enhancement in this PR is the ability for NodeUtil to automatically wrap CliError (and similar) errors with context about the object instance whose call encountered the error. This allows us to no longer have any excuse for the copy-and-paste antipattern:

``` ruby
def foo=(val)
  config_set('bar', 'foo', val: val)
rescue Cisco::CliError => e     # Ick!
  raise "[bar #{@name}] '#{e.command}' : #{e.clierror}"   # raise a new error
end
```

In the new logic, all you need to do is define the `to_s` method for your class and the value returned by this method will be automatically prepended to any CliErrors:

``` diff
# bgp.rb
+    def to_s
+      "BGP #{asnum} VRF '#{vrf}'"
+    end
```

```
Cisco::CliError: [BGP 100 VRF 'red'] The command 'foobar shutdown' was rejected with error:
                 ^^^^^^^^^^^^^^^^^^^
                 automatically inserted
```

In the process of making this possible, I ended up refactoring a few things:
- `Node` is no longer a true Ruby singleton, but instead something that looks like a singleton. This removes the need for the `Node.lazy_connect` flag and `Node.connect` method, as the minitest can now directly instantiate a Node with the specified connection information.
- I moved all of the various error classes into a single `exceptions.rb` file and clearly spelled out the hierarchy involved. Simplified the naming somewhat by removing namespace nesting for the errors (e.g., Cisco::Client::ConnectionRefused ==> Cisco::ConnectionRefused)


docs/README-develop-best-practices.md
```diff
@@ -401,6 +402,22 @@ end
 
 See [TacacsServer](../lib/cisco_node_utils/tacacs_server.rb) and [SnmpServer](../lib/cisco_node_utils/snmpserver.rb) for examples.
 
+### <a name="co6">CO6: Implement a meaningful `to_s` method
```
mikewiebe, line 405: Perhaps some simple sample code as an example here as well?  If they use the templates they will get one, but not everyone uses templates :)

mikewiebe:
@glennmatthews What tests did you run to verify your changes?

docs/README-develop-best-practices.md
glennmatthews, line 405: Sure, will do.

glennmatthews:
`rake test` passed same as baseline against XR.
I added the `test_node_util.rb` test to exercise the error wrapping logic specifically. 
`test_nxapi.rb`, `test_node_util.rb` and several others pass against NX but I keep hitting issues running `rake test` against NX getting to a point where nxapi is no longer responding - independent of my changes.

HelenHZhang:
:+1:

mikewiebe:
:+1:

PR #271, opened 2016-03-03
Putting back terminal dont-ask in interactive cli

dcheriancisco:
1. removed the un-necessary escaped '-' from YAML regexes
2. addded back terminal dont-ask in YAML which were removed with the grpc-merge

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/vpc.yaml
+++ b/lib/cisco_node_utils/cmd_ref/vpc.yaml
@@ -60,8 +60,6 @@ dual_active_exclude_interface_vlan_bridge_domain:
 fabricpath_emulated_switch_id: 
   _exclude: [N3k, N5k, N6k, N9k]
   kind: int
-  get_command:  'show running-config vpc all'
-  get_context:  ['/^vpc domain\s*(\d+)$/']
   set_context:  ['terminal dont-ask', 'vpc domain <domain>']
   get_value: '/^fabricpath switch-id\s+(\d+)$/'
   set_value: '<state> fabricpath switch-id <swid>'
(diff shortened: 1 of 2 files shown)


lib/cisco_node_utils/cmd_ref/vpc.yaml
```diff
@@ -50,16 +50,19 @@ domain:
...
-    get_value: '/^dual\-active exclude interface\-vlan\s+(\S+)/'
+    get_value: '/^dual-active exclude interface-vlan\s+(\S+)/'
     set_value: "<state> dual-active exclude interface-vlan <range>"
   default_value: 'none'
 
 fabricpath_emulated_switch_id: 
   _exclude: [N3k, N5k, N6k, N9k]
   kind: int
+  get_command:  'show running-config vpc all'
+  get_context:  ['/^vpc domain\s*(\d+)$/']
```
glennmatthews: The `get_command` and `get_context` appear to be identical to the values provided by the `template`. I understand the need for the custom `set_context` for this property, but why these?

dcheriancisco: Not needed, it is my lack of understanding of how template works. I thought if either set_context or get_context is brought to the property level, then the template couldn't be used at all for the property.

lib/cisco_node_utils/cmd_ref/vpc.yaml
```diff
@@ -96,7 +97,9 @@ layer3_peer_routing:
 
 peer_gateway: 
   kind: boolean
-  get_value: '/^peer\-gateway/'
+  get_context:  ['/^vpc domain\s*(\d+)$/']
```
glennmatthews: this line isn't needed I think?

mikewiebe:
:+1:

dcheriancisco:
Done with commits. Please Merge

PR #272, opened 2016-03-03
config() method changes to show output.

chburket:
This does two things:
1. In debug mode, the output of all calls to `config()` will be logged.
2. No matter that the log-level, if the output of config() contains "invalid", the output will be logged as a warning.

If there are other matches I should be looking for, let me know.  I imagine this change will uncover a lot of errors that have previously been hidden.

### DIFF
--- a/tests/basetest.rb
+++ b/tests/basetest.rb
@@ -141,7 +141,7 @@ class TestCase < Minitest::Test
     # Send the entire config as one string but be sure not to return until
     # we are safely back out of config mode, i.e. prompt is
     # 'switch#' not 'switch(config)#' or 'switch(config-if)#' etc.
-   result = @device.cmd(
+    result = @device.cmd(
       'String' => "configure terminal\n" + args.join("\n") + "\nend",
       # NX-OS has a space after '#', IOS XR does not
       'Match'  => /^[^()]+[$%#>] *\z/n)
--- a/tests/basetest.rb
+++ b/tests/basetest.rb
@@ -141,9 +141,10 @@ class TestCase < Minitest::Test
     # Send the entire config as one string but be sure not to return until
     # we are safely back out of config mode, i.e. prompt is
     # 'switch#' not 'switch(config)#' or 'switch(config-if)#' etc.
-    result = @device.cmd('String' => "configure terminal\n" + args.join("\n") + "\nend",
-                # NX-OS has a space after '#', IOS XR does not
-                'Match'  => /^[^()]+[$%#>] *\z/n)
+    result = @device.cmd('String' => "configure terminal\n" + args.join("\n") +
+                         "\nend",
+                         # NX-OS has a space after '#', IOS XR does not
+                         'Match'  => /^[^()]+[$%#>] *\z/n)
 
     if /invalid/i.match(result)
       Cisco::Logger.warn("Config result:\n#{result}")
--- a/tests/basetest.rb
+++ b/tests/basetest.rb
@@ -141,9 +141,15 @@ class TestCase < Minitest::Test
     # Send the entire config as one string but be sure not to return until
     # we are safely back out of config mode, i.e. prompt is
     # 'switch#' not 'switch(config)#' or 'switch(config-if)#' etc.
-    @device.cmd('String' => "configure terminal\n" + args.join("\n") + "\nend",
+    result = @device.cmd('String' => "configure terminal\n" + args.join("\n") + "\nend",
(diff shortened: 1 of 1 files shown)


robert-w-gries:
> No matter that the log-level, if the output of config() contains "invalid", the output will be logged as a warning.

Why is this necessary?  And wouldn't it make more sense to log an invalid config command as an error?

chburket:
That statement was constructed poorly.  It should say:
- If the output of `config()` contains "invalid", the output will be logged as a warning.

This allows config errors to be seen, even when you don't have debug turned on (well, at least the ones that contain the word "invalid", which covers the ones I was seeing).  It's a step back from making the test fail, because even in my limited testing, I saw a handful of config errors, and I didn't want to create a ton of new errors to fix while we are still dealing with the aftermath of the merge.  I'm not against making it a true error, but I wanted to get a feel for how this first step played out.

tests/basetest.rb
```diff
@@ -141,9 +141,16 @@ def config(*args)
     # Send the entire config as one string but be sure not to return until
     # we are safely back out of config mode, i.e. prompt is
     # 'switch#' not 'switch(config)#' or 'switch(config-if)#' etc.
-    @device.cmd('String' => "configure terminal\n" + args.join("\n") + "\nend",
-                # NX-OS has a space after '#', IOS XR does not
-                'Match'  => /^[^()]+[$%#>] *\z/n)
+    result = @device.cmd('String' => "configure terminal\n" + args.join("\n") +
+                         "\nend",
+                         # NX-OS has a space after '#', IOS XR does not
+                         'Match'  => /^[^()]+[$%#>] *\z/n)
```
glennmatthews: Possibly cleaner:

``` ruby
result = @device.cmd(
  'String' => "configure terminal\n" + args.join("\n") + "\nend",
  # NX-OS has a space after '#', IOS XR does not
  'Match'  => /^[^()]+[$%#>] *\z/n)
```

tests/basetest.rb
```diff
@@ -141,9 +141,16 @@ def config(*args)
...
+
+    if /invalid/i.match(result)
+      Cisco::Logger.warn("Config result:\n#{result}")
+    else
+      Cisco::Logger.debug("Config result:\n#{result}")
```
glennmatthews, line 152: We don't generally use 'info' level debugging in the main code because it makes Puppet too noisy by default - but since this is test-only code, would it make sense for this to use `info` level so as to provide an intermediate level of debugging for tests that's less verbose than the full `debug` level?

tests/basetest.rb
```diff
@@ -141,9 +141,16 @@ def config(*args)
...
     # 'switch#' not 'switch(config)#' or 'switch(config-if)#' etc.
-    @device.cmd('String' => "configure terminal\n" + args.join("\n") + "\nend",
-                # NX-OS has a space after '#', IOS XR does not
-                'Match'  => /^[^()]+[$%#>] *\z/n)
+    result = @device.cmd('String' => "configure terminal\n" + args.join("\n") +
+                         "\nend",
+                         # NX-OS has a space after '#', IOS XR does not
+                         'Match'  => /^[^()]+[$%#>] *\z/n)
+
+    if /invalid/i.match(result)
```
glennmatthews: Consider also matching on `/^%/` perhaps?

tests/basetest.rb
chburket: agreed

tests/basetest.rb
chburket, line 152: I tried running test_interface with info vs. debug and ended up with ~600 lines of config output.  Seems too noisy even for a test.

chburket:
I believe I have addressed all comments.  Ready for re-review.

glennmatthews:
:+1:

robert-w-gries:
> It's a step back from making the test fail, because even in my limited testing, I saw a handful of config errors, and I didn't want to create a ton of new errors to fix while we are still dealing with the aftermath of the merge. I'm not against making it a true error, but I wanted to get a feel for how this first step played out.

Ouch, I didn't realize that there were so many config commands silently failing.  If logging the invalid config result as an error causes tests to fail, I think it's a good call to raise warnings for now so we can see the scope of the issue.  We should plan on changing it to log errors eventually, in my opinion.

:+1:

glennmatthews:
@chburket Would you mind doing a full minitest run against n9kv and send the results to the team to demonstrate how many warnings are generated? Would like an understanding of the scope.

PR #274, opened 2016-03-04
Feature/itdnode

saichint:
- Itd feature works on n7k and n9k only.
- for itd_device_group, the probe properties can be set using only one CLI
- for itd_device_group_node, the probe properties can be set using only one CLI and this is exactly same as itd_device_group but only the context is different.
- there are 2 properties (weight, hot_standby) for itd_device_group_node and these are mutually exclusive but they are highly dependent on each other internally and the order in which they are executed is very clumsy, so a new method is added to take care of all these combinations.
# Minitest results:

[saichint@sjc-ads-6603 /ws/saichint-sjc/puppet_agent/cisco-network-node-utils/tests] $ ruby test_itd_device_group_node.rb 172.23.145.152 admin nbv_1234
Run options: --seed 30382
# Running:

Node under test:
- name  - switch
- type  - N7K-C7009
- image - bootflash:///n7000-s2-dk9.7.3.0.D1.1.gbin

.....

Finished in 37.577730s, 0.1331 runs/s, 1.7564 assertions/s.

5 runs, 66 assertions, 0 failures, 0 errors, 0 skips
[saichint@sjc-ads-6603 /ws/saichint-sjc/puppet_agent/cisco-network-node-utils/tests] $ ruby test_itd_device_group.rb 172.23.145.152 admin nbv_1234
Run options: --seed 27935
# Running:

Node under test:
- name  - switch
- type  - N7K-C7009
- image - bootflash:///n7000-s2-dk9.7.3.0.D1.1.gbin

....

Finished in 20.479182s, 0.1953 runs/s, 1.9044 assertions/s.

4 runs, 39 assertions, 0 failures, 0 errors, 0 skips
[saichint@sjc-ads-6603 /ws/saichint-sjc/puppet_agent/cisco-network-node-utils/tests] $ ruby test_itd_device_group.rb 10.122.197.117 admin admin
Run options: --seed 52237
# Running:

Node under test:
- name  - agent-lab13-nx
- type  - N9K-NXOSV
- image - bootflash:///nxos.7.0.3.I3.1.bin

....

Finished in 36.173881s, 0.1106 runs/s, 1.0781 assertions/s.

4 runs, 39 assertions, 0 failures, 0 errors, 0 skips
[saichint@sjc-ads-6603 /ws/saichint-sjc/puppet_agent/cisco-network-node-utils/tests] $ ruby test_itd_device_group_node.rb 10.122.197.117 admin admin
Run options: --seed 50290
# Running:

Node under test:
- name  - agent-lab13-nx
- type  - N9K-NXOSV
- image - bootflash:///nxos.7.0.3.I3.1.bin

.....

Finished in 58.371056s, 0.0857 runs/s, 1.1307 assertions/s.

5 runs, 66 assertions, 0 failures, 0 errors, 0 skips


lib/cisco_node_utils/itd_device_group.rb
```diff
@@ -0,0 +1,224 @@
...
+        group_list << group.split[0]
+      end
+
+      group_list.each do |id|
+        hash[id] = ItdDeviceGroup.new(id, false)
+      end
+      hash
+    end
+
+    # feature itd
```
glennmatthews: This should be added to the Feature class instead of here, shouldn't it?

tests/test_itd_device_group_node.rb
```diff
@@ -0,0 +1,215 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative 'ciscotest'
+require_relative '../lib/cisco_node_utils/itd_device_group_node'
+
+include Cisco
+# TestInterface - Minitest for general functionality
+# of the ItdDeviceGroup class.
+class TestItdDeviceGroupNode < CiscoTestCase
```
glennmatthews, line 21: Use `@skip_unless_supported = 'itd_device_group'` here so as to automatically skip all tests on an unsupported platform. This will also remove the need for the `n79k_platform?` check throughout this class.

saichint, line 21: Added

saichint:
waiting for any other comments.

lib/cisco_node_utils/cmd_ref/itd_device_group.yaml
```diff
@@ -0,0 +1,83 @@
+# itd_device_group
+---
+_exclude: [N3k, N5k, N6k, ios_xr]
+
+_template:
+  get_command: "show running-config all | section itd"
+  get_context:
+      - '/^itd device-group <name>/'
+      - '(?)/^node <ntype> <nname>/'
```
glennmatthews, line 9: If I understand this PR correctly, you're adding support for two types - itd_device_group and itd_device_group_node, and some of the below properties are specific to itd_device_group while others are specific to itd_device_group_node. If that's correct, rather than marking the 'node' as optional in the YAML (meaning itd_device_group properties may or may not have a node, and itd_device_group_node properties also may or may not have a node), I would strongly recommend that you create two separate YAML files - itd_device_group.yaml (which NEVER contains a 'node' line in the context) and itd_device_group_node.yaml (which ALWAYS contains a 'node' line in the context).

saichint, line 9: Actually, itd_device_group_node has 2 extra properties (weight and hot_standby), the other properties of itd_device_group are all applicable to itd_device_group_node (only the context is different, we have an extra node context for the common properties) and I reused all the properties and methods in the ruby file by deriving itd_device_group_node from itd_device_group. This way I have kept the code to minimum. If we separate them into 2 files, then I have to keep almost the entire file same except add the other 2 properties in the new file. If you think, it is ok to repeat, I can break them into 2 files. We can also discuss over phone if you want. Let me know.

lib/cisco_node_utils/itd_device_group.rb
```diff
@@ -0,0 +1,210 @@
...
+      hash = {}
+      group_list = []
+      groups = config_get('itd_device_group',
+                          'all_itd_device_groups')
+      return hash if groups.nil?
+
+      groups.each do |group|
+        # The show cmd shows more than name if the probe is not confgured yet
+        # itd device-group abc frequency 10 timeout 5 retry-down-count 3 ...
+        # so filter it out to just get the name
```
mikewiebe: You should not need to filter if you change the `get_value` yaml entry to the following.

`get_value: '/^itd device-group (\S+)/'`

lib/cisco_node_utils/itd_device_group.rb
```diff
@@ -0,0 +1,210 @@
...
+    end
+
+    # Helper method to delete @set_args hash keys
+    def set_args_keys_default
+      @set_args = { name: @name }
+      @get_args = @set_args
+    end
+
+    def probe_params
+      params = config_get('itd_device_group', 'probe', @get_args)
```
mikewiebe: This should be simplified to:

```
def probe_params
  config_get('itd_device_group', 'probe', @get_args)
end
```

lib/cisco_node_utils/itd_device_group.rb
```diff
@@ -0,0 +1,210 @@
...
+      params
+    end
+
+    # proble configuration is all done in a single line (like below)
```
mikewiebe: `/proble/probe/`

lib/cisco_node_utils/itd_device_group.rb
```diff
@@ -0,0 +1,210 @@
...
+    # probe tcp port 32 frequency 10 timeout 5 retry-down-count 3 ...
+    # probe udp port 23 frequency 10 timeout 5 retry-down-count 3 ...
+    # probe icmp frequency 10 timeout 5 retry-down-count 3 retry-up-count 3
+    # probe dns host 8.8.8.8 frequency 10 timeout 5 retry-down-count 3 ...
+    # also the 'control enable' can be set if the type is tcp or udp only
+    # probe udp port 23 control enable frequency 10 timeout 5 ...
+    def probe
```
mikewiebe: Suggest renaming this method to `probe_get`.  More intuitive.

lib/cisco_node_utils/itd_device_group.rb
```diff
@@ -0,0 +1,210 @@
...
+      when :dns
+        hash[:probe_dns_host] = lparams[2]
+      when :tcp, :udp
+        hash[:probe_port] = lparams[2].to_i
+        hash[:probe_control] = true unless lparams[3].nil?
+      end
+      hash
+    end
+
+    def probe_control
```
mikewiebe, line 115: You should be able to simplify these to just

```
def probe_control
  probe[:probe_control] # Or probe_get[:probe_control] if you follow my suggestion above.
end
```

lib/cisco_node_utils/itd_device_group.rb
```diff
@@ -0,0 +1,210 @@
...
+
+    # proble configuration is all done in a single line (like below)
+    # probe tcp port 32 frequency 10 timeout 5 retry-down-count 3 ...
+    # probe udp port 23 frequency 10 timeout 5 retry-down-count 3 ...
+    # probe icmp frequency 10 timeout 5 retry-down-count 3 retry-up-count 3
+    # probe dns host 8.8.8.8 frequency 10 timeout 5 retry-down-count 3 ...
+    # also the 'control enable' can be set if the type is tcp or udp only
+    # probe udp port 23 control enable frequency 10 timeout 5 ...
+    def probe
+      params = probe_params
```
mikewiebe: I would just get rid of the `probe_params` method and instead `params = config_get...`

lib/cisco_node_utils/itd_device_group.rb
```diff
@@ -0,0 +1,210 @@
...
+        hash[:probe_retry_up] = default_probe_retry_up
+        hash[:probe_type] = default_probe_type
+        return hash
+      end
+      hash[:probe_frequency] = params[1].to_i
+      hash[:probe_timeout] = params[2].to_i
+      hash[:probe_retry_down] = params[3].to_i
+      hash[:probe_retry_up] = params[4].to_i
+
+      lparams = params[0].split
```
mikewiebe: Setting positional attributes like this is ok, but consider using a keyed regexp instead of the more ambiguous `params[index] and lparams[index]`
- Example creating a getter regexp: https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/ace.rb#L77
  - NOTE: Your case will not be nearly this complex.
- Then, to pull out a single attribute: https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/ace.rb#L171

lib/cisco_node_utils/itd_device_group.rb
```diff
@@ -0,0 +1,210 @@
...
+      hash = probe
+      hash[:probe_control]
+    end
+
+    def default_probe_control
+      config_get_default('itd_device_group', 'probe_control')
+    end
+
+    def probe_dns_host
+      hash = probe
```
mikewiebe: Do you need to handle the case where the `:probe_dns_host` key does not exist.. ie, `:probe_type` is `:tcp` or `:udp`?

lib/cisco_node_utils/itd_device_group.rb
```diff
@@ -0,0 +1,210 @@
...
+      hash[:probe_frequency]
+    end
+
+    def default_probe_frequency
+      config_get_default('itd_device_group', 'probe_frequency')
+    end
+
+    def probe_port
+      hash = probe
```
mikewiebe: Similar question here.  This key may not exist right?

lib/cisco_node_utils/itd_device_group.rb
mikewiebe, line 115: Also, same question as below for `:probe_dns_host`.  This key may not exist right?

lib/cisco_node_utils/itd_device_group.rb
```diff
@@ -0,0 +1,210 @@
...
+    def probe_type
+      hash = probe
+      hash[:probe_type]
+    end
+
+    def default_probe_type
+      config_get_default('itd_device_group', 'probe_type')
+    end
+
+    def probe=(type, host, control, freq, ret_up, ret_down, port, timeout)
```
mikewiebe: I would strongly recommend you follow the `ace` design pattern for the setter.  We already have a working design pattern that uses the `ace_set` method in puppet.  This allows you to simply pass all of the arguments to the setter as a hash.
- NodeUtils `ace_set(attrs)` method: https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/ace.rb#L100
- Then in puppet you can do this: https://github.com/cisco/cisco-network-puppet-module/blob/develop/lib/puppet/provider/cisco_ace/cisco.rb#L157

lib/cisco_node_utils/itd_device_group_node.rb
```diff
@@ -0,0 +1,139 @@
...
+
+      @itd_device_group_name = itd_dg_name
+      @itddg = ItdDeviceGroup.itds[itd_dg_name]
+      fail "itd device-group #{itd_dg_name} does not exist" if
+      @itddg.nil?
+      @name = node_name
+      @node_type = node_type
+
+      set_args_keys_default
+      return unless create
```
mikewiebe: Why is this different then our typical pattern of `create if instantiate`?  I suppose in this case though it would be `create_node if instantiate`.

lib/cisco_node_utils/cmd_ref/itd_device_group.yaml
```diff
@@ -0,0 +1,83 @@
...
+
+all_itd_device_group_nodes:
+  multiple:
+  get_context: ['/^itd device-group <name>/']
+  get_value: '/^node (.*)/'
+
+all_itd_device_groups:
+  multiple:
+  get_context: ~
+  get_value: '/^itd device-group (.*)/'
```
mikewiebe: Change to `get_value: '/^itd device-group (\S+)/'`

lib/cisco_node_utils/itd_device_group_node.rb
```diff
@@ -0,0 +1,139 @@
...
+      fail TypeError unless node_name.is_a?(String) || node_name.nil?
+      itd_nodes = {}
+      itd_list = ItdDeviceGroup.itds
+      return itd_nodes if itd_list.nil?
+      itd_list.keys.each do |name|
+        itd_nodes[name] = {}
+        match = config_get('itd_device_group',
+                           'all_itd_device_group_nodes', name: name)
+        next if match.nil?
+        match.each do |line|
```
mikewiebe: Can you add an example line in the comments that you are matching here?

lib/cisco_node_utils/cmd_ref/feature.yaml
```diff
@@ -20,6 +20,12 @@ fabric_forwarding:
   get_value: '/^feature fabric forwarding$/'
   set_value: "feature fabric forwarding"
 
+itd:
+  _exclude: [N3k, N5k, N6k]
```
mikewiebe, line 24: We need to verify if this is supported on N8k.  I will send you an email with connection details to my N8k to verify the cli.

lib/cisco_node_utils/cmd_ref/itd_device_group.yaml
```diff
@@ -0,0 +1,83 @@
+# itd_device_group
+---
+_exclude: [N3k, N5k, N6k, ios_xr]
```
mikewiebe: We need to verify if this is supported on N8k.  I will send you an email with connection details to my N8k to verify the cli.

lib/cisco_node_utils/itd_device_group_node.rb
saichint: added

lib/cisco_node_utils/cmd_ref/feature.yaml
saichint, line 24: Checked with Deepak and we decided to not support n8k until its code comes close to n7k/n9k.

saichint:
Mike,
I have done all the changes you asked for. Latest minitest results are attached. Let me know if you have additional comments.

lib/cisco_node_utils/cmd_ref/itd_device_group.yaml
mikewiebe: Can you add `N8k` to the exclusion list where applicable?

saichint: Added N8k to exclude list

lib/cisco_node_utils/cmd_ref/itd_device_group.yaml
mikewiebe, line 9: @glennmatthews Have you and Sai been able to sync on this point?  I don't have a strong opinion here, but I would favor less duplication of YAML entries.

saichint, line 9: It is not just yaml but the ruby code as well. Currently with this design, the ruby code also does not duplicate. When standby ip provider comes for itd (it is buggy right now so we deferred), the same code gets duplicated yet again so we will have 3 copies of the same thing (yaml file and ruby) if we break up files.

lib/cisco_node_utils/cmd_ref/itd_device_group.yaml
```diff
@@ -0,0 +1,83 @@
...
+  get_command: "show running-config all | section itd"
+  get_context:
+      - '/^itd device-group <name>/'
+      - '(?)/^node <ntype> <nname>/'
+  set_context: ["itd device-group <name>", "(?)node <ntype> <nname>"]
+
+all_itd_device_group_nodes:
+  multiple:
+  get_context: ['/^itd device-group <name>/']
+  get_value: '/^node (.*)/'
```
mikewiebe: I think this could be cleaner if you match on both node afi and node address.

get_value: '/^node (\S+) (\S+)/'

This will allow you to streamline this code a bit more.

https://github.com/cisco/cisco-network-node-utils/pull/274/files#diff-c621588334e7906861d7071affb1c0abR56

saichint: Done

lib/cisco_node_utils/itd_device_group_node.rb
```diff
@@ -0,0 +1,145 @@
...
+      @set_args[:state] = val == default_weight ? 'no' : ''
+      @set_args[:weight] = val
+      config_set('itd_device_group', 'weight', @set_args)
+    end
+
+    # Call this for setting hot_standby and weight together because
+    # the CLI is pretty weird and it accepts these params in a very
+    # particular way and they cannot even be reset unless proper
+    # order is followed
+    def hs_weight=(hs, wt)
```
mikewiebe: IMO, For multiline setters like this it's better to ommit the `=`.  Just use `def hs_weight_set(hs, wt)`.

That way you can call it like this `o.hs_weight_set(hs, wt)` instead of like this `o.send(:hs_weight=, hot_standby, weight)`  It's also more consistent with our approach elsewhere.

Example:

https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/router_ospf_vrf.rb#L100

saichint: This I am not in complete agreement with you. 
If you look at 
https://github.com/saichint/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/fabricpath_global.rb#L323
it is following different approach. Also, fabricpath code came later than ospf, I believe. So I don't think it is consistent everywhere. 
If you insist, I will change it but I am not going to agree that this will make consistent code, it would be just an opinion difference.

mikewiebe: Looks like fabricpath slipped through :smile: . I am in favor of the more concise approach, but this is fine for now.

saichint: Ok. changed it as part of goodwill :smile:. I agree with the concise approach, I was only contesting the consistency

lib/cisco_node_utils/cmd_ref/itd_device_group.yaml
glennmatthews, line 9: Thanks Sai - I understand the reason now for having this as a shared file. I'm OK with your approach now that I know the reasoning.

mikewiebe:
:+1:

PR #275, opened 2016-03-07
US75671: cisco_vrf:nu:new SILK1 attributes

HelenHZhang:
All unit tests passed on eXR and n9kv.


lib/cisco_node_utils/cmd_ref/vrf.yaml
```diff
@@ -41,6 +41,25 @@ destroy:
   ios_xr:
     set_value: 'no vrf <vrf>'
 
+mhost_ipv4:
```
chburket: Did you consider combining mhost_ipv4 and mhost_ipv6?

HelenHZhang: I thought there are 2 different CLI commands, which need to be 2 items in the YAML file. Is it true?

chburket: Any number of CLI commands can share yaml entries.  It's just a question of "does it make sense".

mikewiebe: I think that combining v4 and v6 does make sense in this case given that the commands are virtually identical except for the afi.

```
get_value '/^mhost <afi> default-interface (\S+)$/'
set_value: '<state> mhost <afi> default-interface <intf>'
```

```
```

glennmatthews: Agree with Mike and Charles. A second question - is 'mhost' a sufficiently specific name for this property, or should it really be 'mhost_default_interface'?

glennmatthews:
Please update CHANGELOG.md as well.

tests/test_vrf.rb
```diff
@@ -107,6 +107,59 @@ def test_description
     v.destroy
   end
 
+  def test_mhost
+    v = Vrf.new('test_mhost')
+    t_intf = 'Loopback100'
+    if platform == :nexus
```
mikewiebe: https://github.com/cisco/cisco-network-node-utils/pull/279, added a new method `validate_property_excluded?()` that can be used instead of checking against a specific platform or os.

```
if validate_property_excluded?('vrf', 'mhost_ipv4')
  # assert_nil and assert_raises checks here
end
```

HelenHZhang: fixed it.

tests/test_vrf.rb
```diff
@@ -107,6 +107,59 @@ def test_description
     v.destroy
   end
 
+  def test_mhost
+    v = Vrf.new('test_mhost')
+    t_intf = 'Loopback100'
+    if validate_property_excluded?('vrf', 'mhost_ipv4')
```
mikewiebe: This should be `mhost_default_interface` instead of `mhost_ipv4` after the yaml change right?

HelenHZhang: fixed it

HelenHZhang:
I resolved all comments, re-ran all minitests, all passed on eXR and n9kv.

mikewiebe:
:+1:

chburket:
Back to Glenn's comment... should this property be named mhost_ipv4_default_interface instead of just mhost_ipv4?  You're not getting/setting the mhost, you are getting/setting the default interface of the mhost, right?

HelenHZhang:
changed mhost_ipv4[6] to mhost_ipv4[6]_default_interface.
re-ran all minitests on eXR and n9kv, and all passed. Could you review it again?

chburket:
:+1:

PR #276, opened 2016-03-07
Cisco package APIs for XR

sushirol:


lib/cisco_node_utils/yum.rb
```diff
@@ -66,27 +66,78 @@ def self.detect_vrf
...
+    def self.query(pkg, src)
+      if "#{platform}" == 'nexus'
+        fail TypeError unless pkg.is_a? String
+        fail ArgumentError if pkg.empty?
+        b = config_get('yum', 'query', pkg)
+        fail "Multiple matching packages found for #{pkg}" if b && b.size > 1
+
+      elsif "#{platform}" == 'ios_xr'
+        filename = src.strip.tr(':', '/').split('/').last
+        name_arch_regex_xr = /^(.*\d.*-[\d.]*-r\d\d\d\d*.).(\w{4,}).rpm/
```
glennmatthews: This is a pretty gnarly regexp, and you're declaring it in two different places. Can you make it a constant and add comments showing examples of what it should and should not be matching?

glennmatthews: For that matter, can you share any of the name regex logic with Nexus, or restructure the logic so the two have a common code flow? Right now it looks like you're doing something kinda like Nexus but in a completely different way (not using validate() or decompose_name() but instead doing the checks inline). Maybe I'm misunderstanding things.

lib/cisco_node_utils/yum.rb
```diff
@@ -66,27 +66,78 @@ def self.detect_vrf
...
+    def self.remove(pkg, src)
+      if "#{platform}" == 'nexus'
+        config_set('yum', 'remove', pkg)
+      elsif "#{platform}" == 'ios_xr'
+        filename = src.strip.tr(':', '/').split('/').last
+        name_arch_regex_xr = /^(.*\d.*-[\d.]*-r\d\d\d\d*.).(\w{4,}).rpm-(.*)$/
+        if filename =~ name_arch_regex_xr
+          pkg_name = Regexp.last_match(1)
+          debug "Removing package #{pkg_name}"
+          `ip netns exec default sdr_instcmd install \
```
glennmatthews: Consider defining `'ip netns exec default sdr_instcmd'` as a constant to reuse for readability.

lib/cisco_node_utils/yum.rb
```diff
@@ -66,27 +66,78 @@ def self.detect_vrf
...
     end
 
     # returns version of package, or false if package doesn't exist
-    def self.query(pkg)
-      fail TypeError unless pkg.is_a? String
-      fail ArgumentError if pkg.empty?
-      b = config_get('yum', 'query', pkg)
-      fail "Multiple matching packages found for #{pkg}" if b && b.size > 1
+    def self.query(pkg, src)
+      if "#{platform}" == 'nexus'
```
glennmatthews: Why this approach instead of `if platform == :nexus`?

lib/cisco_node_utils/yum.rb
```diff
@@ -66,9 +76,22 @@ def self.detect_vrf
...
+        if pkg_info
+          pkg_name = "#{pkg_info[1]}-#{pkg_info[2]}-#{pkg_info[3]}"
+          debug "Installing package #{pkg_name}"
+          `#{EXEC_IN_DEFAULT_NS} install activate pkg 0x0 #{pkg_name}`
+          `#{EXEC_IN_DEFAULT_NS} install commit sdr`
+        else
+          fail 'Failed to install the requested rpm.'
+        end
+      end
       # HACK: The current nxos host installer is a multi-part command
```
glennmatthews: Is this check specific to Nexus (in which case please move it up into the if block above), or is it needed for XR as well (in which case please update the comment to make this clear).

glennmatthews:
Do you have any ability to run minitest for XR to validate these changes?

lib/cisco_node_utils/yum.rb
sushirol: @glennmatthews This check is needed for XR as well. I will update the comment.

sushirol:
@glennmatthews 
No. I can not use minitest since i am using bash command to install packages. 
I validated these changes through puppet. 
/opt/puppetlabs/puppet/bin/puppet apply demo_cisco_patch_rpm.pp --debug

lib/cisco_node_utils/yum.rb
```diff
@@ -20,8 +20,9 @@
 require_relative 'node_util'
 
 module Cisco
-  # This Yum class provides cisco package management functions through nxapi.
+  # This Yum class provides cisco package management functions.
   class Yum < NodeUtil
+    EXEC_IN_DEFAULT_NS = 'ip netns exec default sdr_instcmd'
```
glennmatthews: This is a bit misleading since `sdr_instcmd` is the actual command being executed - I would expect EXEC_IN_DEFAULT_NS to be simply `'ip netns exec default'`.

chburket:
:+1:

lib/cisco_node_utils/yum.rb
```diff
@@ -32,9 +33,16 @@ def self.decompose_name(file_name)
       # ex: b+z-ip2.x64_64
       name_arch_regex = /^([\w\-\+]+)\.(\w+)$/
 
-      file_name.match(name_ver_arch_regex) ||
-        file_name.match(name_ver_arch_regex_nx) ||
-        file_name.match(name_arch_regex)
+      # xrv9k-k9sec-1.0.0.0-r61102I.x86_64.rpm-XR-DEV-16.02.22C
+      name_arch_regex_xr = /^(.*\d.*)-([\d.]*)-(r\d\d\d\d*.).(\w{4,}).rpm/
```
glennmatthews: The above aren't super clear, but this seems to me like it should be `name_ver_arch_regex_xr` as it's parallel to `name_ver_arch_regex_nx`, not `name_arch_regex`.

lib/cisco_node_utils/yum.rb
```diff
@@ -20,8 +20,9 @@
 require_relative 'node_util'
 
 module Cisco
-  # This Yum class provides cisco package management functions through nxapi.
+  # This Yum class provides cisco package management functions.
   class Yum < NodeUtil
+    EXEC_IN_DEFAULT_NS = 'ip netns exec default sdr_instcmd'
     def self.decompose_name(file_name)
```
glennmatthews: Can you add a comment describing what this function is expected to return? Looks like the match-group (package_name, package_version, os_version, ???)

lib/cisco_node_utils/yum.rb
glennmatthews: Looking at the 6.0.0 release, it looks like release RPMs may follow the format `'xrv9k-k9sec-1.0.0.0-r600.x86_64.rpm-6.0.0'` which your match group `(r\d\d\d\d*.)` would fail to match against. Also, you should escape the literal `.`s before `rpm` and before the architecture match group. Perhaps the below is preferable:

`name_arch_regex_xr = /^(.*\d.*)-([\d.]*)-(r\d+.*)\.(\w{4,})\.rpm/`

lib/cisco_node_utils/yum.rb
```diff
@@ -66,27 +76,73 @@ def self.detect_vrf
...
+      elsif platform == :ios_xr
+        filename = pkg.strip.tr(':', '/').split('/').last
+        pkg_info = Yum.decompose_name(filename)
+        if pkg_info
+          pkg_name = "#{pkg_info[1]}-#{pkg_info[2]}-#{pkg_info[3]}"
+          debug "Installing package #{pkg_name}"
+          `#{EXEC_IN_DEFAULT_NS} install activate pkg 0x0 #{pkg_name}`
+          `#{EXEC_IN_DEFAULT_NS} install commit sdr`
+        else
+          fail 'Failed to install the requested rpm.'
```
glennmatthews: Can you add more context to this failure message? Looks like you get here if you're not going to try to install it (because the decompose_name call failed), not because installation failed.

lib/cisco_node_utils/yum.rb
```diff
@@ -66,27 +76,73 @@ def self.detect_vrf
...
-      # therefore a post-validation check is needed here to verify the
-      # actual outcome.
+      elsif platform == :ios_xr
+        filename = pkg.strip.tr(':', '/').split('/').last
+        pkg_info = Yum.decompose_name(filename)
+        if pkg_info
+          pkg_name = "#{pkg_info[1]}-#{pkg_info[2]}-#{pkg_info[3]}"
+          debug "Installing package #{pkg_name}"
+          `#{EXEC_IN_DEFAULT_NS} install activate pkg 0x0 #{pkg_name}`
+          `#{EXEC_IN_DEFAULT_NS} install commit sdr`
```
glennmatthews: What if anything do you need to do to check for errors from either of these commands?

lib/cisco_node_utils/yum.rb
```diff
@@ -20,8 +20,9 @@
 require_relative 'node_util'
 
 module Cisco
-  # This Yum class provides cisco package management functions through nxapi.
+  # This Yum class provides cisco package management functions.
   class Yum < NodeUtil
+    EXEC_IN_DEFAULT_NS = 'ip netns exec default sdr_instcmd'
     def self.decompose_name(file_name)
       # ex: chef-12.0.0alpha.2+20150319.git.1.b6f-1.el5.x86_64.rpm
       name_ver_arch_regex = /^([\w\-\+]+)-(\d+\..*)\.(\w{4,})(?:\.rpm)?$/
```
glennmatthews: This regex is for matching third-party RPMs (as opposed to OS RPMs). Does it apply to XR as well, and if not, why not?

lib/cisco_node_utils/yum.rb
```diff
@@ -66,27 +80,77 @@ def self.detect_vrf
...
     # returns version of package, or false if package doesn't exist
-    def self.query(pkg)
-      fail TypeError unless pkg.is_a? String
-      fail ArgumentError if pkg.empty?
-      b = config_get('yum', 'query', pkg)
-      fail "Multiple matching packages found for #{pkg}" if b && b.size > 1
+    def self.query(pkg, src)
+      if platform == :nexus
+        fail TypeError unless pkg.is_a? String
+        fail ArgumentError if pkg.empty?
```
glennmatthews: Don't these two input validation checks also apply to XR?

lib/cisco_node_utils/yum.rb
```diff
@@ -66,27 +80,77 @@ def self.detect_vrf
...
+
+      elsif platform == :ios_xr
+        filename = src.strip.tr(':', '/').split('/').last
+        pkg_info = Yum.decompose_name(filename)
+        pkg_name = pkg_info[1]
+        should_version = pkg_info[2]
+        xr_version = pkg_info[3]
+        platform_var = pkg_info[4]
+        query_package_name = "#{pkg_name}-#{should_version}" \
+                            "-#{xr_version}.#{platform_var}"
```
glennmatthews: If this needs to be wrapped to keep it under 80 characters, I'd prefer:

``` ruby
        query_package_name = \
          "#{pkg_name}-#{should_version}-#{xr_version}.#{platform_var}"
```

glennmatthews:
A couple of nits otherwise :+1:

sushirol:
@glennmatthews @chburket I have pushed a new set of changes. Ready for a review.

lib/cisco_node_utils/yum.rb
```diff
@@ -20,8 +20,15 @@
...
-  # This Yum class provides cisco package management functions through nxapi.
+  # This Yum class provides cisco package management functions.
   class Yum < NodeUtil
+    EXEC_IN_DEFAULT_NS = 'ip netns exec default'
+
+    # This function accepts name of the rpm package and returns a match group.
+    # Match group when platform is ios_xr :
+    #   1. package name, 2. package version, 3. os version, 4. platform
+    # Match group when platform is nexus :
+    #   1. package name 2. package version 3. platform
```
glennmatthews: What about name_arch_regex, which only has two match groups? Perhaps someone from the Nexus side can comment as to the design/usage of this function as it seems very obtuse to me - @mikewiebe, @chrisvanheuveln, @robert-w-gries?

lib/cisco_node_utils/yum.rb
```diff
@@ -46,11 +61,19 @@ def self.validate(pkg)
...
+          elsif platform == :ios_xr
+            # ex query_name xrv9k-k9sec-1.0.0.0-r61102I.x86_64
+            query_name = \
+              "#{pkg_info[1]}-#{pkg_info[2]}-#{pkg_info[3]}.#{pkg_info[4]}"
+          end
         end
       end
       should_ver = pkg_info[2] if pkg_info && pkg_info[3]
       ver = query(query_name)
+      debug "Installed package version #{ver}, expected package version" \
```
glennmatthews: Doesn't this need to be `Cisco::Logger.debug`?

lib/cisco_node_utils/yum.rb
```diff
@@ -66,27 +89,84 @@ def self.detect_vrf
...
+      if platform == :nexus
+        vrf = vrf.nil? ? detect_vrf : "vrf #{vrf}"
+        config_set('yum', 'install', pkg, vrf)
 
-      # HACK: The current nxos host installer is a multi-part command
-      # which may fail at a later stage yet return a false positive;
-      # therefore a post-validation check is needed here to verify the
-      # actual outcome.
+      elsif platform == :ios_xr
+        filename = pkg.strip.tr(':', '/').split('/').last
```
glennmatthews: Why is the `tr(':', '/')` needed here? 

More generally, it seems odd to me that you're discarding the directory structure altogether - if I call `install('/foo/bar/xrv9k-foo.rpm')`, does `sdr_instcmd install activate pkg 0x0 xrv9k-foo` really work? How does it magically find the package?

lib/cisco_node_utils/yum.rb
```diff
@@ -66,27 +89,84 @@ def self.detect_vrf
...
       fail ArgumentError if pkg.empty?
-      b = config_get('yum', 'query', pkg)
-      fail "Multiple matching packages found for #{pkg}" if b && b.size > 1
+      if platform == :nexus
+        b = config_get('yum', 'query', pkg)
+        fail "Multiple matching packages found for #{pkg}" if b && b.size > 1
+
+      elsif platform == :ios_xr
+        package_info = `#{EXEC_IN_DEFAULT_NS} sdr_instcmd show install \
+                     package #{pkg} none`
```
glennmatthews, line 122: Can't you use grpc and config_get for this, given it's a show command?

lib/cisco_node_utils/yum.rb
sushirol: @glennmatthews name_arch_regex is used by nexus code. i didnt see any XR rpm package following similar filename pattern.

lib/cisco_node_utils/yum.rb
sushirol: I just followed the existing code to print debug messages. I can change it to Cisco::Logger.debug.

lib/cisco_node_utils/yum.rb
glennmatthews: My point was that you said that platform nexus has three match groups, but in the name_arch_regex it only has two - so this comment is incomplete/inaccurate.

chburket:
Going to close this PR and open a new one with reworked code, for clarity.

PR #277, opened 2016-03-07
Fix BGP minitest errors/warnings.

chburket:
Fixes for errors/warnings in minitests, mostly BGP related.
Improved flexibility of config method with regards to warning on error.

### DIFF
--- a/tests/basetest.rb
+++ b/tests/basetest.rb
@@ -162,7 +162,7 @@ class TestCase < Minitest::Test
       # NX-OS has a space after '#', IOS XR does not
       'Match'  => /^[^()]+[$%#>] *\z/n)
 
-    if !warn_match.nil? && warn_match.match(result)
+    if warn_match && warn_match.match(result)
       Cisco::Logger.warn("Config result:\n#{result}")
     else
       Cisco::Logger.debug("Config result:\n#{result}")
--- a/tests/basetest.rb
+++ b/tests/basetest.rb
@@ -140,7 +140,7 @@ class TestCase < Minitest::Test
   # Execute the specified config commands and warn if the
   # output matches the default "warning" regex.
   def config(*args)
-    config_and_warn_on_match(/invalid|^%/i, *args)
+    config_and_warn_on_match(/^invalid|^%/i, *args)
   end
 
   # Execute the specified config commands. Use this version
--- a/tests/basetest.rb
+++ b/tests/basetest.rb
@@ -138,20 +138,20 @@ class TestCase < Minitest::Test
   end
 
   # Execute the specified config commands and warn if the
-  # ouput matches the following regex: /invalid|^%/i
+  # output matches the default "warning" regex.
   def config(*args)
-     config_and_warn_on_match(/invalid|^%/i, *args)
+    config_and_warn_on_match(/invalid|^%/i, *args)
   end
 
   # Execute the specified config commands. Use this version
   # of the config method if you expect possible config errors
   # and do not wish to log them as a warning.
   def config_no_warn(*args)
-     config_and_warn_on_match(nil, *args)
+    config_and_warn_on_match(nil, *args)
   end
 
   # Execute the specified config commands and warn if the
-  # ouput matches the specified regex.5  Specifying nil for
+  # ouput matches the specified regex.  Specifying nil for
   # warn_match means "do not warn".
   def config_and_warn_on_match(warn_match, *args)
     # Send the entire config as one string but be sure not to return until
--- a/tests/basetest.rb
+++ b/tests/basetest.rb
@@ -137,7 +137,23 @@ class TestCase < Minitest::Test
     @device = nil
   end
 
+  # Execute the specified config commands and warn if the
+  # ouput matches the following regex: /invalid|^%/i
   def config(*args)
+     config_and_warn_on_match(/invalid|^%/i, *args)
+  end
+
+  # Execute the specified config commands. Use this version
+  # of the config method if you expect possible config errors
+  # and do not wish to log them as a warning.
+  def config_no_warn(*args)
+     config_and_warn_on_match(nil, *args)
+  end
+
+  # Execute the specified config commands and warn if the
+  # ouput matches the specified regex.5  Specifying nil for
+  # warn_match means "do not warn".
+  def config_and_warn_on_match(warn_match, *args)
     # Send the entire config as one string but be sure not to return until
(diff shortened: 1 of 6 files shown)


tests/test_snmpuser.rb
```diff
@@ -185,7 +185,7 @@ def test_snmpuser_destroy
   end
 
   def test_snmpuser_auth_password_equal_invalid_param
-    name = 'testV3PwEqualInvalid2'
```
chburket, line 188: removed the "magic string" from the made-up user names.

tests/basetest.rb
```diff
@@ -137,7 +137,23 @@ def teardown
     @device = nil
   end
 
+  # Execute the specified config commands and warn if the
+  # ouput matches the following regex: /invalid|^%/i
```
glennmatthews: s/ouput/output

Also probably not worth including the actual regex in the comment in addition to below - seems likely to become out of sync over time.

tests/basetest.rb
```diff
@@ -137,7 +137,23 @@ def teardown
...
+
+  # Execute the specified config commands. Use this version
+  # of the config method if you expect possible config errors
+  # and do not wish to log them as a warning.
+  def config_no_warn(*args)
+     config_and_warn_on_match(nil, *args)
+  end
+
+  # Execute the specified config commands and warn if the
+  # ouput matches the specified regex.5  Specifying nil for
```
glennmatthews: '5'?

tests/test_bgp_neighbor.rb
```diff
@@ -295,7 +295,7 @@ def test_local_as
   end
 
   def test_log_neighbor_changes
-    if node.product_id[/ios_xr|N(5|6|7)/]
+    if node.product_id[/IOSXR|N(5|6|7)/]
```
glennmatthews: This works for now but it'll be fragile for other XR platforms when we add those. Can we instead do:

`if platform == :ios_xr || node.product_id[N(5|6|7)]`

tests/test_bgp_neighbor_af.rb
```diff
@@ -89,57 +89,72 @@ def setup_ios_xr
...
+    cfg << "router bgp #{asn} address-family #{evpn}"
+    cfg << "router bgp #{asn} vrf #{vrf} rd auto"
+    cfg << "router bgp #{asn} vrf #{vrf} address-family #{af_v4}"
+    cfg << "router bgp #{asn} vrf #{vrf} address-family #{af_v6}"
+    cfg << "router bgp #{asn} vrf #{vrf} address-family #{evpn}"
+    cfg << "router bgp #{asn} vrf #{vrf} neighbor #{nbr} remote-as #{asn}"
+    cfg << "router bgp #{asn} vrf #{vrf} neighbor #{nbr} " \
+           "address-family #{af_v4}"
+
+    config_no_warn(*cfg)
```
glennmatthews: What are the 'expected' warnings here? Can you add a comment documenting this?

chburket:
Pushed changes that address review comments.  Ready for re-review.

tests/test_bgp_neighbor_af.rb
```diff
@@ -89,57 +89,73 @@ def setup_ios_xr
...
+    cfg << "route-policy #{str2}" << 'end-policy'
+    cfg << "route-policy #{str1.reverse}" << 'end-policy'
+    cfg << "route-policy #{str2.reverse}" << 'end-policy'
+    cfg << 'route-policy foo_bar' << 'end-policy'
+    cfg << 'route-policy baz_inga' << 'end-policy'
+    cfg << 'rd-set auto' << 'end-set'
+
+    # If any of the above config already exists, the output will contain
+    # a warning that matches our magic "warning regex" (due to the presence
+    # of a '%'), so just execute without checking for errors.
```
robert-w-gries: > If any of the above config already exists, the output will contain a warning that matches our magic "warning regex"

Is this an issue specific to `route-policy`?

chburket: route-policy and rd-set both seem to warn if they already exist:

% WARNING: Policy object route-policy foo_bar' exists! Reconfiguring it via CLI will replace current definition. Use 'abort to cancel.

tests/basetest.rb
```diff
@@ -137,7 +137,23 @@ def teardown
     @device = nil
   end
 
+  # Execute the specified config commands and warn if the
+  # output matches the default "warning" regex.
   def config(*args)
+    config_and_warn_on_match(/invalid|^%/i, *args)
```
robert-w-gries: Adding an anchor to the `invalid` regexp should help reduce false positives:

``` diff
-    config_and_warn_on_match(/invalid|^%/i, *args)
+    config_and_warn_on_match(/^invalid|^%/i, *args)
```

tests/test_bgp_neighbor_af.rb
glennmatthews: I wouldn't say anything about 'our magic "warning regex"' - the presence of '%' is an IOS convention for warning/error messages. Perhaps just:

``` ruby
    # If any of the above config already exists, we will get a
    # CLI warning message about replacing existing configuration.
    # Ignore these warnings, as they're intentional.
```

tests/test_snmpuser.rb
robert-w-gries, line 188: Nit:  Adding an anchor to the `/invalid/` regexp should remove the need for this change

tests/basetest.rb
chburket: Agreed, but since I don't have a list of all possible error message formats, and I am currently only logging warnings (vs errors), I thought it would be better to have a few false positives than false negatives, until we get more of a chance to go over the warnings.  I'm fine changing it, though, it others agree.

chburket:
New round of changes to review.

tests/basetest.rb
```diff
@@ -137,7 +137,23 @@ def teardown
     @device = nil
   end
 
+  # Execute the specified config commands and warn if the
+  # output matches the default "warning" regex.
   def config(*args)
+    config_and_warn_on_match(/^invalid|^%/i, *args)
```
robert-w-gries, line 143: I see you went with the anchored regexp.  @glennmatthews Any comments on this change?

robert-w-gries:
:+1:

tests/basetest.rb
chburket, line 143: Glenn said he had no strong opinion.  :)

PR #278, opened 2016-03-08
Increased GRPC timeout value for initial connection

jonnytdevops:
This is needed for GRPC to work correctly on our dev machines.


glennmatthews:
:+1:

Note to other reviewers - this is the timeout value used only when initially trying to establish the client connection, in order to fail in a reasonable time if the server is unresponsive. Once the connection is established, we use a timeout of 120 seconds for subsequent requests.

chburket:
:+1:

PR #280, opened 2016-03-09
Add hook - whitespace check for non-Ruby files

glennmatthews:
Adds a trailing whitespace check in the git pre-commit-validate-diffs hook. Whitespace in ruby files is managed by rubocop, so this check explicitly excludes .rb files and only covers other staged files (yaml, md).

Example:

```
glmatthe@fe-ucs36:~/cisco-network-node-utils$ git status
On branch develop
Your branch is up-to-date with 'origin/develop'.
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

        modified:   bin/git/hooks/pre-commit/validate-diffs
        modified:   lib/cisco_node_utils/cmd_ref/acl.yaml
glmatthe@fe-ucs36:~/cisco-network-node-utils$ git commit -m 'Test whitespace change'
*** Running Git hook script: pre-commit-check_unstaged_changes...
        ...[ OK ]
*** Running Git hook script: pre-commit-rubocop...
        Running RuboCop lint checks
warning: parser/current is loading parser/ruby22, which recognizes
warning: 2.2.4-compliant syntax, but you are running 2.2.3.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
        ...[ OK ]
*** Running Git hook script: pre-commit-validate-diffs...
        Checking for non-ASCII filenames
0
        Checking for trailing whitespace
lib/cisco_node_utils/cmd_ref/acl.yaml:27:                      
lib/cisco_node_utils/cmd_ref/acl.yaml:31:   
        ...[ FAIL ] Fix trailing whitespace before committing
```

PR #279, opened 2016-03-08
Add n8k(fretta) excludes + validate_property_excluded? method

mikewiebe:
- Adds N8k where appropriate to the feature yaml `_exclude` list.
- Adds a new method `validate_property_excluded?` that uses the `_exclude` entry as the single source of truth when testing for `nil` and `Cisco::UnsupportedError`.
- Re-ordering of the platforms in `_exclude` in numerical order.

**MiniTests**
- N8k

```
# Running:


Node under test:
  - name  - n9k-157
  - type  - N85-C8508
  - image - bootflash:///n8500-dk9.7.0.3.F1.0.212.bin

TestPortchannelGlobal#test_get_set_resilient = 181.41 s = .
TestPortchannelGlobal#test_get_set_load_defer = 1.50 s = .
TestPortchannelGlobal#test_get_hash_distribution = 1.48 s = .
TestPortchannelGlobal#test_get_set_port_channel_load_balance_hash_poly = 1.46 s = S
TestPortchannelGlobal#test_get_set_port_channel_load_balance_sym_concat_rot = 123.14 s = .
TestPortchannelGlobal#test_get_set_port_channel_load_balance_asym_rot = 1.49 s = S

Finished in 311.791594s, 0.0192 runs/s, 0.0513 assertions/s.

  1) Skipped:
TestPortchannelGlobal#test_get_set_port_channel_load_balance_hash_poly [tests/test_portchannel_global.rb:146]:
Test not supported on this platform


  2) Skipped:
TestPortchannelGlobal#test_get_set_port_channel_load_balance_asym_rot [tests/test_portchannel_global.rb:182]:
Test not supported on this platform

6 runs, 16 assertions, 0 failures, 0 errors, 2 skips
Coverage report generated for MiniTest to /nobackup/mwiebe/REPOS_NODE_UTILS/develop/cisco-network-node-utils/coverage. 1008 / 1949 LOC (51.72%) covered.
```
- N9k

```
Run options: -v -- --seed 42921

# Running:


Node under test:
  - name  - dt-n9k5-1
  - type  - N9K-C9396PX
  - image - bootflash:///nxos.7.0.3.I3.1.bin

TestPortchannelGlobal#test_get_set_load_defer = 0.93 s = .
TestPortchannelGlobal#test_get_set_port_channel_load_balance_hash_poly = 0.89 s = S
TestPortchannelGlobal#test_get_set_port_channel_load_balance_sym_concat_rot = 1.61 s = .
[...]

PR #281, opened 2016-03-09
Add spec checking whitespace in non-Ruby files

glennmatthews:
Counterpart to #280 - add a Rake task to check for whitespace, so that we get validation in Travis-CI as well. This will show a lot of failures - #279 is fixing a number of them but there may be a number left over still.


mikewiebe:
Are you also planning to fix the whitespace issues in this commit or do you want to commit without them?

glennmatthews:
Gonna fix the remaining issues now that #279 has been merged. Thanks for checking.

mikewiebe:
:+1:

robert-w-gries:
:+1:

PR #286, opened 2016-03-09
Corrected minitests

rkorlepa:
Fixed the minitest case scenarios for failing if F3 card was not present in the switch.
Moved compatible_interface? to ciscotest.rb and included it where ever necessary.

This also contains a small fix related to getting all the instances of bridge-domain, it was causing issues in puppet layer so storing all the hash keys in string format rather than in integer.

### DIFF
--- a/lib/cisco_node_utils/bridge_domain.rb
+++ b/lib/cisco_node_utils/bridge_domain.rb
@@ -93,7 +93,7 @@ module Cisco
           bd_list.map { |elem| BridgeDomain.bd_ids_to_array(elem) }
           .flatten.uniq.sort
       final_bd_list.each do |id|
-        hash[id] = BridgeDomain.new(id, false)
+        hash[id.to_s] = BridgeDomain.new(id, false)
       end
       hash
     end
--- a/tests/ciscotest.rb
+++ b/tests/ciscotest.rb
@@ -205,7 +205,7 @@ class CiscoTestCase < TestCase
     end
   end
 
-  def compatible_interface?
+  def mt_full_interface?
     # MT-full tests require a specific linecard; either because they need a
     # compatible interface or simply to enable the features. Either way
     # we will provide an appropriate interface name if the linecard is present.
--- a/tests/ciscotest.rb
+++ b/tests/ciscotest.rb
@@ -204,4 +204,17 @@ class CiscoTestCase < TestCase
       ["default interface #{intf_name}"]
     end
   end
+
+  def compatible_interface?
+    # MT-full tests require a specific linecard; either because they need a
+    # compatible interface or simply to enable the features. Either way
+    # we will provide an appropriate interface name if the linecard is present.
+    # Example 'show mod' output to match against:
+    #   '9  12  10/40 Gbps Ethernet Module  N7K-F312FQ-25 ok'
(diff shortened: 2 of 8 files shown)


lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -93,7 +93,7 @@ def self.bds
           bd_list.map { |elem| BridgeDomain.bd_ids_to_array(elem) }
           .flatten.uniq.sort
       final_bd_list.each do |id|
-        hash[id] = BridgeDomain.new(id, false)
+        hash[id.to_s] = BridgeDomain.new(id, false)
```
mikewiebe, line 96: Why is this change needed?

tests/test_bridge_domain.rb
```diff
@@ -43,11 +43,11 @@ def teardown
   def test_single_bd_create_destroy
     bd = BridgeDomain.new('100', true)
     bds = BridgeDomain.bds
-    assert(bds.key?(100), 'Error: failed to create bridge-domain 100')
+    assert(bds.key?(100.to_s), 'Error: failed to create bridge-domain 100')
```
mikewiebe: Why is this change needed?

tests/test_encapsulation.rb
```diff
@@ -14,6 +14,7 @@
 
 require_relative 'ciscotest'
 require_relative '../lib/cisco_node_utils/encapsulation'
+require_relative '../lib/cisco_node_utils/vdc'
```
mikewiebe, line 17: Do you need to enable any features for these tests to run properly?

tests/ciscotest.rb
```diff
@@ -204,4 +204,17 @@ def get_interface_cleanup_config(intf_name)
       ["default interface #{intf_name}"]
     end
   end
+
+  def compatible_interface?
```
mikewiebe: Please change this method name to `mt_full_interface?`.  This is very specific to mtfull and now that it's not in a standalone file, we need to be more specific with the name.

tests/test_encapsulation.rb
rkorlepa, line 17: Was testing something and forgot to remove it..
Removed it now.

tests/ciscotest.rb
rkorlepa: Changed it as suggested.

tests/test_bridge_domain.rb
rkorlepa: changed it to string instead of .to_s:
assert(bds.key?('100')

lib/cisco_node_utils/bridge_domain.rb
rkorlepa, line 96: Wanted to keep all the return values consistent with other getters in this node util class, all the getter return string or bolean values.

tests/test_bridge_domain.rb
```diff
@@ -43,11 +43,11 @@ def teardown
...
 
     bd.destroy
     bds = BridgeDomain.bds
-    refute(bds.key?(100), 'Error: failed to destroy bridge-domain 100')
+    refute(bds.key?(100.to_s), 'Error: failed to destroy bridge-domain 100')
```
mikewiebe: Did you mean to change these to `100` instead of `100.to_s`?

rkorlepa: Yes forgot to push the changes, have done it now.

PR #290, opened 2016-03-10
Configuration file support for client and tests

glennmatthews:
Add support for global (/etc/cisco_node_utils.yaml) and per-user (~/cisco_node_utils.yaml) configuration files - see `docs/cisco_node_utils.yaml.example` below. This replaces the `$NODE` environment variable previously used in minitest and the gRPC client, and also removes the `ARGV` hack that minitest used as a secondary option. 

Tests are now run as:
`rake test` / `ruby tests/test_foo.rb` -- use the node information defined as `default` in the YAML file(s)
`rake test TESTOPTS='--environment=n7k'` / `ruby tests/test_foo.rb --environment n7k` -- use the node defined as `n7k` in the YAML file(s).

That's right - you can define multiple nodes in your config file and switch between them by name when running tests. :smile: This should make regression testing and CI integration more straightforward.

From a deployment perspective, the expectation is that gRPC (IOS XR) nodes will have as an additional provisioning step the creation of `/etc/cisco_node_utils.yaml`, provisioning this file with the XR username/password to use for management, and marking this file as `chmod 0600` so that only root can read it.

The YAML loading is managed by new class `Cisco::Environment`. I've added an RSpec test for this class, and moved much of the negative testing of client input validation from test_node.rb into spec tests for the various client classes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,21 +15,7 @@ Cisco Network Elements support a rich set of features to make networks robust, e
   * Make changes in your branch.
 * Testing
   * Create a minitest script for any new APIs or new functionality
-  * Run all the tests to ensure there was no collateral damage to existing code. There are two ways you can specify the router or switch (virtual or physical) to test against when running the full test suite:
-    1. Create a YAML config file as `/etc/cisco_node_utils.yaml` or `~/cisco_node_utils.yaml` to specify the host, username, and password. There is an example file in `docs/cisco_node_utils.yaml.example`. You can specify a single test node as `default` in this YAML, or you can describe multiple test hosts and specify the desired host with the `--environment` option:
-
-        ```bash
-        rake test TEST_OPTS='--environment=nxapi_remote'
-        ```
-
-    2. Enter the connection information at runtime:
-
-        ```
-        rake test
-        Enter address or hostname of node under test: 192.168.100.1
-        Enter username for node under test:           user
-        Enter password for node under test:           password
-        ```
+  * Run all the tests to ensure there was no collateral damage to existing code. See [README-test-execution](docs/README-test-execution.md) for details.
 
 * Committing
   * Check for unnecessary whitespace with `git diff --check` before committing.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,12 +15,11 @@ Cisco Network Elements support a rich set of features to make networks robust, e
   * Make changes in your branch.
 * Testing
   * Create a minitest script for any new APIs or new functionality
-  * Run all the tests to ensure there was no collateral damage to existing code. There are two ways you can specify the Nexus switch (virtual or physical) to test against when running the full test suite:
-    1. Use the NODE environment variable to specify the address, username, and password:
+  * Run all the tests to ensure there was no collateral damage to existing code. There are two ways you can specify the router or switch (virtual or physical) to test against when running the full test suite:
+    1. Create a YAML config file as `/etc/cisco_node_utils.yaml` or `~/cisco_node_utils.yaml` to specify the host, username, and password. There is an example file in `docs/cisco_node_utils.yaml.example`. You can specify a single test node as `default` in this YAML, or you can describe multiple test hosts and specify the desired host with the `--environment` option:
 
         ```bash
-        export NODE="192.168.100.1 user password"
-        rake test
+        rake test TEST_OPTS='--environment=nxapi_remote'
         ```
 
     2. Enter the connection information at runtime:
--- a/docs/README-maintainers.md
+++ b/docs/README-maintainers.md
@@ -47,6 +47,7 @@ When we are considering publishing a new release, all of the following steps mus
     - N31xx
     - N56xx
     - N6xxx
+    - N7xxx
     - N9xxx
 
 3. Triage any minitest failures.
--- a/docs/README-maintainers.md
+++ b/docs/README-maintainers.md
@@ -8,7 +8,7 @@ Guidelines for the core maintainers of the cisco-network-node-utils project - ab
 * Does `rubocop` pass? (TODO - this will be part of our CI integration to run automatically)
 * Is `CHANGELOG.md` updated appropriately?
 * Are new minitests added? Do they provide sufficient coverage and consistent results?
-* Do minitests pass on both N9K and N3K?
+* Do minitests pass on all supported platforms
 
 ## Setting up git-flow
 
@@ -45,6 +45,8 @@ When we are considering publishing a new release, all of the following steps mus
   * Platforms (all with latest released software or release candidate)
     - N30xx
     - N31xx
+    - N56xx
+    - N6xxx
     - N9xxx
 
 3. Triage any minitest failures.
--- a/docs/README-test-execution.md
+++ b/docs/README-test-execution.md
@@ -1,0 +1,57 @@
+# Executing the tests provided for this gem
+
+## RSpec tests
+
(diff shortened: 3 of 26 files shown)


docs/README-maintainers.md
```diff
@@ -45,6 +45,8 @@ When we are considering publishing a new release, all of the following steps mus
   * Platforms (all with latest released software or release candidate)
     - N30xx
     - N31xx
+    - N56xx
+    - N6xxx
```
mikewiebe, line 49: Add `N7xxx`

lib/cisco_node_utils/environment.rb
```diff
@@ -0,0 +1,93 @@
...
+require 'yaml'
+require_relative 'logger'
+
+module Cisco
+  # Class representing the configuration environment
+  class Environment
+    @environments = {}
+    @default_environment_name = 'default'
+
+    class << self
```
mikewiebe, line 28: Not familiar with this syntax.  Why do you specify the attr_accessor inside this block?

glennmatthews, line 28: This is how to create a _class_ attr_accessor (`Cisco::Environment.default_environment_name`) instead of an instance attr_accessor (`foo = Cisco::Environment.new ; foo.default_environment_name`)

lib/cisco_node_utils/environment.rb
```diff
@@ -0,0 +1,93 @@
...
+      port:     nil, # only applicable to gRPC
+      username: nil,
+      password: nil,
+    }
+
+    def self.environments
+      if @environments.empty?
+        @environments = merge_config('/etc/cisco_node_utils.yaml',
+                                     @environments)
+        @environments = merge_config('~/cisco_node_utils.yaml',
```
mikewiebe, line 48: By merging this after system-wide it overwrites system right?

glennmatthews, line 48: Well, it overwrites any system values that are redefined by the per-user, yes - I think that's desirable. You can also use this to define _additional_ per-user values (separate from but in addition to the system values) if you want.

lib/cisco_node_utils/environment.rb
HelenHZhang, line 28: good to know this. I was wondering this too.

docs/cisco_node_utils.yaml.example
```diff
@@ -0,0 +1,28 @@
+# The entry labeled 'default' will be used unless another entry is specified
+# default:
+#   host: foo
```
robert-w-gries, line 5: Just to be clear:  If I have a node `default` and three nodes `foo1`, `foo2`, `foo3`, will the command `rake test` still run the minitests against the `default` node?

glennmatthews, line 5: Yes. Would it be clearer if I changed 'specified' to 'selected by the user'?

robert-w-gries, line 5: Yep, that'd be great

CONTRIBUTING.md
```diff
@@ -15,12 +15,11 @@ Cisco Network Elements support a rich set of features to make networks robust, e
...
   * Create a minitest script for any new APIs or new functionality
-  * Run all the tests to ensure there was no collateral damage to existing code. There are two ways you can specify the Nexus switch (virtual or physical) to test against when running the full test suite:
-    1. Use the NODE environment variable to specify the address, username, and password:
+  * Run all the tests to ensure there was no collateral damage to existing code. There are two ways you can specify the router or switch (virtual or physical) to test against when running the full test suite:
+    1. Create a YAML config file as `/etc/cisco_node_utils.yaml` or `~/cisco_node_utils.yaml` to specify the host, username, and password. There is an example file in `docs/cisco_node_utils.yaml.example`. You can specify a single test node as `default` in this YAML, or you can describe multiple test hosts and specify the desired host with the `--environment` option:
 
         ```bash
-        export NODE="192.168.100.1 user password"
-        rake test
+        rake test TEST_OPTS='--environment=nxapi_remote'
```
robert-w-gries: Can you re-word this so that it's clear `default` node will automatically run with `rake test`?

Example:

``` diff
-You can specify a single test node as `default` in this YAML, or you can describe multiple test hosts and specify the desired host with the `--environment` option:
+You can specify a single test node as `default` in this YAML, which will automatically be tested when running `rake test`.  You can also describe multiple test hosts and specify the desired host with the `--environment` option:
```

docs/README-maintainers.md
glennmatthews, line 49: Added.

lib/cisco_node_utils/environment.rb
glennmatthews, line 28: Added a comment for clarity.

docs/cisco_node_utils.yaml.example
glennmatthews, line 5: Updated this - let me know if it's clear now.

CONTRIBUTING.md
glennmatthews: Moved this to new README-test-execution file - please take a look.

mikewiebe:
:+1: Looks good @glennmatthews .  Good workflow enhancement here!

docs/README-test-execution.md
```diff
@@ -0,0 +1,57 @@
...
+```
+
+### Running all RSpec tests
+
+
+```bash
+rake spec
+```
+
+## Minitest tests
```
robert-w-gries, line 22: Nit:  `Minitests` should suffice imo :-)

``` diff
-## Minitest tests
+## Minitests
```

robert-w-gries:
:+1:

There seems to be some automagic going on in the rspec tests that I want to understand, but the PR looks good to merge.  Great addition!

I am curious about extending this enhancement to allow for concurrently running minitests against multiple nodes.  Do you think this is possible?

docs/README-test-execution.md
glennmatthews, line 22: Meh. `minitest` is the name of the test module that we're building on. I'd rather not change this.

glennmatthews:
@robert-w-gries Feel free to ask me for clarifications on the rspec automagic either on here or in person.

To your question about concurrence - it should be possible in theory but I don't have any direct experience with Ruby concurrency and parallelism. Feel free to try it and see what works. :-)

lib/cisco_node_utils/environment.rb
glennmatthews, line 48: @mikewiebe I added some additional spec tests to make this inheritance more explicit - hope it's clearer now.

glennmatthews:
``` bash
$ rspec spec/environment_spec.rb -f doc

Cisco::Environment
  .data_from_file
    handles File.expand_path errors
    handles nonexistent files
    handles unreadable files
    handles YAML errors
  .merge_config
    merges valid content
  .environments
    is empty by default
    loads data from global config if present
    loads data from user config if present
    uses both files if present but user data takes precedence
  .environment
    with no config files available
      returns DEFAULT_ENVIRONMENT when called with no args
      returns DEFAULT_ENVIRONMENT when requested by name as "default"
    with examples in docs/cisco_node_utils.yaml.example
      the "nxapi_local" example
        can be loaded explicitly by name
        can be specified as the default then loaded implicitly
        is valid configuration for the NXAPI client
      the "nxapi_remote" example
        can be loaded explicitly by name
        can be specified as the default then loaded implicitly
        is valid configuration for the NXAPI client
      the "grpc_local" example
        can be loaded explicitly by name
        can be specified as default then loaded implicitly
        is valid configuration for the gRPC client
      the "grpc_remote" example
        can be loaded explicitly by name
        can be specified as default then loaded implicitly
        is valid configuration for the gRPC client

Finished in 0.02712 seconds (files took 0.21237 seconds to load)
23 examples, 0 failures
```

PR #291, opened 2016-03-10
test_vrf/test_vrf_af updates

chrisvanheuveln:
- route_target_both_auto\* and route_target*evpn props require nv overlay evpn support, which we have excluded for n3k
- I2 images require an rd when removing the cli
- Fixed route_target_feature_enable() so that it only attempts to enable the feature command if it's actually supported.
- N5k/N6k vrf shutdown is timing sensitive
- Added feature disablement to test_vrf_af since some platforms can give a false positive if the testbed already has them enabled
- test_vrf & test_vrf_af now pass on all NX + I2

### DIFF
--- a/tests/test_vrf.rb
+++ b/tests/test_vrf.rb
@@ -74,7 +74,7 @@ class TestVrf < CiscoTestCase
   end
 
   # This helper is needed on some platforms to allow enough time for the
-  # 'no shutdown' process to complete before 'shutdown' can be successful.
+  # 'shutdown' process to complete before 'no shutdown' can be successful.
   def shutdown_with_sleep(obj, val)
     obj.shutdown = val
   rescue CliError => e
--- a/tests/test_vrf.rb
+++ b/tests/test_vrf.rb
@@ -73,6 +73,17 @@ class TestVrf < CiscoTestCase
     end
   end
 
+  # This helper is needed on some platforms to allow enough time for the
+  # 'no shutdown' process to complete before 'shutdown' can be successful.
+  def shutdown_with_sleep(obj, val)
+    obj.shutdown = val
+  rescue CliError => e
+    raise unless e.message[/ERROR: Shutdown of VRF .* in progress/]
+    sleep 1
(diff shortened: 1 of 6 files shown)


tests/test_vrf.rb
```diff
@@ -84,11 +84,15 @@ def test_shutdown
     end
     v.shutdown = true
     assert(v.shutdown)
+
+    sleep 20 if node.product_id[/N(5|6)/]
```
glennmatthews: Ouch. Is this a consistent amount of time or would we be better to have a retry loop here?

glennmatthews:
One nit, otherwise :+1:

tests/test_vrf.rb
chrisvanheuveln: Ouch is right. It's consistently long but I haven't narrowed it down to a time length. A retry loop would be nice but what would you test against to break out? Calling the setter will raise a CliError...

glennmatthews: Yeah you'd have to try it, rescue CliError, sleep a few, try again until either it succeeds or you exceed a maximum retry count.

chrisvanheuveln: Cool. I didn't realize I could `retry` after a rescue.

New diff coming...

tests/test_vrf_af.rb
```diff
@@ -118,7 +132,7 @@ def route_target(af)
     refute(v.default_route_target_both_auto_evpn,
            'default value for route target both auto evpn should be false')
 
-    if platform == :ios_xr
+    if platform == :ios_xr || node.product_id[/N3/]
```
mikewiebe: Please update these to use the `validate_property_excluded?(feature, property)` api.

https://github.com/cisco/cisco-network-node-utils/blob/develop/tests/test_vrf.rb#L78

chrisvanheuveln: done

mikewiebe:
:+1:

PR #292, opened 2016-03-11
Fixed issue in radius_server

jonnytdevops:
If someone where to set key => unset on a new (i.e. not already existing) resource, then an error would occur. This PR fixes that.


glennmatthews:
:+1:

robert-w-gries:
We should add/update a minitest in order to validate this case since the minitests were all passing before this fix

jonnytdevops:
@robert-w-gries Simple test added

chburket:
:+1:

robert-w-gries:
:+1:

PR #294, opened 2016-03-11
test_vxlan_vtep platform fixes

chrisvanheuveln:
- vxlan_vtep uses the interface class, which was converted at some point to use key-val yaml; vxlan_vtep needed to be updated as well
- Added appropriate excludes and unsupported checks


robert-w-gries:
:+1:

PR #295, opened 2016-03-11
test_vxlan_vtep_vni platform fixes

chrisvanheuveln:
- <vni> needs to be optional or else all_vnis fails
- no ingress_replication support on N7k, which means no peer-list either
- Some of the setters need to remove ingress-replication when setting other commands; however, they assumed ingress-replication was supported everywhere, so I added add'l checks for that.
- minitest now passes on all NX + I2 (yes, checked N8k as well). XR is n/a.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/vxlan_vtep_vni.yaml
+++ b/lib/cisco_node_utils/cmd_ref/vxlan_vtep_vni.yaml
@@ -6,16 +6,17 @@ _template:
   get_command: "show running-config interface all | section 'interface nve'"
   get_context:
     - '/^interface <name>$/i'
-    - '/^member vni <vni> ?(associate-vrf)?$/'
+    - '(?)/^member vni <vni> ?(associate-vrf)?$/'
   set_context:
     - 'interface <name>'
-    - 'member vni <vni> <assoc_vrf>'
+    - '(?)member vni <vni> <assoc_vrf>'
 
 all_vnis:
   multiple:
   get_value: '/^member vni (\d+|\d+-\d+) ?(associate-vrf)?$/'
 
 ingress_replication:
+  _exclude: [N7k]
   kind: string
   get_value: '/^ingress-replication protocol (\S+)$/'
   set_value: '<state> ingress-replication protocol <protocol>'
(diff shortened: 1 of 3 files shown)


lib/cisco_node_utils/cmd_ref/vxlan_vtep_vni.yaml
```diff
@@ -6,16 +6,17 @@ _template:
   get_command: "show running-config interface all | section 'interface nve'"
   get_context:
     - '/^interface <name>$/i'
-    - '/^member vni <vni> ?(associate-vrf)?$/'
+    - '(?)/^member vni <vni> ?(associate-vrf)?$/'
```
robert-w-gries, line 9: Can you remind me what the `(?)` does?

lib/cisco_node_utils/vxlan_vtep_vni.rb
```diff
@@ -104,12 +104,16 @@ def destroy
...
+      node.cmd_ref.supports?('vxlan_vtep_vni', 'ingress_replication')
+    end
+
     def ingress_replication
       config_get('vxlan_vtep_vni', 'ingress_replication', @get_args)
     end
 
     def remove_add_ingress_replication(protocol)
-      if ingress_replication.empty?
+      if ingress_replication.to_s.empty?
```
robert-w-gries, line 118: Since `ingress_replication` has `kind: string` in the YAML entry, there should be no need for converting the result to a string

lib/cisco_node_utils/cmd_ref/vxlan_vtep_vni.yaml
chrisvanheuveln, line 9: It makes the line optional. The test was failing during the `all_vnis` call which doesn't include a vni -- since it's trying to collect all vnis.

lib/cisco_node_utils/vxlan_vtep_vni.rb
chrisvanheuveln, line 118: This needed a nil check (N7k doesn't support ingress-replication so the getter will return nil). 
I used this approach to handle both the nil and empty string checks. `Nil.to_s` results in `''`.

robert-w-gries, line 118: Can you add a comment explaining that it's handling an N7k edge case?

chrisvanheuveln:
@robert-w-gries merge?

robert-w-gries:
:+1:

PR #296, opened 2016-03-14
Change RuntimeError to CliError

rkorlepa:
Issue seen on Chris's setup are all valid ones or should throw an error as they are negative testcases. The issue here was it was throwing CliError instead of RuntimeError, which was changed due to the fix from @glennmatthews to automatically wrap these errors into CliError #266

### DIFF
--- a/tests/ciscotest.rb
+++ b/tests/ciscotest.rb
@@ -18,6 +18,8 @@ require_relative 'basetest'
 require_relative 'platform_info'
 require_relative '../lib/cisco_node_utils/interface'
 require_relative '../lib/cisco_node_utils/node'
+require_relative '../lib/cisco_node_utils/vlan'
+require_relative '../lib/cisco_node_utils/bridge_domain'
 
 include Cisco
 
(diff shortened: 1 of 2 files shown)


mikewiebe:
Please fix the rubocop errors.

tests/test_bridge_domain.rb
```diff
@@ -24,6 +24,11 @@ class TestBridgeDomain < CiscoTestCase
   @@cleaned = false # rubocop:disable Style/ClassVars
 
   def cleanup
+    Vlan.vlans.each do |vlan, obj|
+      # skip reserved vlans
+      next if vlan == '1'
+      obj.destroy
+    end
```
chrisvanheuveln: This is probably common enough code that it can live in `ciscotest.rb` as `remove_all_vlans`, _a la_ 
https://github.com/cisco/cisco-network-node-utils/blob/develop/tests/ciscotest.rb#L188

rkorlepa: Moved the code to common location and also created another remove_all_bridge_domains section for bds

tests/ciscotest.rb
```diff
@@ -221,4 +223,22 @@ def mt_full_interface?
 
     "ethernet#{slot}/1"
   end
+
+  def remove_all_vlans
```
chrisvanheuveln: Nit: please move your new remove_all_\* methods up to where the other remove_all_\* methods live and keep them in alpha order. We will likely have a few more of these remove_all's coming...

:+1:  on the rest of your changes.

rkorlepa: Done changed the order.

PR #301, opened 2016-03-16
Interface test fixes for n9k

robert-w-gries:
## Changes
- Client now raises 4xx and 5xx error codes
  - Error 432 is ignored because it only signifies lack of configured property
- `test_interface.rb` now uses teardown method to clean `interfaces[0]`
- Refactored `interface_ethernet_default` to accept `interfaces[0]` as input instead of `interfaces_id[0]`
### I2

``` bash
  1) Skipped:
TestInterfaceServiceVni#all_skipped [/home/robgrie/cisco-network-node-utils/tests/ciscotest.rb:56]:
Skipping TestInterfaceServiceVni; feature 'interface_service_vni' is unsupported on this node
```
### I3
- Known platform bug in `test_negotiate_auto_portchannel`
- Likely platform bug in `test_negotiate_auto_ethernet`

``` bash

  1) Failure:
TestInterface#test_negotiate_auto_ethernet [/home/robgrie/cisco-network-node-utils/tests/test_interface.rb:774]:
[interface ethernet1/1] The command 'no negotiate auto' was rejected with error:
ERROR: cannot turn off auto-negotiate with auto-negotiate speeds

  2) Failure:
TestInterface#test_negotiate_auto_portchannel [/home/robgrie/cisco-network-node-utils/tests/test_interface.rb:685]:
Error: port-channel10 negotiate auto value not false.
Expected: false
  Actual: true

  1) Skipped:
TestInterfaceServiceVni#all_skipped [/home/robgrie/cisco-network-node-utils/tests/ciscotest.rb:56]:
Skipping TestInterfaceServiceVni; feature 'interface_service_vni' is unsupported on this node
```

### DIFF
--- a/tests/test_interface.rb
+++ b/tests/test_interface.rb
@@ -34,6 +34,7 @@ class TestInterface < CiscoTestCase
 
   def setup
     super
+    interface_ethernet_default(interfaces[0])
     if platform == :nexus
       @port_channel = 'port-channel'
       # rubocop:disable Style/AlignHash
@@ -476,7 +477,6 @@ class TestInterface < CiscoTestCase
     description = 'This is a test description ! '
     interface.description = description
     assert_equal(description.rstrip, interface.description)
-    interface_ethernet_default(interfaces[0])
   end
 
   def test_encapsulation_dot1q
@@ -489,7 +489,6 @@ class TestInterface < CiscoTestCase
     subif.encapsulation_dot1q = 25
     assert_equal(25, subif.encapsulation_dot1q)
     subif.destroy
-    interface_ethernet_default(interfaces[0])
   end
 
   def test_interface_mtu_change
@@ -501,7 +500,6 @@ class TestInterface < CiscoTestCase
     assert_equal(1580, interface.mtu)
     interface.mtu = interface.default_mtu
     assert_equal(interface.default_mtu, interface.mtu)
-    interface_ethernet_default(interfaces[0])
   end
 
   def test_interface_mtu_invalid
@@ -517,7 +515,6 @@ class TestInterface < CiscoTestCase
     assert_equal(1550, interface.mtu)
     interface.mtu = interface.default_mtu
     assert_equal(interface.default_mtu, interface.mtu)
-    interface_ethernet_default(interfaces[0])
   end
 
   def test_mtu_invalid_loopback
@@ -546,7 +543,6 @@ class TestInterface < CiscoTestCase
         speed_change_disallowed(e.message)
       end
     end
-    interface_ethernet_default(interfaces[0])
   end
 
   def test_duplex
@@ -569,7 +565,6 @@ class TestInterface < CiscoTestCase
         speed_change_disallowed(e.message)
       end
     end
-    interface_ethernet_default(interfaces[0])
   end
 
   def test_interface_shutdown_valid
@@ -579,7 +574,6 @@ class TestInterface < CiscoTestCase
 
     interface.shutdown = false
     refute(interface.shutdown, 'Error: shutdown state is not false')
-    interface_ethernet_default(interfaces[0])
   end
 
   def test_svi_prop_nil_when_ethernet
@@ -595,7 +589,6 @@ class TestInterface < CiscoTestCase
   #     interface.switchport_mode = :access
   #     addresses = interface.prefixes
   #     assert_empty(addresses)
-  #     interface_ethernet_default(interfaces[0])
   #   end
   #
   #   def test_interface_get_prefix_list_with_ipv4_address_assignment
@@ -609,7 +602,6 @@ class TestInterface < CiscoTestCase
   #     assert(prefixes.has_key?("192.168.1.100"))
   #     interface.switchport_mode = :access
   #     prefixes = nil
-  #     interface_ethernet_default(interfaces[0])
   #   end
   #
   #   def test_interface_get_prefix_list_with_ipv6_address_assignment
@@ -623,7 +615,6 @@ class TestInterface < CiscoTestCase
   #     assert(prefixes.has_key?("fd56:31f7:e4ad:5585::1"))
   #     interface.switchport_mode = :access
   #     prefixes = nil
-  #     interface_ethernet_default(interfaces[0])
   #   end
   #
   #   def test_interface_prefix_list_with_both_ip4_and_ipv6_address_assignments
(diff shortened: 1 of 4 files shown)


mikewiebe:
:+1:

tests/test_interface.rb
```diff
@@ -1444,6 +1417,7 @@ def test_ipv4_pim_sparse_mode
     #   interface Ethernet1/1
     #     ip pim sparse-mode
     #
+    i.switchport_mode = :disabled
```
chrisvanheuveln: Did you test against all platforms? This minitest has a ton of corner cases thanks to all the linecard peculiarities...

In particular I think this one will fail on my N7k when it's using the default module-types (ie, non-F3 cards), e.g.

```
n7k8(config-if)# int e1/1
n7k8(config-if)# no switchport 
ERROR: Ethernet1/1: requested config change not allowed   <---------------<< !!!!

n7k8(config-if)# sh run vdc
vdc n7k8 id 1
  limit-resource module-type m1 m1xl m2xl f2e 
  allocate interface Ethernet1/1-48
```

The reason is that this e1/1 is an L2-only interface so you can't disable switchport-mode.  

This is probably a bigger problem for us to solve than in this PR but for now can you add a rescue/skip to catch this condition?

:+1: for everything else

tests/test_interface.rb
```diff
@@ -1350,20 +1325,18 @@ def test_interface_ipv4_all_interfaces
     rescue Minitest::Assertion
       # clean up before failing
       config(*cfg)
-      InterfaceChannelGroup.new(interfaces[1]).channel_group = false
+      interface_ethernet_default(interfaces[1])
```
glennmatthews, line 1328: Did you check this change on XR? IIRC this code was needed due to an issue where 'default interface Gig0/0/0/0' wasn't removing 'bundle id 10' from the interface.

tests/test_interface.rb
```diff
@@ -1350,20 +1325,18 @@ def test_interface_ipv4_all_interfaces
...
 
   def test_interface_vrf_default
-    config('interface loopback1', 'vrf member foo')
     interface = Interface.new('loopback1')
     interface.vrf = interface.default_vrf
     assert_equal(DEFAULT_IF_VRF, interface.vrf)
   end
 
   def test_interface_vrf_empty
-    config('interface loopback1', 'vrf member foo')
```
glennmatthews, line 1366: With this change to both test cases, you're no longer testing the case where an interface is in a VRF and you're setting it back to the default, right? That seems like a gap.

tests/test_interface.rb
robert-w-gries: The test seems to pass for n7k:

``` bash
TestInterface#test_ipv4_pim_sparse_mode = 10.42 s = .
```

`interfaces[0]` grabs `eth3/1` as the interface to test against, so I'm not sure if we've coded around this issue or if my node just happens to have compatible interfaces.

tests/test_interface.rb
robert-w-gries, line 1366: I think combining the two tests into one should resolve this and remove the redundancy of the tests:
1. Create loopback interface
2. Assert vrf is empty
3. Set vrf to `foo`
4. Assert vrf value is `foo`
5. Set vrf to default value
6. Assert vrf is default value
7. Cleanup interface

tests/test_interface.rb
chrisvanheuveln: Try my n7k8 node.

robert-w-gries: Test now skips if it hits that edge case:

``` bash
TestInterface#test_ipv4_pim_sparse_mode = 13.69 s = S

Finished in 13.688575s, 0.0731 runs/s, 0.0000 assertions/s.

  1) Skipped:
TestInterface#test_ipv4_pim_sparse_mode [tests/test_interface.rb:1420]:
Interface does not support switchport disable
```

chrisvanheuveln:
:+1:

glennmatthews:
:+1:

tests/test_interface.rb
robert-w-gries, line 1328: Looks like `default interface Gig0/0/0/0` deletes the interface, but `bundle id 10` doesn't stick around afterwards.

``` bash
RP/0/RP0/CPU0:BigSunstone-for-Puppet-278-suns(config)#show running-config
[snip]
interface Loopback1
!
interface MgmtEth0/RP0/CPU0/0
 ipv4 address 192.168.1.7 255.255.255.0
!
interface GigabitEthernet0/0/0/0
 bundle id 10 mode on
!
interface GigabitEthernet0/0/0/1
 shutdown
!
interface GigabitEthernet0/0/0/2
 shutdown
!
! Configure default route.
router static
 address-family ipv4 unicast
  0.0.0.0/0 192.168.1.1
 !
!

RP/0/RP0/CPU0:BigSunstone-for-Puppet-278-suns(config)#default interface GigabitEthernet0/0/0/0
RP/0/RP0/CPU0:BigSunstone-for-Puppet-278-suns(config)#commit

RP/0/RP0/CPU0:BigSunstone-for-Puppet-278-suns(config)#show running-config
[snip]
interface Loopback1
!
interface MgmtEth0/RP0/CPU0/0
 ipv4 address 192.168.1.7 255.255.255.0
!
interface GigabitEthernet0/0/0/1
 shutdown
!
interface GigabitEthernet0/0/0/2
 shutdown
!
! Configure default route.
router static
 address-family ipv4 unicast
  0.0.0.0/0 192.168.1.1
 !
!

RP/0/RP0/CPU0:BigSunstone-for-Puppet-278-suns(config)#interface GigabitEthernet0/0/0/0
RP/0/RP0/CPU0:BigSunstone-for-Puppet-278-s(config-if)#shutdown
RP/0/RP0/CPU0:BigSunstone-for-Puppet-278-s(config-if)#commit
RP/0/RP0/CPU0:BigSunstone-for-Puppet-278-suns(config)#show running-config
[snip]
interface Loopback1
!
interface MgmtEth0/RP0/CPU0/0
 ipv4 address 192.168.1.7 255.255.255.0
!
interface GigabitEthernet0/0/0/0
 shutdown
!
interface GigabitEthernet0/0/0/1
 shutdown
!
interface GigabitEthernet0/0/0/2
 shutdown
!
! Configure default route.
router static
 address-family ipv4 unicast
  0.0.0.0/0 192.168.1.1
 !
!
```

PR #302, opened 2016-03-17
Feature/vxlan bridge domain

rkorlepa:
Seperate out member_vni from bridge_domain node util and create a new one which can handle bd range, as member vni is done on a range.And also move the common functions to utils class.
This also contains the changes to the yaml file, before the getter for name and state was using "show bridge-domain", have changed it to "show run bridge-domain" a more global cli for better parsing in other platforms.
This also contains changing the setter for encap profile dot1q, before it was taking in two arguments so moved it to a list.


lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -139,62 +139,13 @@ def default_fabric_control
...
-    end
-
     # Bridge-Domain name assigning case
     # bridge-domain 100
     #   name PepsiCo
     def bd_name
-      config_get('bridge_domain', 'name', bd: @bd_ids)
+      res = config_get('bridge_domain', 'name', bd: @bd_ids)
+      return default_bd_name unless res[2]
+      res[2]
```
chrisvanheuveln: These references look wrong. It looks like you're using the yaml template incorrectly. You should be using the template to get into the proper cli context but your individual properties should have their own get_value's; e.g.

```
# bridge-domain 100
#   shutdown
#   fabric-control
#   name PepsiCo
_template:
  get_command: "show running-config bridge-domain"
-  get_value: '/^bridge-domain[\s\S]*bridge-domain <bd>\s+(shutdown)?\s*(fabric-control)?\s*(?:name (\S+)|\s*)/'
+ context: '^bridge-domain <bd>'

name:
    get_value: "name (.*)"
    set_value: "<state> name <name>"
    default_value: ~
```

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -139,62 +139,13 @@ def default_fabric_control
...
-    def default_member_vni
-      config_get_default('bridge_domain', 'member_vni')
-    end
-
     # Bridge-Domain name assigning case
     # bridge-domain 100
     #   name PepsiCo
     def bd_name
-      config_get('bridge_domain', 'name', bd: @bd_ids)
+      res = config_get('bridge_domain', 'name', bd: @bd_ids)
```
chrisvanheuveln: Use the same name in the yaml that you use for the getter/setter; e.g.

`config_get('bridge_domain', 'bd_name', bd: @bd_ids)`

lib/cisco_node_utils/bridge_domain_vni.rb
```diff
@@ -0,0 +1,165 @@
...
+
+      @bd_ids_list = Utils.string_to_array(@bd_ids)
+      create if instantiate
+    end
+
+    def to_s
+      "Bridge Domain #{bd_ids}"
+    end
+
+    def self.rangebds
```
chrisvanheuveln: Use snakecase here: `self.range_bds`  

Consider renaming it to follow the convention we use in our other providers: `self.all_bds` or `self.all_vnis` or `self.all_member_vnis`... whatever this applies to.

Please add some comments to this method and the next. It's not obvious what these are for without digging into the code.

lib/cisco_node_utils/bridge_domain_vni.rb
```diff
@@ -0,0 +1,165 @@
...
+      end
+      hash
+    end
+
+    def curr_bd_vni_hash
+      final_bd_vni = {}
+      curr_vni = config_get('bridge_domain_vni', 'member_vni')
+      curr_bd_vni = config_get('bridge_domain_vni', 'member_vni_bd')
+      return final_bd_vni if curr_vni.empty? || curr_bd_vni.empty?
+
```
chrisvanheuveln, line 93: Could you add some sample cli in a comment here? It would help to understand what your inputs look like.

lib/cisco_node_utils/bridge_domain_vni.rb
```diff
@@ -0,0 +1,165 @@
...
+      curr_vni_list = Utils.string_to_array(curr_vni)
+      curr_bd_vni_list = Utils.string_to_array(curr_bd_vni)
+
+      hash_map = Hash[curr_bd_vni_list.zip(curr_vni_list.map)]
+      @bd_ids_list.each do |bd|
+        final_bd_vni[bd.to_i] = hash_map[bd.to_i] if hash_map.key?(bd.to_i)
+      end
```
chrisvanheuveln, line 100: And also a sample hash entry.

lib/cisco_node_utils/bridge_domain_vni.rb
```diff
@@ -0,0 +1,165 @@
...
+    # This function will first add bds to the system bridge-domain and then
+    # create the bds. If bds already existing then just create. Else add the
+    # non added bds to system range first then create all. This is to avoid the
+    # idempotency issue as system add command throws error if a bd is already
+    # present in the system range.
+    def create
+      sys_bds_array = Utils.string_to_array(system_bd_add)
+      if (@bd_ids_list - sys_bds_array).any?
+        add_bds = Utils
+                  .unsorted_list_to_string(@bd_ids_list - sys_bds_array)
```
chrisvanheuveln: you shouldn't need this wrap

lib/cisco_node_utils/bridge_domain_vni.rb
```diff
@@ -0,0 +1,165 @@
...
+        config_set('bridge_domain_vni', 'system_bd_add', addbd: add_bds)
```
chrisvanheuveln: It's not clear to me how this will work when you have multiple ranges, i.e. if I currently have bds `10-20` but create wants to have `5,15-20,25`, what will `add_bds` be adding?

lib/cisco_node_utils/cmd_ref/bridge_domain_vni.yaml
```diff
@@ -0,0 +1,28 @@
+# bridge_domain
+# bridge_domain feature is available only on n7k
+---
+_exclude: [N3k, N5k, N6k, N9k, ios_xr]
```
chrisvanheuveln: exclude `N8k` as well

lib/cisco_node_utils/cmd_ref/bridge_domain_vni.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+create:
+  set_value: "bridge-domain <crbd>"
```
chrisvanheuveln: Nit: I know variable naming is subjective but it's easier on the maintainers if you're consistent; i.e. `/crbd/bd/`

lib/cisco_node_utils/bridge_domain_vni.rb
```diff
@@ -0,0 +1,165 @@
...
+      end
+      config_set('bridge_domain_vni', 'create', crbd: @bd_ids)
```
chrisvanheuveln: Will the `vni` cli be available at this stage if `feature vni` is not enabled yet?

lib/cisco_node_utils/cmd_ref/bridge_domain_vni.yaml
```diff
@@ -0,0 +1,28 @@
+# bridge_domain
+# bridge_domain feature is available only on n7k
+---
+_exclude: [N3k, N5k, N6k, N9k, ios_xr]
+
```
chrisvanheuveln, line 5: Use a template with these yaml entries. Something like:

```
_template:
  get_command: "show running-config bridge-domain"
  context:
    - "(?)bridge-domain <bd>"
```

lib/cisco_node_utils/cmd_ref/bridge_domain_vni.yaml
```diff
@@ -0,0 +1,28 @@
...
+create:
+  set_value: "bridge-domain <crbd>"
+
+member_vni:
+  get_command: "show running-config bridge-domain | inc 'member vni'"
+  get_value: '/^  member vni (.*)/'
+  set_value: "<vnistate> vni <vni> ; bridge-domain <bd> ; <membstate> member vni <membvni>"
```
chrisvanheuveln: Confused.  `<vnistate> vni <vni>` is an independent command from the others, isn't it?

I expect the config to be something like:

```
n7k8(config)# vni 4100 ; bridge-dom 41 ; member vni 4100
n7k8(config-bdomain)# sh run | sec brid|vni
system bridge-domain 41 
vni 4100
bridge-domain 41
  member vni 4100
```

I suggest you create a separate top-level `vni:` entry instead. It just means you call `config_set` twice when setting these values.

```
member_vni:
  get_value: '/^member vni (\d+)/'
  set_value: "<state> member vni <vni>"

vni:
  get_value: '/^vni (\d+)/'
  set_value: "<state> vni <vni> "
```

lib/cisco_node_utils/cmd_ref/bridge_domain_vni.yaml
```diff
@@ -0,0 +1,28 @@
...
+  default_value: ''
+
+member_vni_bd:
+  get_command: "show running-config bridge-domain | inc 'member vni' prev 1"
+  get_value: '/^bridge-domain (.*)/'
```
chrisvanheuveln, line 17: Use template

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -139,62 +139,13 @@ def default_fabric_control
...
-    end
-
-    def default_member_vni
-      config_get_default('bridge_domain', 'member_vni')
-    end
-
     # Bridge-Domain name assigning case
     # bridge-domain 100
     #   name PepsiCo
     def bd_name
```
chrisvanheuveln: Please keep the property methods in alpha order (when it makes sense).

lib/cisco_node_utils/bridge_domain_vni.rb
```diff
@@ -0,0 +1,165 @@
...
+                   bd: @bd_ids, membstate: '', membvni: val)
+      end
+    end
+
+    def default_member_vni
+      config_get_default('bridge_domain_vni', 'member_vni')
+    end
+
+    # getter for system bridge-domain
+    def system_bd_add
```
chrisvanheuveln: Confusing name. I would just use `system_bridge_domain`. 

The `add` / `remove` keywords can be passed as variables if needed.

lib/cisco_node_utils/cisco_cmn_utils.rb
```diff
@@ -166,6 +166,53 @@ def self.zero_pad_macaddr(mac)
       sprintf('%04x.%04x.%04x', o1, o2, o3)
     end
 
+    # This will expand the string to a list of bds as integers
+    def self.string_to_array(string)
```
chrisvanheuveln: As mentioned in the puppet review, use `range_summarize` instead of these two new methods, if possible.

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
```diff
@@ -3,6 +3,10 @@
 ---
 _exclude: [N3k, N5k, N6k, N8k, N9k, ios_xr]
 
+_template:
+  get_command: "show running-config bridge-domain"
+  get_value: '/^bridge-domain[\s\S]*bridge-domain <bd>\s+(shutdown)?\s*(fabric-control)?\s*(?:name (\S+)|\s*)/'
```
chrisvanheuveln: See my earlier comment about misusing your template (https://github.com/cisco/cisco-network-node-utils/pull/302/files#diff-dcbfe5a528545ac12dacc955d843bae3R148)

Use something like this instead:

```
_template:
  get_command: "show running-config bridge-domain"
-  get_value: '/^bridge-domain[\s\S]*bridge-domain <bd>\s+(shutdown)?\s*(fabric-control)?\s*(?:name (\S+)|\s*)/'
+ context: '^bridge-domain <bd>'
```

lib/cisco_node_utils/cmd_ref/encapsulation.yaml
```diff
@@ -22,4 +22,4 @@ dot1q_map:
   get_value: '/^.* vni <profile>\s+dot1q ([\d\s,-]+) vni ([\d\s,-]+)$/'
   set_context: ['encapsulation profile vni <profile>']
   set_value: '<state> dot1q <vlans> vni <vnis>'
-  default_value: ~
+  default_value: ''
```
chrisvanheuveln: Why this change? `nil` seems like the correct usage in this case.

lib/cisco_node_utils/cmd_ref/bridge_domain_vni.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+range_bds:
+  multiple: true
+  get_command: "show running-config bridge-domain | inc 'member vni' prev 1"
+  get_value: '/^bridge-domain (.*)/'
+
+system_bd_add:
+  get_command: "show running-config bridge-domain"
+  get_value: '/^system bridge-domain (\S+)/'
+  set_value: "system bridge-domain add <addbd>"
```
chrisvanheuveln: I don't see a remove for this but I suspect you need that as well. I would just make the `add` variable. Also note prev comments about name change and using a template.

```
-system_bd_add:
+system_bridge_domain:
-  get_command: "show running-config bridge-domain"
  get_value: '/^system bridge-domain (\S+)/'
-  set_value: "system bridge-domain add <addbd>"
+  set_value: "system bridge-domain <action> <bd>"
```

lib/cisco_node_utils/encapsulation.rb
```diff
@@ -62,14 +63,23 @@ def destroy
     # PROPERTIES
     # ----------
 
+    def batch_the_string(string)
```
chrisvanheuveln: Use `range_summarize`

lib/cisco_node_utils/encapsulation.rb
```diff
@@ -62,14 +63,23 @@ def destroy
...
+      Utils.unsorted_list_to_string(Utils.string_to_array(string.to_s))
+    end
+
     def dot1q_map
-      config_get('encapsulation', 'dot1q_map', profile: @encap_name)
+      result = config_get('encapsulation', 'dot1q_map', profile: @encap_name)
+      return '' unless result != ''
```
chrisvanheuveln:
```
-      return '' unless result != ''
+      return default_dot1q_map if result.to_s.empty?    # to_s handles nil case too
```

tests/test_bridge_domain.rb
```diff
@@ -223,4 +151,41 @@ def test_invalid_bd_create
       BridgeDomain.new('90, 5000-5004,100')
     end
   end
+
```
chrisvanheuveln, line 154: I think it may be good to disable the feature in the `cleanup` method: https://github.com/cisco/cisco-network-node-utils/pull/302/files#diff-65aa1d6fa6431ccd35936e979551261cR26

```
def cleanup
  config_no_warn("no feature vni")
  ...
```

tests/test_bridge_domain_vni.rb
```diff
@@ -0,0 +1,112 @@
...
+require_relative '../lib/cisco_node_utils/bridge_domain_vni'
+
+include Cisco
+
+# TestBridgeDomainVNI - Minitest for bridge domain class.
+class TestBridgeDomainVNI < CiscoTestCase
+  @skip_unless_supported = 'bridge_domain_vni'
+  @@cleaned = false # rubocop:disable Style/ClassVars
+
+  def cleanup
```
chrisvanheuveln, line 25: Disable feature vni as well.

tests/test_bridge_domain_vni.rb
```diff
@@ -0,0 +1,112 @@
...
+    super
+    cleanup unless @@cleaned
+    @@cleaned = true # rubocop:disable Style/ClassVars
+  end
+
+  def teardown
+    cleanup unless @@cleaned
+    super
+    cleanup
+  end
```
chrisvanheuveln, line 43: This is wrong. You want the cleanup to occur before the first test and after every test. Also, the `super` call should occur at the end of teardown (as opposed to setup which needs it to be first)

```
  def teardown
-    cleanup unless @@cleaned
    cleanup
    super
  end
```

tests/test_bridge_domain_vni.rb
```diff
@@ -0,0 +1,112 @@
...
+    bd.destroy
+  end
+
+  def test_multiple_bd_member_vni
+    mt_full_interface?
+    bd = BridgeDomainVNI.new('100-110,150,170-171')
+    assert_equal(bd.default_member_vni, bd.member_vni,
+                 'Error: Bridge-Domain is mapped to different vnis')
+
+    vni = '6000-6010,6050,5070-5071'
```
chrisvanheuveln, line 69: Nit: You could add some whitespace just to ensure that your whitespace cleanup is working.

tests/test_encapsulation.rb
```diff
@@ -56,11 +56,15 @@ def test_encapsulation_dot1_mapping
                  'Error: dot1q is not matching')
     dot1q = '100-110'
     vni = '5000-5010'
-    encap.send(:set_dot1q_map=, true, dot1q, vni)
+    encap.dot1q_map = [dot1q, vni]
     assert_equal(dot1q, encap.dot1q_map[0].gsub(/\s+/, ''),
```
chrisvanheuveln: This test tells me that the getter is not removing whitespace. It should be normalizing the state. Please fix the getter and change these tests so that they just use the raw getter result.

lib/cisco_node_utils/bridge_domain.rb
rkorlepa: Done changed

lib/cisco_node_utils/bridge_domain_vni.rb
rkorlepa: Done changed

lib/cisco_node_utils/bridge_domain_vni.rb
rkorlepa, line 93: Added

lib/cisco_node_utils/bridge_domain_vni.rb
rkorlepa, line 100: Added

lib/cisco_node_utils/bridge_domain_vni.rb
rkorlepa: bd create will do a create of all the bds passed by the user. This check is for doing system bridge-domain add <bds>, if we add already existing bds then cli throws out a warning on the n7k device which is caught as a CliError here, so to avoid this I first subtract already existing system bridge-domain range and then apply addbd.

So in ur case existing is 10-20 and going to create is 5,15-20,25. So 15-20 is already existing in system bridge-domain so skip this range and do add for 5,25 and bd create for 5,15-20,25. bd create will be a noop if its already exiting in the system.

lib/cisco_node_utils/cmd_ref/bridge_domain_vni.yaml
rkorlepa: added

lib/cisco_node_utils/cmd_ref/bridge_domain_vni.yaml
rkorlepa: changed

lib/cisco_node_utils/bridge_domain_vni.rb
rkorlepa: No during create of bds there is no need for feature vni or vni cli. Its required only during member_vni for which there is a feature enable.
And bridge-domain is not dependent on vni, it can be created any time by the user.

lib/cisco_node_utils/cmd_ref/bridge_domain_vni.yaml
rkorlepa: Changed

lib/cisco_node_utils/bridge_domain_vni.rb
rkorlepa: Changed

lib/cisco_node_utils/cisco_cmn_utils.rb
rkorlepa: I need these two methods string_to_array and unsorted_list_to_array because they are extensively used in bridge_domain_vni for data manipulation.

lib/cisco_node_utils/cmd_ref/bridge_domain_vni.yaml
rkorlepa: Changed

lib/cisco_node_utils/encapsulation.rb
rkorlepa: Changed

lib/cisco_node_utils/encapsulation.rb
rkorlepa: Changed

lib/cisco_node_utils/cmd_ref/encapsulation.yaml
rkorlepa: Changed back to nil was testing something and forgot to remove it.

tests/test_bridge_domain.rb
rkorlepa, line 154: Added

tests/test_bridge_domain_vni.rb
rkorlepa, line 25: Added

tests/test_bridge_domain_vni.rb
rkorlepa, line 43: Changed

tests/test_encapsulation.rb
rkorlepa: Done changed. 
No need of removing spaces here the getter already does that

tests/test_bridge_domain_vni.rb
rkorlepa, line 69: Added

lib/cisco_node_utils/bridge_domain.rb
rkorlepa: Changed it. Removed the get_value and moved them accordingly to their respective properties.

lib/cisco_node_utils/bridge_domain.rb
rkorlepa: Changed

lib/cisco_node_utils/bridge_domain_vni.rb
chrisvanheuveln: Ah, okay. I was expecting a loop to handle the add for each bd. I didn't realize the add cli could handle multiple individual bd's at once.

lib/cisco_node_utils/cisco_cmn_utils.rb
chrisvanheuveln: Okay, but they are very specific to bridge_domain_vni so please rename them to reflect that; i.e. `Utils.string_to_array` is pretty vague for a common utility method that's actually only used by bd vni code. Also, they should not be in this file unless you expect they'll be needed in more than one provider.

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
```diff
@@ -3,11 +3,54 @@
...
+# bridge-domain 100
+#   shutdown
+#   fabric-control
+#   name PepsiCo
+# so we cannot use get_context: '/^bridge-domain 100/' because it will search
+# for the first context and there is nothing under that. The properties are
+# present in the second context. Also name could appear in any order and hence
+# the if conditions.
+bd_name:
+  get_value: '/^bridge-domain[\s\S]*bridge-domain <bd>\s*(?:shutdown\s+fabric-control\s+name (\S+)|fabric-control\s+name (\S+)|shutdown\s+name (\S+)|name (\S+))/'
```
chrisvanheuveln: Ugh. This regex could be a lot simpler if our command ref infrastructure could handle multiline matching. But we'll save that for another day... :-(

In this case you can still simplify a bit by doing this (still ugly but a little better) pattern:

```
  get_value: '/bridge-domain <bd>\s*.*\s*.*\s*name (\S+)/
```

chrisvanheuveln: edit: My suggestion below works for shutdown,name, fabric-control but doesn't work for member-vni. It's a ridiculous cli.. Anyway, I say still use it and we'll figure out a different solution for member-vni.

This works as long as bridge-domain only has a combination of these three keywords in use. I see add'l keywords in the 7k cli so I would expect this to break if any of those other keywords are used.

Our command ref infrastructure doesn't currently support multi-line matching so that's not an option yet. (feel free to implement it! :-)

For now I think the solution is to take advantage of the trailing whitespace on the second bridge-domain that nvgens, i.e. the show run for this config:

```
system bridge-domain 100-113 

bridge-domain 100

bridge-domain 100 
  shutdown 
  fabric-control
  name fred100
```

...returns this pattern:

```
...\nbridge-domain 100\n\nbridge-domain 100 \n...
```

...so we can use that to differentiate the two and write our yaml like this:

```
_template:
  get_command: "show running-config bridge-domain"
  context: '/bridge-domain <bd>\s$/'

bd_name:
  get_value: '/name (\S+)/'
  set_value: "<state> name <name>"
  default_value: ~

fabric_control:
  get_value: '/^fabric-control#/'
  set_value: "<state> fabric-control"
  default: false

shutdown:
  get_value: '/^shutdown$/'
  set_value: "<state> shutdown"
  default: false
```

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -121,6 +121,30 @@ def destroy
     #                      PROPERTIES                      #
     ########################################################
 
+    # Bridge-Domain name assigning case
+    # bridge-domain 100
+    #   name PepsiCo
+    def bd_name
+      res = config_get('bridge_domain', 'bd_name', bd: @bd_ids)
+      return default_bd_name unless res
+      res.compact.join
```
chrisvanheuveln: See my yaml comment related to this: https://github.com/cisco/cisco-network-node-utils/pull/302/files#diff-dea5252853dd670210b2588b4305051eR49

...with that change you should be able to just call config_get() by itself:

```
  def bd_name
    config_get('bridge_domain', 'bd_name', bd: @bd_ids)
  end
```

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -121,6 +121,30 @@ def destroy
...
+    def bd_name=(str)
+      fail TypeError unless str.is_a?(String)
+      if str.empty?
+        config_set('bridge_domain', 'bd_name', bd: @bd_ids, state: 'no',
+                   name: '')
+      else
+        config_set('bridge_domain', 'bd_name', bd: @bd_ids, state: '',
+                   name: str)
+      end
+    end
```
chrisvanheuveln, line 139: You can simplify this to:

```
    def bd_name=(str)
      str = str.to_s
      state = str.empty? ? 'no' : ''
      config_set('bridge_domain', 'bd_name', state: state, bd: @bd_ids
    end
```

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -121,6 +121,30 @@ def destroy
...
+
+    def default_bd_name
+      sprintf('Bridge-Domain%s', @bd_ids)
```
chrisvanheuveln: This should be:

```
    def default_bd_name
      config_get_default('bridge_domain', 'bd_name')
    end
```

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -139,84 +163,14 @@ def default_fabric_control
...
     # Bridge-Domain Shutdown case
     # bridge-domain 100
     #   shutdown
     def shutdown
       result = config_get('bridge_domain', 'shutdown', bd: @bd_ids)
       # Valid result is either: "active"(aka no shutdown) or "shutdown"
-      result[/DOWN/] ? true : false
+      return false unless result
+      result[/shutdown/] ? true : false
     end
```
chrisvanheuveln, line 167:
```
    def shutdown
      config_get('bridge_domain', 'shutdown', bd: @bd_ids)
    end
```

lib/cisco_node_utils/cmd_ref/bridge_domain_vni.yaml
```diff
@@ -0,0 +1,34 @@
...
+_exclude: [N3k, N5k, N6k, N8k, N9k, ios_xr]
+
+create:
+  set_value: "bridge-domain <bd>"
+
+member_vni:
+  get_command: "show running-config bridge-domain | inc 'member vni'"
+  get_value: '/^  member vni (.*)/'
+  set_value: "bridge-domain <bd> ; <membstate> member vni <membvni>"
+  default_value: ''
```
chrisvanheuveln, line 13: In this case the running-config output is a bit useless thanks to the multiple bridge-domain entries, i.e. this is less than helpful:

```
bridge-domain 100

bridge-domain 100 
bridge-domain 100
  member vni 5000
```

So... I think we have no choice in this case but to use `show bridge-domain <bd>:

```

n7k8(config)# sh bridge-domain 100

Bridge-domain 100  (0 ports in all)
Name:: Bridge-Domain100
 Administrative State: UP               Operational State: DOWN
        vni5000
```

```
member_vni:
  get_command: "show bridge-domain"
  context: 'Name:: Bridge-Domain<bd>'
  get_value: '/vni(\S+)/'
```

lib/cisco_node_utils/cisco_cmn_utils.rb
rkorlepa: Yes this is used only in bridge_domain_vni file so moved it there.

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
rkorlepa: Thanks Chris, this is a good suggestion didn't know that there was a space in the show cli of bridge-domain. Have implemented it.

lib/cisco_node_utils/bridge_domain.rb
rkorlepa: Cannot just return nil for bd_name if its not initialized, because there is a default bd_name assigned to it when bd is created by the system and it is Bridge-Domain100. This case is similar to the vlan one where default vlan name is returned if there is no name in the running config.

lib/cisco_node_utils/bridge_domain.rb
rkorlepa, line 139: Changed it

lib/cisco_node_utils/bridge_domain.rb
rkorlepa: Same reason as before, default bd name exists as Bridge-Domain100 (100 in the end is based on bdid)

lib/cisco_node_utils/bridge_domain.rb
rkorlepa, line 167: shutdown is a boolean value, it needs to return true or false, based on yaml it would return 'shutdown' hence need to compare and return true/false.

shutdown:
  get_value: '/^shutdown$/'
  set_value: "<state> shutdown"
  default: false

glennmatthews, line 167: Try setting `kind: boolean` in the YAML - that should result in the automatic conversion to `true`/`false`.

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -104,30 +104,51 @@ def self.bds
...
     #                      PROPERTIES                      #
     ########################################################
 
+    # Bridge-Domain name assigning case
+    # bridge-domain 100
+    #   name PepsiCo
+    def bd_name
+      res = config_get('bridge_domain', 'bd_name', bd: @bd_ids)
+      res.nil? ? default_bd_name : res
+    end
```
chrisvanheuveln, line 132: This is better but you shouldn't need the nil check since default_bd_name returns nil. All you should need here is:

```
    def bd_name
      config_get('bridge_domain', 'bd_name', bd: @bd_ids)
    end
```

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -104,30 +104,51 @@ def self.bds
...
+
+    def bd_name=(str)
+      str = str.to_s
+      state = str.empty? ? 'no' : ''
+      config_set('bridge_domain', 'bd_name', bd: @bd_ids, state: state,
+                   name: str)
+    end
+
+    def default_bd_name
+      sprintf('Bridge-Domain%s', @bd_ids)
```
chrisvanheuveln: I'm still not sure why this sprintf is here. This should just be:

```
    def default_bd_name
      config_get_default('bridge_domain', 'bd_name')
    end
```

rkorlepa: We cannot use get default because, when we create a bd name is assigned automatically;
As below the bridge-domain 100 has a name assigned as Bridge-Domain100 without explicitly assigning one.

Hence doing it in this way.

n7k-venus(config)# system bridge-domain 100
n7k-venus(config)# bridge-domain 100
n7k-venus(config-bdomain)# exit
n7k-venus(config)# show bridge-domain 100

Bridge-domain 100  (0 ports in all)
Name:: Bridge-Domain100
 Administrative State: UP               Operational State: DOWN

n7k-venus(config)#

lib/cisco_node_utils/bridge_domain.rb
rkorlepa, line 132: bridge-domain name is never nil its either Bridge-Domain<id> or the one provided by user.

n7k-venus(config)# system bridge-domain 100
n7k-venus(config)# bridge-domain 100
n7k-venus(config-bdomain)# exit
n7k-venus(config)# show bridge-domain 100

Bridge-domain 100 (0 ports in all)
Name:: Bridge-Domain100
Administrative State: UP Operational State: DOWN

n7k-venus(config)#

lib/cisco_node_utils/bridge_domain_vni.rb
```diff
@@ -0,0 +1,206 @@
...
+      @bd_ids = bds.to_s.gsub(/\s+/, '')
+      fail 'bridge-domain value is empty' if @bd_ids.empty? # no empty strings
+
+      @bd_ids_list = BridgeDomainVNI.string_to_array(@bd_ids)
+      create if instantiate
+    end
+
+    def to_s
+      "Bridge Domain #{bd_ids}"
+    end
```
chrisvanheuveln, line 39: Where are you using this `to_s`?

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
```diff
@@ -3,59 +3,72 @@
...
+#   name Amazon
+# bridge-domain 109
+#   shutdown
+# bridge-domain 110
+#   name Microsoft
+
+_template:
+  get_command: "show running-config bridge-domain"
+  get_context: ['/^bridge-domain <bd>\s$/']
+  set_context: ["bridge-domain <bd>"]
```
chrisvanheuveln: You can make set_context optional this way:

```
  set_context: ["(?)bridge-domain <bd>"]
```

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
```diff
@@ -3,59 +3,72 @@
...
+
 all_bds:
   multiple: true
-  get_command: "show running-config bridge-domain"
+  get_context: ~
+  set_context: ~
```
chrisvanheuveln: If set_context is optional then you don't need to specify it here or on the properties that follow.

chrisvanheuveln:
This review is looking pretty good overall now. Just a few loose ends to tie up.

lib/cisco_node_utils/bridge_domain.rb
chrisvanheuveln, line 132: Our implementation is concerned with configuration state. The default config state is nil. Using show bridge-domain to get the name value is unnecessary.

lib/cisco_node_utils/bridge_domain.rb
chrisvanheuveln: Again, you're adding complexity and special casing where they're not needed. The more consistent we make our api's the easier they are to troubleshoot and maintain.

lib/cisco_node_utils/bridge_domain.rb
rkorlepa, line 132: If it is only the configuration state then default config as nil would work. May be I was thinking in a more complex way about it. I thought we had to return the current properties of bridge-domain. So if its just the configuration state then its not required.

lib/cisco_node_utils/bridge_domain.rb
rkorlepa: Yes based on previous comment its not required for the getter for bd_name.

lib/cisco_node_utils/bridge_domain_vni.rb
rkorlepa, line 39: This is for the CliErrors raised by any NodeUtil to avoid the cli_error_check function

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
rkorlepa: Added.

lib/cisco_node_utils/cmd_ref/bridge_domain.yaml
rkorlepa: Yes after adding the optional for set_context this is not required.
Removed it.

PR #304, opened 2016-03-21
minitest fixes for test_vpc

chrisvanheuveln:
- Tests now pass on n30,n31,n5,n6,n7,n9-I2,n9-I3
- Cleanup for feature-set fabricpath
- Includes infra fix for command_reference.rb: `supports?`
  - top-level excludes result in an empty `@hash`, so `value[property]` fails with:
    `undefined method`[]' for #Cisco::UnsupportedCmdRef:`
  - Fix tested against current `validate_property_excluded?` callers

### DIFF
--- a/lib/cisco_node_utils/fabricpath_global.rb
+++ b/lib/cisco_node_utils/fabricpath_global.rb
@@ -46,13 +46,18 @@ module Cisco
 
     def self.fabricpath_feature
       fabricpath = config_get('fabricpath', 'feature')
-      fail 'fabricpath_feature not found' if fabricpath.nil?
       return :disabled if fabricpath.nil?
       fabricpath.to_sym
     rescue Cisco::CliError => e
       # cmd will syntax reject when feature is not enabled
       raise unless e.clierror =~ /Syntax error/
       return :disabled
+    end
+
+    def self.fabricpath_enable
+      # TBD: Move to Feature provider
+      FabricpathGlobal.fabricpath_feature_set(:enabled) unless
+        FabricpathGlobal.fabricpath_feature == :enabled
     end
 
     def self.fabricpath_feature_set(fabricpath_set)
@@ -90,8 +95,7 @@ module Cisco
     end
 
     def create
-      FabricpathGlobal.fabricpath_feature_set(:enabled) unless
-      :enabled == FabricpathGlobal.fabricpath_feature
+      FabricpathGlobal.fabricpath_enable
     end
 
     def destroy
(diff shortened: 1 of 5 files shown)


lib/cisco_node_utils/vpc.rb
```diff
@@ -170,11 +168,13 @@ def fabricpath_multicast_load_balance
...
 
-    def fabricpath_multicast_load_balance=(val)
-      set_args_keys(state: val ? '' : 'no')
-      # This requires fabricpath_emulated_switch_id to be set first
-      fail 'fabricpath_switch_id configuration is required in the vPC domain' if
-        !fabricpath_emulated_switch_id && val
+    def fabricpath_multicast_load_balance=(state)
+      FabricpathGlobal.fabricpath_enable
+
+      fail 'This property also requires fabricpath_emulated_switch_id' if
```
mikewiebe: Is this the case where the message generated by the switch is unclear?  If so, please add some comments to indicate why the specific fail message is needed.

mikewiebe:
@chrisvanheuveln Can you run this against n8k?  I have it back up now.

lib/cisco_node_utils/vpc.rb
chrisvanheuveln: Note: This is prior code, I was just cleaning up the message.

In this case, if you configure `fabricpath multicast load-balance` first, without a `fabricpath switch-id` as well, it will configure just fine (no warning message) but it won't work until the switch-id is also added. 

IMO, this enters the too-helpful arena. We have plenty of prior art where independent commands are accepted and nvgen'd even though they have dependencies, so I suggest we remove this check altogether.

chrisvanheuveln:
It should skip all tests but it's not because your n8k reports as   `- type  - N9K-NXOSV`

lib/cisco_node_utils/fabricpath_global.rb
```diff
@@ -55,6 +54,12 @@ def self.fabricpath_feature
       return :disabled
     end
 
+    def self.fabricpath_enable
+      # TBD: Move to Feature provider
+      FabricpathGlobal.fabricpath_feature_set(:enabled) unless
+        FabricpathGlobal.fabricpath_feature == :enabled
```
robert-w-gries, line 60: Shouldn't feature code be in the `Feature` provider?

chrisvanheuveln, line 60: Totally agree, hence the  `# TBD: Move to Feature provider` comment :-)

This def I added is just a helper to make the calls cleaner but it's still an interim step. These Fabricpath feature calls are scattered through four different providers so moving it into Feature is outside the scope of my changes, but yes let's refactor it asap.

mikewiebe:
:+1:

chrisvanheuveln:
Found another n8k, all tests properly skip.

robert-w-gries:
:+1:

PR #305, opened 2016-03-21
Feature/itdservice

saichint:
itd_sevice has some issues on the nxos, so there are workarounds which are added to the code to get around these issues. Since Mike is familiar with these issues as he is in the loop, it is better if he reviews this PR.
Bugs have been filed for all the issues seen so far.
1. nxos accepts multiple next-hops for the same interface but this is incorrect. 
2. Only one Virtual IP can be configured on the switch, otherwise, when the service is made active and then shutdown, it crashes the application. There is also no uniqueness in the virtual ip even though it seems to have lots of parameters attached to it. So the entire string is taken as argument instead of making this as a separate provider.
3. There are memory leaks on the switch when we create/delete itd groups, nodes, services etc. If we continue to run the tests for many times, the switch becomes unusable and needs reboot.
4. vrf cannot be removed as an attribute to the service so this is not going to be supported in this release.
5. when nat destination is set, all ingress interfaces MUST have next-hop. This is not checked currently on the nxos and will be taken care in the puppet module.
# Mini test results:

[saichint@sjc-ads-7106 /ws/saichint-sjc/puppet_agent/cisco-network-node-utils/tests] $ ruby test_itd_service.rb
Run options: --seed 19730                                                                                      
# Running:

Node under test:
- name  - switch
- type  - N7K-C7010
- image - bootflash:///n7000-s2-dk9.7.3.0.D1.1.gbin

...........

Finished in 53.864832s, 0.2042 runs/s, 1.0582 assertions/s.

11 runs, 57 assertions, 0 failures, 0 errors, 0 skips

[saichint@sjc-ads-7106 /ws/saichint-sjc/puppet_agent/cisco-network-node-utils/tests] $ ruby test_itd_service.rb -e n9k  
Run options: -e n9k --seed 32332                                                                                     
# Running:

Node under test:
- name  - agent-lab13-nx
- type  - N9K-NXOSV  
- image - bootflash:///nxos.7.0.3.I3.1.bin

...........

Finished in 155.888309s, 0.0706 runs/s, 0.3656 assertions/s.

11 runs, 57 assertions, 0 failures, 0 errors, 0 skips


lib/cisco_node_utils/cmd_ref/itd_service.yaml
```diff
@@ -0,0 +1,120 @@
+# itd_service
+---
+_exclude: [N3k, N5k, N6k, N8k, ios_xr]
+
+_template:
+  get_command: "show running-config all | section itd"
+  get_context:
+      - '/^itd <name>/'
```
mikewiebe: Can you just put this on one line?: `get_context: ['/^itd <name>/']`

lib/cisco_node_utils/cmd_ref/itd_service.yaml
```diff
@@ -0,0 +1,120 @@
...
+peer_vdc:
+  _exclude: [N9k]
+  get_value: '/^peer vdc (\S+) service (\S+)$/'
+  set_value: "<state> peer vdc <vdc> service <service>"
+  default_value: []
+
+shutdown:
+  kind: boolean
+  auto_default: true
+  get_value: '/^no shut$/'
```
mikewiebe, line 111: _nit_ `'/^no shutdown$/'`

lib/cisco_node_utils/cmd_ref/itd_service.yaml
```diff
@@ -0,0 +1,120 @@
...
+---
+_exclude: [N3k, N5k, N6k, N8k, ios_xr]
+
+_template:
+  get_command: "show running-config all | section itd"
+  get_context:
+      - '/^itd <name>/'
+  set_context: ["itd <name>"]
+
+access_list:
```
mikewiebe, line 10: How do you envision this working in the puppet manifest?  I feel like it's more intuitive to have `access_list => ''` rather then `access_list => false` to remove the access list.

lib/cisco_node_utils/itd_service.rb
```diff
@@ -0,0 +1,518 @@
...
+
+    def default_exclude_access_list
+      config_get_default('itd_service', 'exclude_access_list')
+    end
+
+    def failaction
+      config_get('itd_service', 'failaction', @get_args)
+    end
+
+    def failaction=(state)
```
mikewiebe: For the method and yaml prop name can we use `fail_action` instead?  Easier on the eyes :)

lib/cisco_node_utils/itd_service.rb
```diff
@@ -0,0 +1,518 @@
...
+
+    # this is an array like:
+    # [['ethernet 1/1', '1.1.1.1'], ['port-channel 100', '2.2.2.2'],
+    # ['vlan 2', '3.3.3.3']]
+    # show command output is like: Eth1/1, Po100, Vlan2
+    # so translate back to the input format
+    def ingress_interface
+      list = config_get('itd_service', 'ingress_interface', @get_args)
+      list.each do |intf, _next_hop|
+        intf[/Eth/] = 'ethernet ' if intf.include?('Eth')
```
mikewiebe: Little more concise..

``` ruby
intf.gsub!('Eth', 'ethernet')
intf.gsub!('Po', 'port-channel')
intf.gsub!('Vlan', 'vlan')
```

lib/cisco_node_utils/itd_service.rb
```diff
@@ -0,0 +1,518 @@
...
+        intf[/Po/] = 'port-channel ' if intf.include?('Po')
+        intf[/Vlan/] = 'vlan ' if intf.include?('Vlan')
+      end
+      list
+    end
+
+    # only one next-hop is allowed per interface but
+    # due to nxos issues, it allows more than one;
+    # so the workaround is to clean up the current ingress
+    # intf and configure all of them from the manifest
```
mikewiebe: Don't mention manifest in the comments since this will also by use by other orchestration agents like chef.

lib/cisco_node_utils/itd_service.rb
```diff
@@ -0,0 +1,518 @@
...
+    def ingress_interface=(list)
+      cur_list = ingress_interface
```
mikewiebe, line 204: Can you move the cleanup into a separate method?

tests/test_itd_service.rb
```diff
@@ -0,0 +1,291 @@
...
+    itd.peer_vdc = itd.default_peer_vdc
+    assert_equal(itd.default_peer_vdc,
+                 itd.peer_vdc)
+    itd.destroy
+  end
+
+  def test_peer_local
+    itd = ItdService.new('new_group')
+    service = 'ser1'
+    if node.product_id =~ /N7/
```
mikewiebe: Please use `validate_property_excluded?` instead of hardcoding the platform.

https://github.com/cisco/cisco-network-node-utils/blob/6a0a58a2bf5627b6c91a97a5ae2abb18e2488cb4/tests/test_portchannel_global.rb#L55

tests/test_itd_service.rb
```diff
@@ -0,0 +1,291 @@
...
+    itd.shutdown = itd.default_shutdown
+    assert_equal(itd.default_shutdown,
+                 itd.shutdown)
+    itd.destroy
+  end
+
+  def test_peer_vdc
+    itd = ItdService.new('new_group')
+    parray = %w(vdc1 ser1)
+    if node.product_id =~ /N9/
```
mikewiebe: Please use `validate_property_excluded?` instead of hardcoding the platform.

https://github.com/cisco/cisco-network-node-utils/blob/6a0a58a2bf5627b6c91a97a5ae2abb18e2488cb4/tests/test_portchannel_global.rb#L55

tests/test_itd_service.rb
```diff
@@ -0,0 +1,291 @@
...
+    assert_equal(itd.load_bal_method_start_port,
+                 itd.default_load_bal_method_start_port)
+    assert_equal(itd.load_bal_method_proto,
+                 itd.default_load_bal_method_proto)
+    itd.destroy
+  end
+
+  def test_nat_destination
+    itd = ItdService.new('new_group')
+    if node.product_id =~ /N9/
```
mikewiebe: Please use `validate_property_excluded?` instead of hardcoding the platform.

https://github.com/cisco/cisco-network-node-utils/blob/6a0a58a2bf5627b6c91a97a5ae2abb18e2488cb4/tests/test_portchannel_global.rb#L55

lib/cisco_node_utils/cmd_ref/itd_service.yaml
saichint, line 111: Actually the show command shows "no shut" and not "no shutdown". So we have to leave it this way.

lib/cisco_node_utils/cmd_ref/itd_service.yaml
saichint, line 10: I was following this:
https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/cmd_ref/stp_global.yaml#L158

I originally started doing the way you asked here, but during channel_group development, Chris suggested to use "false" for default and ever since I changed it to that model. 
I can go back to my original thinking, but we need a standard way, we can discuss this tomorrow and go from there

lib/cisco_node_utils/cmd_ref/itd_service.yaml
mikewiebe, line 111: Ok.

lib/cisco_node_utils/itd_service.rb
```diff
@@ -0,0 +1,522 @@
...
+      list.each do |intf, _next_hop|
+        intf.gsub!('Eth', 'ethernet ')
+        intf.gsub!('Po', 'port-channel ')
+        intf.gsub!('Vlan', 'vlan ')
+      end
+      list
+    end
+
+    def ingress_interface_cleanup
+      cur_list = ingress_interface
```
mikewiebe, line 204: Can you return early if `cur_list` is empty?

tests/test_itd_service.rb
```diff
@@ -0,0 +1,291 @@
...
+    assert_equal('exclude', itd.exclude_access_list)
+    itd.access_list = itd.default_access_list
+    itd.exclude_access_list = itd.default_exclude_access_list
+    assert_equal(itd.default_access_list,
+                 itd.access_list)
+    assert_equal(itd.default_exclude_access_list,
+                 itd.exclude_access_list)
+    config 'no ip access-list include'
+    config 'no ip access-list exclude'
+    itd.destroy
```
mikewiebe: For `test_itd_service_create_destroy` you are validating the `destroy` method so explicit calls to `destroy` are needed.  For this testcase though, you should not need an explicit destroy because you are issuing a `no feature itd` in the `teardown` method.  Same is true for other test cases below.

lib/cisco_node_utils/itd_service.rb
saichint, line 204: added

saichint:
@mikewiebe I changed the default value of string types to ''. If there are no more comments, please merge this and I can start working on the puppet-module side.

saichint:
Sorry forgot to change peer_local attribute as well for the false vs ''. Done that now

PR #307, opened 2016-03-22
Added IOS XR support for tacacs_server_host

jonnytdevops:
I have created a new test file for this '_style2' which is created in the style that we have written all of our other minitests. The current tacacs_server_host minitests are too time consuming to convert to EXR due to the large amount of vshell calls and differing design.

Thanks


jonnytdevops:
```
TestTacacsServerHost#test_create_destroy_single_ipv6 = 9.85 s = .
W, [2016-03-22T17:12:07.512591 #91507]  WARN -- : 8.8.8.8 is configured multiple times on the device (possibly using different ports). This is unsupported by this API and the duplicate entries are being deleted.
TestTacacsServerHost#test_create_destroy_single_with_duplicates = 14.45 s = .
TestTacacsServerHost#test_create_destroy_single = 10.17 s = .
TestTacacsServerHost#test_create_destroy_multiple = 13.67 s = .

Finished in 50.862280s, 0.0786 runs/s, 0.8454 assertions/s.

4 runs, 43 assertions, 0 failures, 0 errors, 0 skips
```

tests/test_tacacs_server_host_style2.rb
```diff
@@ -0,0 +1,219 @@
+#
```
chburket: As discussed, we need to either:
a) replace the existing test_tacacs_server_host minitest with this new minitest (without losing significant test coverage)
b) make sure the old minitest skips for XR

lib/cisco_node_utils/cmd_ref/tacacs_server_host.yaml
```diff
@@ -1,37 +1,66 @@
...
+_template:
+  nexus:
+    get_command: "show run tacacs all"
+  ios_xr:
+    get_command: "show running-config tacacs-server"
 
 encryption:
-  set_value: "%s tacacs-server host %s key %s %s"
+  nexus:
+    set_value: "<state> tacacs-server host <ip> key <key>"
```
chrisvanheuveln: The `key` vars changed from `key %s %s` to `key <key>`. Do we not need the second var anymore? 

The old setter used to pass `encryption_password` there:

```
          config_set('tacacs_server_host', 'encryption', 'no', @name,
                     encryption_type,
                     encryption_password)
```

lib/cisco_node_utils/tacacs_server_host.rb
```diff
@@ -23,41 +23,107 @@ class TacacsServerHost < NodeUtil
...
+        end
+      end
+
+      create if instantiate
+
+      return if platform == :ios_xr
+
+      if instantiate
+        # 'feature tacacs+' must be enabled to create a host
+        TacacsServer.new.enable unless TacacsServer.enabled
```
chrisvanheuveln: Shouldn't this be in `def create`? which you call from line 41?

lib/cisco_node_utils/tacacs_server_host.rb
```diff
@@ -23,41 +23,107 @@ class TacacsServerHost < NodeUtil
     attr_reader :name
     @hosts = {}
 
-    def initialize(name, create=true)
+    def initialize(name, instantiate=true, host_port=nil)
```
chrisvanheuveln, line 26: It's not clear to me why you need to initialize with `host_port` now.
Is it a required part of the context now? 
If so, move `instantiate` to the end of the list since it's not part of the context.

`def initialize(name, host_port=nil, instantiate=true)`

lib/cisco_node_utils/tacacs_server_host.rb
```diff
@@ -23,41 +23,107 @@ class TacacsServerHost < NodeUtil
...
+
+      hosts_list = config_get('tacacs_server_host', 'hosts')
+      return hosts if hosts_list.empty?
 
-      return @hosts unless TacacsServer.enabled
+      hosts_list.each do |name|
+        if platform == :ios_xr
+          host_port = config_get('tacacs_server_host', 'port', ip: name)
+          host_port = host_port[0] if host_port.is_a?(Array)
+          host_port = host_port.to_i
```
chrisvanheuveln, line 61: `'port'` should always return an array since it's defined as a `multiple:`. What happens if `host_port` is nil?

Also, do you really need to do the `to_i`? Not saying it's wrong but I see that you removed `kind: int` from the yaml entry.

lib/cisco_node_utils/cmd_ref/tacacs_server_host.yaml
```diff
@@ -1,37 +1,66 @@
...
+  ios_xr:
+    context: ["tacacs-server host <ip> port <port>"]
+    get_value: '/key \d+\s+(\S+)/'
```
chburket: Don't need a `get_value` here, since we are only ever using this entry from `set_config`

lib/cisco_node_utils/tacacs_server_host.rb
```diff
@@ -23,41 +23,107 @@ class TacacsServerHost < NodeUtil
...
+      hosts
+    end
+
+    def create
+      destroy if platform == :ios_xr
+      config_set('tacacs_server_host',
+                 'host',
+                 state: '',
+                 ip:    name,
+                 port:  @port)
```
chrisvanheuveln, line 77: I think you need a separate setter for NX and XR. I'm fairly certain this will fail on NX since you're supplying an extra key:

```
host:
  nexus:
    set_value: "<state> tacacs-server host <ip>"
```

You have other properties that also look suspicious. Please check those out on both NX and XR.

lib/cisco_node_utils/cmd_ref/tacacs_server_host.yaml
```diff
@@ -1,37 +1,66 @@
...
+  multiple: true
+  get_command: "show running-config tacacs-server"
+  get_value: '/^tacacs-server host <ip> port (\d+)/'
+  default_value: ''
 
 hosts:
   multiple: true
-  get_command: "show run tacacs all"
   get_value: '/^tacacs-server host (\S+)/'
+  default_value: ''
```
chburket: Why the new default_value?  And why is the default an empty string when this will normally return an array?

lib/cisco_node_utils/tacacs_server_host.rb
jonnytdevops, line 26: @chrisvanheuveln Please refer to @glennmatthews for further details. This is the same issue that we had with radius_server (which is merged). Basically, the port forms part of the "uniqueness" of the host, and thus must be provided at creation time.

lib/cisco_node_utils/tacacs_server_host.rb
jonnytdevops, line 61: There is no way to specify that each element of the array must be an int, thus the to_i. This is used in many other providers.

lib/cisco_node_utils/tacacs_server_host.rb
jonnytdevops, line 77: No. The Cisco Node-Utils YAML has been specifically designed to allows multiple keys that are not required for a specific platform, without failing. Please refer to @glennmatthews for further details.

jonnytdevops:
Hi Folks, is there any scope to get this merged now? Thanks!

lib/cisco_node_utils/tacacs_server_host.rb
chrisvanheuveln, line 61: Okay, but again can config_get return nil for this? (since nil.to_i == 0)

chrisvanheuveln:
@jonnytpuppet JT, you need to address all comments before we'll merge this. You've made the review more difficult by amending your initial commit so it's hard to tell what you've changed from earlier commits. If you continue to do this then you need to explicitly respond to each and every comment.

jonnytdevops:
@chrisvanheuveln This is how we have been doing our PRs and reviews since we began. I believe all comments are addressed, except the test one, but I believe that is in discussion with management. Thanks!

lib/cisco_node_utils/tacacs_server_host.rb
jonnytdevops, line 61: Line 55 returns if the value is nil or empty. Thanks!

chrisvanheuveln:
@jonnytpuppet  _"This is how we have been doing our PRs and reviews since we began."_

Yes, I know, and we've disliked it from the beginning. Please stop. :-)  

We spend a lot of time on reviews and your method adds unnecessary overhead. Going forward please either: use separate follow-up commits once your code is in review; or explicitly respond to each comment.

---

Regarding this review I agree the only outstanding issue seems to be the test comment.

jonnytdevops:
@chrisvanheuveln sure, no problem. We can squash the commits after review at the end to neaten it up.

tests/test_tacacs_server_host_style2.rb
```diff
@@ -0,0 +1,219 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative 'ciscotest'
+require_relative '../lib/cisco_node_utils/tacacs_server_host'
+
+# TestTacacsServerHost - Minitest for TacacsServerHost node utility.
+class TestTacacsServerHost < CiscoTestCase
```
chburket: Can you rename this class to something unique (e.g. `TestTacacsServerHostStyle2`)?  Having the class named the same as the class in `test_tacacs_server_host.rb` is a bit confusing (difficult to tell which test was run by looking at the output).

jonnytdevops:
@chburket Code updated, I'll squash after +1, thanks

chburket:
Does the existing test_tacacs_server_host.rb pass/skip properly on XR?

jonnytdevops:
@chburket Nope, it doesn't. We're using the new test style now for IOS XR (which should also pass on Nexus)

jonnytdevops:
@chburket @chrisvanheuveln  Original tests updated for IOS and style2 test removed. All tests are passing bar one (and it's extremely trivial), as that depends on another provider which is still in development.

For reference, the failing test is testing this line of code:

```
config_get_default('tacacs_server', 'encryption_type')
```

Is it possible for this to be merged please? Thanks

```
Finished in 341.495921s, 0.0674 runs/s, 0.2196 assertions/s.

  1) Failure:
TestTacacsServerHost#test_tacacsserverhost_get_default_encryption_type [tests/test_tacacs_server_host.rb:285]:
Error: Tacacs Host default encryption type incorrect.
Expected: 0
  Actual: nil

23 runs, 75 assertions, 1 failures, 0 errors, 0 skips
```

tests/test_tacacs_server_host.rb
```diff
@@ -147,16 +166,19 @@ def test_tacacsserverhost_get_port
     port = DEFAULT_TACACS_SERVER_HOST_PORT
     assert_equal(port, host.port, 'Error: Tacacs Host port incorrect')
 
-    # when configured
-    port = 1138
-    config("tacacs-server host #{@host_name} port #{port}")
-    assert_equal(port, host.port, 'Error: Tacacs Host port incorrect')
+    # IOS XR support the same host using different ports
```
chburket, line 169: Should this comment say "IOS XR does not support..."?

chburket:
One possible nit, otherwise :+1:

tests/test_tacacs_server_host.rb
jonnytdevops, line 169: I can't think of anything to change this to. IOS support having the same host using different ports, so this means that instead of updating the port, it just adds another host.

chburket, line 169: Ah, then I think it's fine.

jonnytdevops:
@chburket @chrisvanheuveln Should I squash this now for merge? Thanks!

chburket:
Probably should wait for a review from someone on the NX side.

jonnytdevops:
@mikewiebe it would be great if this could be reviewed please, Thanks!

mikewiebe:
@jonnytpuppet Looks like @chrisvanheuveln  is already reviewing this PR so I will let him weigh in and approve.

tests/test_tacacs_server_host.rb
```diff
@@ -165,6 +187,8 @@ def test_tacacsserverhost_get_default_port
   end
 
   def test_tacacsserverhost_set_port
+    return if platform == :ios_xr
```
chrisvanheuveln: As discussed, aren't these the ones you were going to change to use `validate_property_excluded?` and assert for unsupported?

PR #310, opened 2016-03-22
vtp fixes

mikewiebe:
- Refactored to use non-structured output since structured output is not supported on `n5|6k`
- Fixed a few bugs in test_feature.rb
- Tests pass on n8k, n9k (I2 and I3)
- Failures on n5k, n6k and n7k due platform defects (see below).

```
  1) Failure:
TestVtp#test_password_special_characters [tests/test_vtp.rb:196]:
Expected: "hello!//\\#%$x"
  Actual: "hello!//\\\\\\\\#%$x"


  2) Failure:
TestVtp#test_create_double [tests/test_vtp.rb:88]:
Cisco::CliError expected but nothing was raised.
```
- Failure 1 caused by incorrect nxapi handling of backslashes.
- Failure 2 should generate a CLI error on n5|6|7k but does not.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/feature.yaml
+++ b/lib/cisco_node_utils/cmd_ref/feature.yaml
@@ -74,3 +74,10 @@ vni:
     set_value: 'feature vn-segment-vlan-based'
   N8k: *feature_mt_lite
   N9k: *feature_mt_lite
+
+vtp:
+  kind: boolean
+  get_command: "show running vtp"
+  get_value: '/^feature vtp$/'
+  set_value: "<state> feature vtp"
+  default_value: false
--- a/lib/cisco_node_utils/cmd_ref/feature.yaml
+++ b/lib/cisco_node_utils/cmd_ref/feature.yaml
@@ -77,7 +77,7 @@ vni:
 
 vtp:
   kind: boolean
-  get_command: "show running vtp"
+  get_command: "show running | i ^feature"
   get_value: '/^feature vtp$/'
   set_value: "<state> feature vtp"
   default_value: false
--- a/lib/cisco_node_utils/feature.rb
+++ b/lib/cisco_node_utils/feature.rb
@@ -157,6 +157,8 @@ module Cisco
       cli_error_check(result)
     end
 
+    # Special Case: The only way to remove a vtp instance
+    # is by disabling the feature.
     def self.vtp_disable
       return unless vtp_enabled?
       config_set('feature', 'vtp', state: 'no')
--- a/lib/cisco_node_utils/feature.rb
+++ b/lib/cisco_node_utils/feature.rb
@@ -151,6 +151,26 @@ module Cisco
     end
 
     # ---------------------------
+    def self.vtp_enable
+      return if vtp_enabled?
+      result = config_set('feature', 'vtp', state: '')
+      cli_error_check(result)
+    end
+
+    def self.vtp_disable
+      return unless vtp_enabled?
+      config_set('feature', 'vtp', state: 'no')
+    end
+
+    def self.vtp_enabled?
+      config_get('feature', 'vtp')
+    rescue Cisco::CliError => e
+      # cmd will syntax reject when feature is not enabled.
+      raise unless e.clierror =~ /Syntax error/
+      return false
+    end
+
+    # ---------------------------
     def self.cli_error_check(result)
       # The NXOS feature cli may not raise an exception in some conditions and
       # instead just displays a STDOUT error message; thus NXAPI does not detect
--- a/lib/cisco_node_utils/vtp.rb
+++ b/lib/cisco_node_utils/vtp.rb
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 require_relative 'node_util'
+require_relative 'logger'
 
 module Cisco
   # Vtp - node utility class for VTP configuration management
@@ -26,29 +27,22 @@ module Cisco
 
     # Constructor for Vtp
     def initialize(instantiate=true)
-      enable if instantiate && !Vtp.enabled
-    end
-
-    def self.enabled
-      config_get('vtp', 'feature')
-    rescue Cisco::CliError => e
-      # cmd will syntax reject when feature is not enabled
-      raise unless e.clierror =~ /Syntax error|Service not enabled/
-      return false
-    end
-
-    def enable
-      config_set('vtp', 'feature', '')
-    end
-
-    # Disable vtp feature
-    def destroy
-      config_set('vtp', 'feature', 'no')
+      Feature.vtp_enable if instantiate
     end
 
     # Get vtp domain name
     def self.domain
-      enabled ? config_get('vtp', 'domain') : ''
+      if Feature.vtp_enabled?
+        config_get('vtp', 'domain')
+      else
+        config_get_default('vtp', 'domain')
+      end
+    end
+
+    # The only way to remove a vtp domain is to turn the vtp
+    # feature off.
+    def destroy
+      Feature.vtp_disable
     end
 
     def domain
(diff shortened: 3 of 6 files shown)


lib/cisco_node_utils/cmd_ref/feature.yaml
```diff
@@ -74,3 +74,10 @@ vni:
     set_value: 'feature vn-segment-vlan-based'
   N8k: *feature_mt_lite
   N9k: *feature_mt_lite
+
+vtp:
+  kind: boolean
+  get_command: "show running vtp"
```
chrisvanheuveln: I would use `'show running | i ^feature' instead.

Or... we could use a template for this file in which case most of these don't need to specify a get_command.

lib/cisco_node_utils/feature.rb
```diff
@@ -151,6 +151,26 @@ def self.vni_enabled?
...
     # ---------------------------
+    def self.vtp_enable
+      return if vtp_enabled?
+      result = config_set('feature', 'vtp', state: '')
+      cli_error_check(result)
+    end
+
+    def self.vtp_disable
+      return unless vtp_enabled?
+      config_set('feature', 'vtp', state: 'no')
```
chrisvanheuveln, line 164: Please add a comment to this one explaining why it exists, i.e. why it's a special case. I don't want others to start adding disable methods to this file.

lib/cisco_node_utils/vtp.rb
```diff
@@ -59,19 +53,13 @@ def domain
     def domain=(d)
       fail ArgumentError unless d && d.is_a?(String) &&
                                 d.length.between?(1, MAX_VTP_DOMAIN_NAME_SIZE)
```
chrisvanheuveln: Nit: A little cleaner:

```
       d = d.to_s
       fail ArgumentError unless d.length.between?(1, MAX_VTP_DOMAIN_NAME_SIZE)
```

lib/cisco_node_utils/vtp.rb
```diff
@@ -81,18 +69,11 @@ def password=(password)
...
-      rescue Cisco::CliError => e
-        raise unless e.clierror =~ /password cannot be set for NULL domain/
-        unless password == default_password
-          raise 'Setting VTP password requires first setting VTP domain'
-        end
+      Feature.vtp_enable
+      if password == default_password
+        config_set('vtp', 'password', state: 'no', password: '')
+      else
+        config_set('vtp', 'password', state: '', password: password)
```
chrisvanheuveln: Style Nit: I prefer a single `config_set` call, and default_password is `''` so hard-coding password on the config_set doesn't help, so you could just do:

```
    def password=(password)
      Feature.vtp_enable
      state = (password == default_password) ? 'no' : ''
      config_set('vtp', 'password', state: state, password: password)
   end
```

lib/cisco_node_utils/vtp.rb
```diff
@@ -103,17 +84,18 @@ def default_password
...
     def filename=(uri)
       fail TypeError if uri.nil?
-      enable unless Vtp.enabled
+      Feature.vtp_enable
       if uri.empty?
-        config_set('vtp', 'filename', 'no', '')
+        config_set('vtp', 'filename', state: 'no', uri: '')
       else
-        config_set('vtp', 'filename', '', uri)
+        config_set('vtp', 'filename', state: '', uri: uri)
```
chrisvanheuveln:
```
    def filename=(uri)
      Feature.vtp_enable
      uri = uri.to_s
      state = uri.empty? ? 'no' : ''
      config_set('vtp', 'filename', state: state, uri: uri)
    end
```

tests/test_feature.rb
```diff
@@ -41,7 +41,7 @@ def vdc_lc_state(type=nil)
   # feature test helper
   def feature(feat)
     # Get the feature name string from the yaml
-    ref = cmd_ref.lookup('feature', feat).to_s[/set_value: feature (.*)/]
+    ref = cmd_ref.lookup('feature', feat).to_s[/set_value:.*feature (.*)/]
```
chrisvanheuveln, line 44: If you're going to change this I'd rather you change it to `\s+`

tests/test_feature.rb
```diff
@@ -51,7 +51,7 @@ def feature(feat)
 
     # Get current state of feature, then disable it
     pre_clean_enabled = Feature.send("#{feat}_enabled?")
-    config("no #{feat_str}") if pre_clean_enabled
+    config("no feature #{feat_str}") if pre_clean_enabled
```
chrisvanheuveln, line 54: Hmm. Thought I'd fixed that already. :-(

tests/test_vtp.rb
```diff
@@ -77,57 +71,42 @@ def test_vtp_domain_enable_disable
...
       vtp_domain(nil)
     end
   end
 
-  def test_vtp_create_domain_empty
+  def test_create_domain_empty
     assert_raises(ArgumentError) do
       vtp_domain('')
     end
   end
```
chrisvanheuveln, line 106: Nit: I know you're just changing the method names here but it would be more efficient (each 'no feature' cleanup takes time) if you combined these three (or more) negative tests into one test case:

```
  def test_domain_negative
    assert_raises(ArgumentError) { vtp_domain(node) }
    assert_raises(ArgumentError) { vtp_domain(nil) }
    assert_raises(ArgumentError) { vtp_domain('') }

    # Create same domain twice
    vtp_domain('accounting')
    assert_raises(Cisco::CliError) { vtp_domain('accounting') }
  end
```

tests/test_vtp.rb
```diff
@@ -304,19 +283,19 @@ def test_vtp_version_default
...
     vtp.version = 1
-    assert(Vtp.enabled)
+    assert(Feature.vtp_enabled?)
     assert_equal(1, vtp.version)
   end
 
-  def test_vtp_feature_enable_disable
-    Vtp.new.enable
-    assert(Vtp.enabled, 'Error: vtp is not enabled')
+  def test_feature_enable_disable
```
chrisvanheuveln, line 266: You already have this test in `test_feature.rb` so it's not needed here anymore.

lib/cisco_node_utils/cmd_ref/feature.yaml
mikewiebe: Your right.. I meant to change it but just moved it over from our old vtp.yaml file.

tests/test_feature.rb
mikewiebe, line 44: `\s+` won't work since this has to match `<state> feature vtp`.  I will change this to optionally match the `<state>` bit.

mikewiebe, line 44: Changed my mind on this.  I believe that `.*` is the best option here instead of trying to optionally match `<state>`.

lib/cisco_node_utils/feature.rb
mikewiebe, line 164: Done

lib/cisco_node_utils/vtp.rb
mikewiebe: Agreed.

tests/test_vtp.rb
mikewiebe, line 106: I am usually in favor of a higher level of test granularity so that we don't hold a large set of test cases hostage if one of them fails but in this case I'll allow it :)

tests/test_vtp.rb
mikewiebe, line 266: Instead of removing this testcase I added a test for `Feature.vtp_disable`.  I considered adding it to `test_feature.rb` but figured here was better since it's a special case.

lib/cisco_node_utils/vtp.rb
mikewiebe: Done

lib/cisco_node_utils/vtp.rb
mikewiebe: Done

tests/test_feature.rb
chrisvanheuveln, line 44: That's fine. I didn't realize you need to catch `<state>`.

tests/test_vtp.rb
chrisvanheuveln, line 266: yup, agreed.

chrisvanheuveln:
:+1:

PR #311, opened 2016-03-23
Addition of exr work for radius_global

HelenCampbell:


tests/test_radius_global.rb
```diff
@@ -65,11 +65,13 @@ def test_radius_global
     assert_equal(global.timeout,
                  2)
 
-    global.key_set('44444444', nil)
-    assert_equal(Cisco::RadiusGlobal.radius_global[id].key,
-                 '44444444')
-    assert_equal(global.key,
-                 '44444444')
+    if platform == :nexus
```
chburket, line 68: What's the reason behind skipping the "key" setter test on XR?

I'm also not sure why the setter is named `key_set` instead of `key=` (it wasn't done in this PR, I know).

glennmatthews, line 68: IIRC, the reason (for the name at least) was that `key=` would only take a single argument, by convention, while `key_set` actually takes 2 arguments.

HelenCampbell:
Hey @chburket, Regarding skipping the key setter on XR - When setting a key in XR if the key is passed in as unencrypted it encrypts it anyway, which would then which would make this test consistently fail as the get would return the encrypted key.

chburket:
@HelenCampbell Ok, I see the difference now that I have played around on the CLI of both XR and NX.  

On XR and NX, if you don't specify the key format, then it defaults to "cleartext" (0).  This should mean that XR/NX encrypts the key and then displays it in its encrypted state when doing a show.  

This works as expected (at least how I expect) on XR.  I find the behavior on NX to be inconsistent/odd.  Using the key value "44444444" (as in the minitest), NX seems to show the key the same no matter if I choose format 0 or 7 (or leave the format off).  If I use some other key value (eg "QsEfThUkO" as in the config guide), then NX seems to work like XR (the key is displayed in its encrypted form when shown).  On XR, the encrypted version of the key changes,  On NX, it seems to stay the same.

Not sure exactly what to do here.  At some point we probably want to enhance the node-utils API to allow the user to specify the key format along with the value, but that's outside the scope of this PR.

I think I would suggest making the test work the same on both NX and XR by setting a different value for the key ("QsEfThUkO" for example) and then just testing to make sure the key is not nil when reading the value (with a comment something like "# the key will be encrypted so just make sure it is not nil")

mikewiebe:
@HelenCampbell Agree with @chburket on his suggestion to make sure the test works the same way.  Can you include both NX and XR test results after the refactoring?

HelenCampbell:
Run options: -v -- --seed 10689

# Running:

Node under test:
- name  - a7ytwzy0ptd0nvd
- type  - R-IOSXRV9000-CH
- image - 6.0.0.20I

TestRadiusGlobal#test_radius_global = 36.68 s = .

Finished in 42.033661s, 0.0238 runs/s, 0.3093 assertions/s.

1 runs, 13 assertions, 0 failures, 0 errors, 0 skips

@chburket @mikewiebe Refactored - give me a shout if you need anything else. I can't run the tests on NX as I don't have an environment, would someone on your side be able to run that for me?

mikewiebe:
@HelenCampbell Sure thing.  I will run it now.

chburket:
:+1:

mikewiebe:
@HelenCampbell Tests pass on all nexus platforms (n3k, n9k(I2 and I3), n8k, n5k, n6k, n7k).  Will merge

mikewiebe:
:+1:

PR #312, opened 2016-03-23
test_router_ospf, test_router_ospf_vrf fixes

chrisvanheuveln:
- I ran into some issues with feature disablement on one platform so I took the opportunity to just remove it altogether since that was the long-term plan anyway; also moved feature enable into the Feature provider.
- test_router_ospf.rb was one of the original minitests. It had a lot of repetitive tests so I gutted it.
- Minitests now pass on all NX platforms

### DIFF
--- a/lib/cisco_node_utils/router_ospf.rb
+++ b/lib/cisco_node_utils/router_ospf.rb
@@ -15,7 +15,6 @@
 # limitations under the License.
 
 require_relative 'node_util'
-require_relative 'feature'
 
 module Cisco
   # RouterOspf - node utility class for process-level OSPF config management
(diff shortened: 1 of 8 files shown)


lib/cisco_node_utils/router_ospf.rb
```diff
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 require_relative 'node_util'
+require_relative 'feature'
```
mikewiebe: I don't think you need this...

mikewiebe:
@chrisvanheuveln One comment, otherwise :+1:

lib/cisco_node_utils/router_ospf.rb
chrisvanheuveln: You're right. Fixed.

PR #314, opened 2016-03-24
Fix private vlan test for N5K

davidefdl:


tests/test_vlan_private.rb
```diff
@@ -387,7 +387,7 @@ def test_private_vlan_association_failure
 
       pv_type = 'isolated'
       assert_raises(RuntimeError, 'vlan misconf did not raise RuntimeError') do
-        v2.private_vlan_type = pv_type
+        v3.private_vlan_type = pv_type
```
robert-w-gries, line 390: Why is this change necessary?

chrisvanheuveln:
Changes look fine but have you tested on all platforms?  n3,n5,n6,n7,n9 ?

tests/test_vlan_private.rb
chrisvanheuveln, line 390: You have to look at the prev commands. v3 is vlan 200 but vlan 200 became part of an association, so now you can't isolate it, hence the RuntimeError.

robert-w-gries, line 390: Ack

robert-w-gries:
:+1: pending test results

davidefdl:
I have tetested for 7k,9k,3k,5k. Do not have 6k

davidefdl:
Ran the tests also for 6K from one under developer Portal .
To notice that pvlan will not work for 9Kv since is not supported. You need a physical 9K switch

robert-w-gries:
@chrisvanheuveln @mikewiebe Should we mark N9kv as an unsupported platform for this property?

chrisvanheuveln:
Hmm. We haven't done any n9kv validation yet, strictly speaking. And it's not really on our "supported" platforms list. I suggest we overlook it for this PR but punt to Sanjay for a long-term decision.

davidefdl:
Is the label N9kv present already in exclude? If yes I can change the code to have that.
@chrisvanheuveln @robert-w-gries @mikewiebe

chrisvanheuveln:
@davidefdl If not you can add it. But if you do that then make sure that your minitest explicitly tests for non-support, a la `validate_property_excluded?()` checks.

davidefdl:
How do we add a new flag? is there a location where N7K etc are defined? @chrisvanheuveln

robert-w-gries:
I added some N3k product ids in #316 that you can use as a reference.
1. [Add `N9kv` as a member of `KNOWN_PLATFORMS`](https://github.com/cisco/cisco-network-node-utils/pull/316/files#diff-39df58fcc11e8557b9418cc9a019d6c4R437)
   - Make sure the `product_id` regexp matches the N9kv product id: `N9K-NXOSV`
   - [See XR regexp for reference](https://github.com/cisco/cisco-network-node-utils/pull/316/files#diff-39df58fcc11e8557b9418cc9a019d6c4R442)
2. [Add `N9kv` to schema.yaml so Travis doesn't fail](https://github.com/cisco/cisco-network-node-utils/pull/316/files#diff-b577ba891bdf86768665cd33cc6355f3R15)
3. [Add `n9kv` test to `test_command_reference.rb`](https://github.com/cisco/cisco-network-node-utils/pull/316/files#diff-f5135295d0cf9bc1e8a8b0b2c583d04bR227)

robert-w-gries:
Since I changed how the [`platform_to_filter` method works](https://github.com/cisco/cisco-network-node-utils/pull/316/files#diff-39df58fcc11e8557b9418cc9a019d6c4R439), you would have to wait until my PR is merged.  I say hold off on that change, and we add a separate User Story for adding a N9kv exclude.

mikewiebe:
Agreed that we should hold off on adding N9Kv.  That should be a different user story.  I am going to go ahead and merge these changes because I need them for a different fix for the n8k.

PR #316, opened 2016-03-24
Feature/interface test all plats

robert-w-gries:
## Changes
1. Specify `@http.use_ssl = false` to avoid EOFError from `net/protocol`
   - Note: still seeing EOFError error on n3k-52 (C3064) but it may be resolved by reload
2. Added C3064, C3132, and C3172 as valid product ids in YAML
3. Refactored speed tests to properly test valid values
4. Refactored product id code in `command_reference` to reduce overhead of adding platforms
5. Moved `feature-set fex` to feature provider
## Test Results

|  | N3k | N5k | N6k | N7k | N8k | N9k I2 | N9k I3 | XR |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| test_interface | 1 F <br>1 S | ✅ | ✅ | ✅ | ✅ | 1 F <br>1 S | 2 F | ✅ |
| test_interface_switchport | ✅ | ✅ <br>3 S | ✅ <br>3 S | ✅ <br>3 S | ✅ | ✅ | ✅ | ✅ <br>24 S |
| test_interface_svi | ✅ | ✅ <br>4 S | ✅ <br>4 S | ✅ | ✅ | ✅ | ✅ | ✅ <br>Skipped |
### N3k / N9k I2
- Known issue with vlan interfaces not updating ip configuration

``` bash
  1) Failure:
TestInterface#test_interface_ipv4_all_interfaces [/home/robgrie/cisco-network-node-utils/tests/test_interface.rb:337]:
Expected /^\s+ip address 9.7.1.1\/15$/ to match output of '"show run interface vlan45 all | no-more"':
show run interface vlan45 all | no-more
```
### N8k
- All failures are due to the node reporting as a N9k switch
### N9k I3
- Known issue where `negotiate auto` error is not raised by switch

### DIFF
--- a/lib/cisco_node_utils/client/nxapi/client.rb
+++ b/lib/cisco_node_utils/client/nxapi/client.rb
@@ -186,6 +186,7 @@ class Cisco::Client::NXAPI < Cisco::Client
     debug("Sending HTTP request to NX-API at #{@http.address}:\n" \
           "#{request.to_hash}\n#{request.body}")
     begin
+      @http.use_ssl = false
       response = @http.request(request)
     rescue Errno::ECONNREFUSED, Errno::ECONNRESET
       emsg = 'Connection refused or reset. Is the NX-API feature enabled?'
--- a/lib/cisco_node_utils/client/nxapi/client.rb
+++ b/lib/cisco_node_utils/client/nxapi/client.rb
@@ -186,6 +186,8 @@ class Cisco::Client::NXAPI < Cisco::Client
     debug("Sending HTTP request to NX-API at #{@http.address}:\n" \
           "#{request.to_hash}\n#{request.body}")
     begin
+      # Explicitly use http to avoid EOFError
+      # http://stackoverflow.com/a/23080693
       @http.use_ssl = false
       response = @http.request(request)
     rescue Errno::ECONNREFUSED, Errno::ECONNRESET
--- a/lib/cisco_node_utils/cmd_ref/overlay_global.yaml
+++ b/lib/cisco_node_utils/cmd_ref/overlay_global.yaml
@@ -3,7 +3,7 @@
 _exclude: [ios_xr]
 
 anycast_gateway_mac:
-  _exclude: [N3k, N7k]
+  _exclude: [N3k]
   kind: string
   get_command: "show running all | inc 'fabric forwarding'"
   get_value: '/^fabric forwarding anycast-gateway-mac (\S+)$/'
--- a/lib/cisco_node_utils/cmd_ref/overlay_global.yaml
+++ b/lib/cisco_node_utils/cmd_ref/overlay_global.yaml
@@ -3,7 +3,7 @@
 _exclude: [ios_xr]
 
 anycast_gateway_mac:
-  _exclude: [N3k]
+  _exclude: [N3k, N7k]
   kind: string
   get_command: "show running all | inc 'fabric forwarding'"
   get_value: '/^fabric forwarding anycast-gateway-mac (\S+)$/'
--- a/lib/cisco_node_utils/interface.rb
+++ b/lib/cisco_node_utils/interface.rb
@@ -15,9 +15,7 @@
 # limitations under the License.
 
 require_relative 'cisco_cmn_utils'
-require_relative 'feature'
 require_relative 'node_util'
-require_relative 'pim'
 require_relative 'vrf'
 require_relative 'overlay_global'
 
--- a/lib/cisco_node_utils/interface.rb
+++ b/lib/cisco_node_utils/interface.rb
@@ -475,13 +475,11 @@ module Cisco
     end
 
     def speed
-      config_get('interface', 'speed', name: @name)
+      val = config_get('interface', 'speed', name: @name)
+      val == 'auto' ? val : val.to_i
     end
 
     def speed=(val)
-      if node.product_id =~ /C31\d\d/
-        fail 'Changing interface speed is not permitted on this platform'
-      end
       config_set('interface', 'speed', name: @name, speed: val)
     end
 
@@ -494,9 +492,6 @@ module Cisco
     end
 
     def duplex=(val)
-      if node.product_id =~ /C31\d\d/
-        fail 'Changing interface duplex is not permitted on this platform'
-      end
       config_set('interface', 'duplex', name: @name, duplex: val)
     end
 
--- a/lib/cisco_node_utils/interface.rb
+++ b/lib/cisco_node_utils/interface.rb
@@ -474,6 +474,7 @@ module Cisco
 
     def speed
       val = config_get('interface', 'speed', name: @name)
+      return val if val.nil?
       val == 'auto' ? val : val.to_i
     end
 
--- a/lib/cisco_node_utils/interface.rb
+++ b/lib/cisco_node_utils/interface.rb
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 require_relative 'cisco_cmn_utils'
+require_relative 'feature'
 require_relative 'node_util'
 require_relative 'pim'
 require_relative 'vrf'
@@ -179,31 +180,6 @@ module Cisco
 
     def default_fabric_forwarding_anycast_gateway
       config_get_default('interface', 'fabric_forwarding_anycast_gateway')
-    end
-
-    def fex_feature
-      fex = config_get('fex', 'feature')
-      fail 'fex_feature not found' if fex.nil?
-      fex.to_sym
-    end
-
-    def fex_feature_set(fex_set)
-      curr = fex_feature
-      return if curr == fex_set
-
-      case fex_set
-      when :enabled
-        config_set('fex', 'feature_install', '') if curr == :uninstalled
-        config_set('fex', 'feature', '')
-      when :disabled
-        config_set('fex', 'feature', 'no') if curr == :enabled
-        return
-      when :installed
-        config_set('fex', 'feature_install', '') if curr == :uninstalled
-      when :uninstalled
-        config_set('fex', 'feature', 'no') if curr == :enabled
-        config_set('fex', 'feature_install', 'no')
-      end
     end
 
     def ipv4_acl_in
@@ -862,7 +838,7 @@ module Cisco
       if :fabricpath == mode_set
         fabricpath_feature_set(:enabled) unless :enabled == fabricpath_feature
       elsif :fex_fabric == mode_set
-        fex_feature_set(:enabled) unless :enabled == fex_feature
+        Feature.fex_enable
       end
       config_set('interface', switchport_mode_lookup_string,
                  name: @name, state: '', mode: IF_SWITCHPORT_MODE[mode_set])
--- a/tests/test_interface.rb
+++ b/tests/test_interface.rb
@@ -1072,9 +1072,8 @@ class TestInterface < CiscoTestCase
   end
 
   def test_interface_fabric_forwarding_anycast_gateway
-    # Skip test if incompatible linecard detected
-    msg = 'Required Multi-Tenancy Full compatible linecard for this platform'
-    skip(msg) if node.product_id[/N7/] && !mt_full_interface?
+    # Ensure N7k has compatible interface
+    mt_full_interface? if node.product_id[/N7/]
 
     if validate_property_excluded?('overlay_global', 'anycast_gateway_mac')
       assert_raises(Cisco::UnsupportedError) do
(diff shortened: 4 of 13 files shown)


lib/cisco_node_utils/client/nxapi/client.rb
```diff
@@ -186,6 +186,7 @@ def req(type, command_or_list)
     debug("Sending HTTP request to NX-API at #{@http.address}:\n" \
           "#{request.to_hash}\n#{request.body}")
     begin
+      @http.use_ssl = false
```
chrisvanheuveln, line 191: Please add a comment to this to indicate why it's here, ie. "avoids EOFError from net/protocol on some platforms"

lib/cisco_node_utils/cmd_ref/overlay_global.yaml
```diff
@@ -3,7 +3,7 @@
 _exclude: [ios_xr]
 
 anycast_gateway_mac:
-  _exclude: [N3k]
+  _exclude: [N3k, N7k]
```
chrisvanheuveln: This property works fine on my N7k. What is driving this change?

lib/cisco_node_utils/interface.rb
```diff
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 require_relative 'cisco_cmn_utils'
+require_relative 'feature'
```
chrisvanheuveln: You don't need this require.

Looks like the pim require isn't needed either.

lib/cisco_node_utils/interface.rb
```diff
@@ -499,13 +475,11 @@ def default_mtu
     end
 
     def speed
-      config_get('interface', 'speed', name: @name)
+      val = config_get('interface', 'speed', name: @name)
+      val == 'auto' ? val : val.to_i
```
chrisvanheuveln, line 478: 1. Do you really need the `to_i`? Can we just handle it as a string? Or is this fixing a particular problem?
2. If val is nil, then val.to_i == 0.   Will that be a problem?

tests/test_interface.rb
```diff
@@ -165,6 +170,37 @@ def interface_supports_property?(intf, message)
...
+    valid_speed = nil
+    speeds = capable_speed_values(interface)
+    speeds.each do |value|
+      next if value == 'auto'
+      begin
+        interface.speed = value
+        assert_equal(value.to_i, interface.speed)
+      rescue Cisco::CliError => e
+        next if speed_change_disallowed?(e.message)
+        raise
```
chrisvanheuveln, line 184: This should probably be:

```
-        raise
+        flunk(e.message)
```

tests/test_interface.rb
```diff
@@ -165,6 +170,37 @@ def interface_supports_property?(intf, message)
...
+    end
+    valid_speed
+  end
+
+  # Helper to check for misc speed change disallowed error messages.
+  def speed_change_disallowed?(message)
+    patterns = ['port doesn t support this speed',
+                'Changing interface speed is not permitted',
+                'requested config change not allowed',
+                /does not match the (transceiver speed|port capability)/,
```
chrisvanheuveln, line 198: slashes or quotes here?

tests/test_interface.rb
```diff
@@ -528,43 +564,68 @@ def test_mtu_invalid_loopback
 
   def test_speed
     interface = Interface.new(interfaces[0])
-    if platform == :ios_xr
+    if node.product_id =~ /C3132/
+      skip('Speed is not supported on this platform')
+    elsif validate_property_excluded?('interface', 'speed')
```
chrisvanheuveln: Why the `elsif` ? Why isn't the C3132 covered by this validate_ check?

tests/test_interface.rb
```diff
@@ -528,43 +564,68 @@ def test_mtu_invalid_loopback
...
-        assert_equal('100', interface.speed)
+        interface.speed = value
+        assert_equal(value.to_i, interface.speed)
+        successful_runs += 1
       rescue Cisco::CliError => e
-        speed_change_disallowed(e.message)
+        # Many of the 'capable' speeds are actually not valid values
+        # Try next available speed value if CLI rejects current value
+        next if speed_change_disallowed?(e.message)
+        raise
```
chrisvanheuveln, line 590: flunk

tests/test_interface.rb
```diff
@@ -1321,7 +1352,7 @@ def test_interface_ipv4_all_interfaces
       validate_interface_shutdown(inttype_h)
       validate_vrf(inttype_h)
       config(*cfg)
-      InterfaceChannelGroup.new(interfaces[1]).channel_group = false
+      interface_ethernet_default(interfaces[1])
```
chrisvanheuveln, line 1356: I vaguely remember a problem where 'default interface' does not clear the channel group on some platform (I don't remember which). 
I suggest doing both:

```
      interface_ethernet_default(interfaces[1])
      InterfaceChannelGroup.new(interfaces[1]).channel_group = false
```

...or add the channel_group cleanup to `def interface_ethernet_default`

chrisvanheuveln:
Overall it looks pretty good. Very thorough testing.

lib/cisco_node_utils/cmd_ref/overlay_global.yaml
robert-w-gries: It doesn't seem to work on `n7k-j`.  `feature fabric forwarding` doesn't nvgen

lib/cisco_node_utils/interface.rb
robert-w-gries: Removed

lib/cisco_node_utils/interface.rb
robert-w-gries, line 478: Yes, `val == nil` is a problem that I thought I handled by adding `return val if val.nil?` but I guess I forgot to commit it.

It made sense to me to treat `speed` as an integer where applicable, but if you don't think it adds value I can get rid of it.  I also thought it would help in the puppet layer if we wanted to specify integers.

tests/test_interface.rb
robert-w-gries, line 184: I prefer raising an error than reporting the test as a "failure" when the CLI errors after a config command.  It's usually an indicator to me that our YAML needs to be fixed, either by `exclude` or by specifying a different `default_value`

tests/test_interface.rb
robert-w-gries, line 198: The line is a regex that combines the `does not match the transceiver speed` and `does not match the port capability` error messages into one line

tests/test_interface.rb
robert-w-gries, line 1356: I tested against all platforms (except for 8k) and did not see `channel group` remaining after running `default interface ethernet 1/1`

glennmatthews, line 1356: Does 'all platforms' include XRv9k?

robert-w-gries, line 1356: Yes I've confirmed it on a full sunstone image

lib/cisco_node_utils/cmd_ref/overlay_global.yaml
chrisvanheuveln: Possibly a license or h/w requirement. Trust me, I ran the overlay test on my n7k and this property works.

lib/cisco_node_utils/interface.rb
chrisvanheuveln, line 478: It doesn't _really_ matter but my thinking is:
- The config_get's are usually dealing with strings (as a result of the regex matches)
- On the puppet side the manifest can specify either but the type should normalize / munge it regardless
- If we consistently treat these as strings then we don't have to coerce them anywhere but in the type munge

tests/test_interface.rb
chrisvanheuveln, line 184: Keep it. I don't feel strongly about it. 
I think the only difference you'll see is that you'll get a traceback with `raise` versus a simple failure message with `flunk`.

tests/test_interface.rb
chrisvanheuveln, line 198: I see. I got that they're all used as regex patterns but I didn't realize you need the slashes in this case to avoid a literal match of the `(|)` chars.

tests/test_interface.rb
chrisvanheuveln, line 1356: Sounds good. If it breaks anything I know who to contact. :-)

lib/cisco_node_utils/client/nxapi/client.rb
robert-w-gries, line 191: Added comment.  Unfortunately, the node I was seeing the error on (n3k-52) is now down, so I can't verify the error is fixed.  Regardless, explicitly setting `use_ssl` to `false` looks like it will prevent any EOFErrors from happening

lib/cisco_node_utils/cmd_ref/overlay_global.yaml
robert-w-gries: Seems to be an issue with `n7k-j`.  `feature-set fabric` is apparently installed but isn't allowed to be enabled

``` bash
n7k-j(config)# show feature-set
Feature Set Name      ID        State
--------------------  --------  --------
fcoe                   1          uninstalled
fabricpath             2          disabled
fex                    3          enabled
mpls                   4          uninstalled
fabric                 7          installed

n7k-j(config)# feature-set ?
  fabricpath  Installed feature-set
  fex         Installed feature-set

n7k-j(config)# feature-set fabric

n7k-j(config)# show feature-set
Feature Set Name      ID        State
--------------------  --------  --------
fcoe                   1          uninstalled
fabricpath             2          enabled
fex                    3          enabled
mpls                   4          uninstalled
fabric                 7          installed
```

lib/cisco_node_utils/interface.rb
robert-w-gries, line 478: Fair points. I'll undo the change

lib/cisco_node_utils/cmd_ref/overlay_global.yaml
chrisvanheuveln:
```
n7k-j(config-vdc)# allow feature-set fabric 
ERROR: Feature-set fabric works only in VDC supporting solely F3 module type.
```

No F3:

```
n7k-j(config-vdc)# sh mod
Mod  Ports  Module-Type                         Model              Status
---  -----  ----------------------------------- ------------------ ----------
1    0      Supervisor Module-2                 N7K-SUP2E          active *
3    32     10 Gbps Ethernet XL Module          N7K-M132XP-12L     ok
```

robert-w-gries: Does this need to be handled in the test?

lib/cisco_node_utils/interface.rb
robert-w-gries, line 478: I ended up returning the speed values as integers

tests/test_interface.rb
robert-w-gries: I mistakenly thought the issue with `nil` values being turned into `0` by `val.to_i` was an issue with C3132 platforms.  It's removed now.

lib/cisco_node_utils/cmd_ref/overlay_global.yaml
chrisvanheuveln: I think that would be prudent. Could probably just add a check here to do the needful:

https://github.com/cisco/cisco-network-node-utils/blob/develop/tests/test_overlay_global.rb#L94

...which would do:

```
  skip('Test requires Multi-Tenancy Full compatible linecard on this platform') if
    node.product_id[/N7/] && !mt_full_interface?
```

robert-w-gries:
Re-added N7k platform for `fabric_forwarding` and `anycast_gateway`.

`test_interface_fabric_forwarding_anycast_gateway` test passes for all platforms, except for `N7k-j` which skips with:

``` bash
TestInterface#test_interface_fabric_forwarding_anycast_gateway = 6.85 s = S

Finished in 6.852940s, 0.1459 runs/s, 0.0000 assertions/s.

  1) Skipped:
TestInterface#test_interface_fabric_forwarding_anycast_gateway [/home/robgrie/cisco-network-node-utils/tests/ciscotest.rb:241]:
Unable to find compatible interface in chassis
```

chrisvanheuveln:
:+1: 

I don't care for the msg wording here: https://github.com/cisco/cisco-network-node-utils/pull/316/files#diff-cf2422f8b2cb8aa033a2f17f39d7354fR1076

...but it's close enough. Ship it... :-)

PR #317, opened 2016-03-25
test_fabric_global fixes

chrisvanheuveln:
- Now passes on n5,n6,n7. Not supported on all others.
- `rotate` defaults removed since they are randomized values
- `algorithm` type `symmetric` is only supported on n7k
- minitest general cleanup: removed class vars, added platform checks, cleaned up error messaging
- test_vlan_mt_full.rb can fail if the test vlans are currently configured as bridge-domain vlans. Added a bd cleanup to the initial setup call.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/fabricpath.yaml
+++ b/lib/cisco_node_utils/cmd_ref/fabricpath.yaml
@@ -1,6 +1,5 @@
 # fabricpath
 ---
-# Fabricpath feature is not available on N3K and N9K
 _exclude: [N3k, N8k, N9k, ios_xr]
 
 aggregate_multicast_routes:
@@ -34,9 +33,8 @@ feature_install:
 
 graceful_merge:
   auto_default: false
-  kind: boolean
   get_command: "show run fabricpath all"
-  get_value: '/^fabricpath graceful\-merge disable\s*$/'
+  get_value: '/^fabricpath graceful-merge disable\s*$/'
   set_value: "<state> fabricpath graceful-merge disable"
   default_value: enable
 
@@ -69,17 +67,23 @@ loadbalance_algorithm:
   get_command: "show run fabricpath all"
   get_value: '/^fabricpath load\-balance (\S+)\s*$/'
   set_value: "<state> fabricpath load-balance <algo>"
-  N5k:
+  N5k: &loadbal_alg_n5_n6
     default_value: "source-destination"
+  N6k: *loadbal_alg_n5_n6
   else:
     default_value: "symmetric"
+
+loadbalance_algorithm_symmetric_support:
+  _exclude: [N5k, N6k]
+  kind: boolean
+  default_only: true
 
 loadbalance_multicast_has_vlan:
   _exclude: [N5k, N6k]
   kind: boolean
   auto_default: false
-  get_command: "show fabricpath load-balance | begin Ftag"
-  get_value: '/^Use VLAN: TRUE\s*$/'
+  get_command: 'show run fabricpath all'
+  get_value: '/^fabricpath load-balance multicast .*include-vlan/'
   default_value: true
 
 loadbalance_multicast_reset:
@@ -91,17 +95,17 @@ loadbalance_multicast_rotate:
   kind: int
   get_command: "show fabricpath load-balance | begin Ftag"
   get_value: '/^Rotate amount: (\d+)/'
-  default_value: 1
+  # default_value: n/a. The default rotate amount is randomized to avoid polarization.
 
 loadbalance_multicast_set:
   _exclude: [N5k, N6k]
-  set_value: 'fabricpath load-balance multicast <rotate_amt> <inc_vlan>'
+  set_value: 'fabricpath load-balance multicast <rotate> <has_vlan>'
 
 loadbalance_unicast_has_vlan:
   kind: boolean
   auto_default: false
-  get_command: "show fabricpath load-balance | begin ECMP next 4"
-  get_value: '/^Use VLAN: TRUE\s*$/'
+  get_command: 'show run fabricpath all'
+  get_value: '/^fabricpath load-balance unicast .*include-vlan/'
   set_value: "<state> fabricpath load-balance unicast include-vlan"
   default_value: true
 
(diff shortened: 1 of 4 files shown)


lib/cisco_node_utils/cmd_ref/fabricpath.yaml
```diff
@@ -69,17 +67,23 @@ loadbalance_algorithm:
   get_command: "show run fabricpath all"
   get_value: '/^fabricpath load\-balance (\S+)\s*$/'
   set_value: "<state> fabricpath load-balance <algo>"
-  N5k:
+  N5k: &loadbal_alg_n5_n6
```
mikewiebe: n5k and n6k are implied by the yaml entries themselves.  I would change this to be just `&loadbal_alg` and `*loadbal_alg` or something else meaningful minus the platforms.

lib/cisco_node_utils/cmd_ref/fabricpath.yaml
```diff
@@ -120,23 +124,20 @@ loadbalance_unicast_rotate:
...
 loadbalance_unicast_set:
-  set_value: 'fabricpath load-balance unicast <pref> <rotate_amt> <inc_vlan>'
+  set_value: 'fabricpath load-balance unicast <layer> <rotate> <has_vlan>'
 
 loadbalance_unicast_support:
   kind: string
-  N5k:
-    default_only: "split"
-  N6k:
+  N5k: &loadbal_uni_n5_n6
```
mikewiebe: Same comment here.

tests/test_fabricpath_global.rb
```diff
@@ -42,213 +40,203 @@ def no_feature_fabricpath
...
+      assert_raises(Cisco::UnsupportedError) do
+        fg.aggregate_multicast_routes = true
+      end
+      return
+    end
+
+    fg = FabricpathGlobal.new('default')
+    fg.aggregate_multicast_routes = true
+    assert(fg.aggregate_multicast_routes,
+           'aggregate_multicast_routes: Expected: true')
```
mikewiebe, line 71: Is the failure message really needed here and below?

tests/test_fabricpath_global.rb
```diff
@@ -42,213 +40,203 @@ def no_feature_fabricpath
...
+      assert_raises(Cisco::UnsupportedError) { fg.linkup_delay_always = true }
+    else
+      assert_equal(fg.default_linkup_delay_always, fg.linkup_delay_always)
+      fg.linkup_delay_always = true
+      assert(fg.linkup_delay_always, 'linkup_delay_always: Expected: true')
+      fg.linkup_delay_always = false
+      refute(fg.linkup_delay_always, 'linkup_delay_always: Expected: false')
+    end
+
+    if validate_property_excluded?('fabricpath', 'linkup_delay_enable')
```
mikewiebe, line 111: Could you just combine this with the assert_raises above?  In otherwords, if linkup_delay_always is unsupported, isn't it certain or at least highly likely that linkup_delay_enable is also unsupported?  Regardless, we would catch it if the test fails.

tests/test_fabricpath_global.rb
```diff
@@ -42,213 +40,203 @@ def no_feature_fabricpath
...
-    assert_equal(check_val, @global.loadbalance_algorithm,
-                 "default algo should be #{check_val} but is
-                  #{@global.loadbalance_algorithm}")
-    @global.loadbalance_algorithm = 'source'
-    assert_equal('source', @global.loadbalance_algorithm,
-                 'algo not getting set to source-destination')
+    fg = FabricpathGlobal.new('default')
+    default =
+      cmd_ref.lookup('fabricpath', 'loadbalance_algorithm').default_value
+    assert_equal(default, fg.loadbalance_algorithm)
```
mikewiebe: Why use `cmd_ref.lookup` instead of `default_loadbalance_algorithm`?

lib/cisco_node_utils/cmd_ref/fabricpath.yaml
chrisvanheuveln: We've used this pattern before. The reason is that it is specific to these two platforms. If we bring in another platform with yet another unique value then `loadbal_alg` isn't that meaningful.

tests/test_fabricpath_global.rb
chrisvanheuveln, line 71: Nah, it's just syntactic sugar.
When `assert/refute` fail they don't print out a pretty `Expected: foo / Actual: bar` failure like `assert_equal` does. 
Instead you get `Failed assertion, no message given.` Hence the extra message.

lib/cisco_node_utils/cmd_ref/fabricpath.yaml
mikewiebe: Agreed, but if we add n8k, and n9k then we have to modify `loadbal_lg_n5_n6` to become `loadbal_lg_n5_n6_n8_n9`.  Not that big a deal so not blocking.

tests/test_fabricpath_global.rb
mikewiebe, line 71: ok

tests/test_fabricpath_global.rb
chrisvanheuveln, line 111: I considered it and decided it was better to explicitly check for excluded on each property.

tests/test_fabricpath_global.rb
chrisvanheuveln: Good question. I think that may have been an earlier workaround that I should've removed. I'll fix it.

lib/cisco_node_utils/cmd_ref/fabricpath.yaml
chrisvanheuveln: Changed.

tests/test_fabricpath_global.rb
mikewiebe, line 111: Or..

``` ruby
check_excluded = false
['linkup_delay_always', 'linkup_delay_enable'].each do |prop|
  if validate_property_excluded?('fabricpath', prop)
    assert_raises(Cisco::UnsupportedError) { fg.send("#{prop}=", true) }
    check_excluded = true
  end
end
return if check_excluded
```

chrisvanheuveln, line 111: I'm trying to avoid the problem where one property gets unexcluded while the other does not.

mikewiebe, line 111: ok.

PR #319, opened 2016-03-28
test_platform: fix chassis hash lookup

chrisvanheuveln:
- Tests now pass 100% on n3,n5,n6,n7,n8,n9

### DIFF
--- a/lib/cisco_node_utils/platform.rb
+++ b/lib/cisco_node_utils/platform.rb
@@ -103,7 +103,10 @@ module Cisco
       node.cache_flush # TODO: investigate why this is needed
       all = config_get('inventory', 'chassis')
       return nil if all.nil?
-      chas = all.find { |item| item['name'] == '"Chassis"' }
+
+      # item['name'] value is "\"Chassis\"" on most platforms,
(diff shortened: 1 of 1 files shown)


lib/cisco_node_utils/platform.rb
```diff
@@ -103,7 +103,10 @@ def self.chassis
       node.cache_flush # TODO: investigate why this is needed
       all = config_get('inventory', 'chassis')
       return nil if all.nil?
-      chas = all.find { |item| item['name'] == '"Chassis"' }
+
+      # item['name'] value is "\"Chassis\"" on most platforms,
+      # but n5k/n6k use "Chassis" so make the quotes optional.
+      chas = all.find { |item| item['name'][/^"*Chassis"*$/] }
```
robert-w-gries: Nit: I would prefer `?` to `*` for making the quotation marks optional

chrisvanheuveln:
```
* = zero or more
? = zero or one
```

I don't really care so I'll change it for you. :-)

mikewiebe:
:+1:

PR #320, opened 2016-03-29
test_nxapi fixes

chrisvanheuveln:
- Add platform specific values
- Remove ethernet references
- Tests now pass on n30,n31,n5,n6,n7,n8,n9-I2,n9-I3

### DIFF
--- a/tests/test_nxapi.rb
+++ b/tests/test_nxapi.rb
@@ -15,6 +15,9 @@
 # limitations under the License.
 
 require_relative 'basetest'
+require_relative '../lib/cisco_node_utils/node'
+
+include Cisco
 
 # TestNxapi - NXAPI client unit tests
 class TestNxapi < TestCase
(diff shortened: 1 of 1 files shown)


robert-w-gries:
Only suggestion I have is to add a `teardown` method to remove loopback interfaces

:+1: LGTM

chrisvanheuveln:
Agreed. I'll add.

tests/test_nxapi.rb
```diff
@@ -15,6 +15,9 @@
 # limitations under the License.
 
 require_relative 'basetest'
+require_relative '../lib/cisco_node_utils/node'
+
+include Cisco
```
mikewiebe: Why do you need this include?

chrisvanheuveln: For the Node call here:

https://github.com/cisco/cisco-network-node-utils/pull/320/files#diff-02a78d01adf8c91e7e322c0263fb9f20R39

...but I can just add the module ref to the call to avoid the include. I'll add it.

PR #321, opened 2016-03-29
tests_node_ext: platform fixes

chrisvanheuveln:
- Tests now pass on n30,n31,n5,n6,n7,n8,n9-I2,n9-I3
- Known image issue with last_reason_time on n8,n9_I2

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/show_version.yaml
+++ b/lib/cisco_node_utils/cmd_ref/show_version.yaml
@@ -50,18 +50,19 @@ last_reset_reason:
   get_value: "rr_reason"
 
 last_reset_time:
-  _exclude: [ios_xr]
+  _exclude: [N3k, ios_xr]
   kind: string
   default_value: ''
   get_value: "rr_ctime"
 
 system_image:
-  nexus:
-    get_value: "kick_file_name"
+  get_value: "kick_file_name"
+  N5k: &sys_img_isan
+    get_value: "isan_file_name"
+  N6k: *sys_img_isan
+  N7k: *sys_img_isan
   ios_xr:
     get_value: '/IOS XR.*Version (.*)$/'
-  N7k:
-    get_value: "isan_file_name"
 
 uptime:
   # TODO: redundant with show_system, uptime?
--- a/lib/cisco_node_utils/cmd_ref/system.yaml
+++ b/lib/cisco_node_utils/cmd_ref/system.yaml
@@ -1,8 +1,7 @@
 # system
 ---
 resources:
-  nexus:
-    get_command: "show system resources"
-  # ios_xr:
-    # grpc doesn't allow this command
-    # get_command: 'monitor threads iteration 1'
+  # grpc doesn't allow this command
+  # get_command: 'monitor threads iteration 1'
+  _exclude: [ios_xr]
+  get_command: "show system resources"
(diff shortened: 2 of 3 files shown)


lib/cisco_node_utils/cmd_ref/show_version.yaml
```diff
@@ -50,18 +50,19 @@ last_reset_reason:
...
   get_value: "rr_ctime"
 
 system_image:
-  nexus:
-    get_value: "kick_file_name"
+  get_value: "kick_file_name"
+  N5k: &sys_img_isan
+    get_value: "isan_file_name"
+  N6k: *sys_img_isan
+  N7k: *sys_img_isan
```
robert-w-gries, line 63: Can you remind me what the `&` and `*` symbols do here?

lib/cisco_node_utils/cmd_ref/system.yaml
```diff
@@ -1,8 +1,7 @@
 # system
 ---
 resources:
-  nexus:
-    get_command: "show system resources"
-  # ios_xr:
-    # grpc doesn't allow this command
-    # get_command: 'monitor threads iteration 1'
+  # grpc doesn't allow this command
+  # get_command: 'monitor threads iteration 1'
```
robert-w-gries, line 5: @glennmatthews @chburket Does this comment still apply?

lib/cisco_node_utils/cmd_ref/show_version.yaml
chrisvanheuveln, line 63: `&sys_img_isan` is just a tag. `*sys_img_isan` uses the values found at that tag.
It allows us to specify the same value for all three of those platforms.

robert-w-gries:
:+1:

lib/cisco_node_utils/cmd_ref/show_version.yaml
mikewiebe, line 63: @robert-w-gries , YAML anchors and aliases: See: https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/cmd_ref/README_YAML.md#yaml-anchors-and-aliases

robert-w-gries, line 63: That's neat.  It's sometimes difficult to remember all of the nice features in our YAML :-)

lib/cisco_node_utils/cmd_ref/system.yaml
glennmatthews, line 5: Yes it does. grpc blocks all exec commands that don't start with 'show '.

(Edit) What's the purpose of this change, though - why not leave the file alone?

chrisvanheuveln, line 5: It was okay before, it's better now. :-)

Kidding... I was troubleshooting one of the tests and just changed it since I prefer the explicit _exclude over the implicit "unsupported". But I'll change it back if you'd prefer.

glennmatthews, line 5: No biggie, just wondering. Was this needed to get the 'validate_property_excluded' check to work?

chrisvanheuveln, line 5: Nope. The validate works with both syntax types.

PR #322, opened 2016-03-29
test_bgp_af.rb platform fixes

chrisvanheuveln:
- Now passes on n30,n31,n5,n6,n7,n9-I2,n9-I3. Still needs validation on n8k,ios_xr.
- Added conditional feature enable for evpn safi's
- Added validate_property_unsupported check to the big property_matrix test, so I also removed the old :unsupported checks
- Added config_no_warn() for the XR pre-req calls

### DIFF
--- a/lib/cisco_node_utils/bgp_af.rb
+++ b/lib/cisco_node_utils/bgp_af.rb
@@ -59,7 +59,10 @@ module Cisco
     end
 
     def create
-      Feature.bgp_enable if platform == :nexus
+      if platform == :nexus
+        Feature.bgp_enable
+        Feature.nv_overlay_evpn_enable if @safi[/evpn/]
+      end
       set_args_keys(state: '')
       config_set('bgp', 'address_family', @set_args)
     end
--- a/tests/test_bgp_af.rb
+++ b/tests/test_bgp_af.rb
@@ -352,6 +352,10 @@ class TestBgpAF < CiscoTestCase
 
     # Needed for testing route-policy commands
     config_no_warn('route-policy drop_all', 'end-policy')
+
+    # TBD: Reduce the number of different route-policies used by this test
+    # and move the remaining ones into this method. Then create an
+    # unconfig_ios_xr_dependencies to handle the cleanups.
   end
 
   ########################################################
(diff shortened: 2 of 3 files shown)


chrisvanheuveln:
@chburket Charles, I made some changes that affect the XR side. Please take a look. Thanks.

chburket:
Looks good to me, but I'd like @brturn to take a look, since he knows the matrix code best.

Are you planning to run against ios_xr?

chrisvanheuveln:
Yes. Also still need to validate against n8k. I'll make sure it's working on both before merging.

brturn:
Looks good. +1

chburket:
:+1:

tests/test_bgp_af.rb
```diff
@@ -829,6 +833,6 @@ def test_table_map
     afs.each do |af|
       table_map(55, 'red', af)
     end
-    config('no route-policy sjc') if platform == :ios_xr
+    config_no_warn('no route-policy sjc') if platform == :ios_xr
```
chburket: Out of curiosity, why does this generate a warning?  You created it on line 828, so you know it will exist, right?

lib/cisco_node_utils/bgp_af.rb
```diff
@@ -59,7 +59,10 @@ def self.afs
     end
 
     def create
-      Feature.bgp_enable if platform == :nexus
+      if platform == :nexus
+        Feature.bgp_enable
+        Feature.nv_overlay_evpn_enable if @safi[/evpn/]
```
mikewiebe, line 64: With this conditional enable, do we still need this?

https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/bgp_af.rb#L157

tests/test_bgp_af.rb
chrisvanheuveln: Good point. Took me a minute to see what's happening here. It's making noise because the route-policy can't be deleted while bgp still has a table-policy linked to it. 

The reason it's still linked is due to this part of the test:

```
819    val = false
820    bgp_af.table_map_set('sjc', val)
821    refute(bgp_af.table_map_filter)
822
823    val = true
824    # TODO?  XR does not appear to support 'table-policy sjc filter'
825    if platform == :ios_xr

```

...the test on line 821 wants to test for no filter, but the set on 820 has a side-effect of creating a `table-policy sjc` config. The next thing it does after that is bail out since ios_xr doesn't support filters (so the `table-policy sjc` config is not cleaned up). 

So I need to:
1) Add a validate_property check prior to that test
2) revert the config_no_warn() back to config()

lib/cisco_node_utils/bgp_af.rb
chrisvanheuveln, line 64: Yes, because `advertise_l2vpn_evpn` can exist in a unicast safi.

The change above is necessary for creating `address-family l2vpn evpn`.

mikewiebe, line 64: Ok.

tests/test_bgp_af.rb
chburket: Sounds good.  For a another day, might be better to move the creation of all test route_policies into `config_ios_xr_dependencies`, and then create an `unconfig_ios_xr_dependencies` to remove them.  Seems like those methods could then be called in `setup` and `teardown`.  A task for another PR, though.

mikewiebe:
:+1:

PR #323, opened 2016-03-30
Feature/private vlan interface

davidefdl:
This code drop addresses private vlan interface properties for eth interfaces and svis'

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@ Changelog
   * 'switchport_private_vlan_trunk_allowed_vlan', 'switchport_private_vlan_trunk_native_vlan'
   * 'switchport_mode_private_vlan_host', 'switchport_mode_private_vlan_host_association'
   * 'switchport_mode_private_vlan_host_promiscous', 'switchport_mode_private_vlan_trunk'
+  * 'switchport_private_vlan_association_trunk', 'switchport_private_vlan_mapping_trunk'
 * Extend vpc with vpc+ attributes on Nexus 5k/6k/7k:
   * `fabricpath_emulated_switch_id`
   * `fabricpath_multicast_load_balance` (only on Nexus 7k)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,9 @@ Changelog
   * `ipv4_forwarding`
   * `stp_bpdufilter`, `stp_bpduguard`, `stp_cost`, `stp_guard`, `stp_link_type`, `stp_mst_cost`
   * `stp_mst_port_priority`, `stp_port_priority`, `stp_port_type`, `stp_vlan_cost`, `stp_vlan_port_priority`
+  * 'switchport_private_vlan_trunk_allowed_vlan', 'switchport_private_vlan_trunk_native_vlan'
+  * 'switchport_mode_private_vlan_host', 'switchport_mode_private_vlan_host_association'
+  * 'switchport_mode_private_vlan_host_promiscous', 'switchport_mode_private_vlan_trunk'
 * Extend vpc with vpc+ attributes on Nexus 5k/6k/7k:
   * `fabricpath_emulated_switch_id`
   * `fabricpath_multicast_load_balance` (only on Nexus 7k)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ Changelog
   * 'switchport_mode_private_vlan_host', 'switchport_mode_private_vlan_host_association'
   * 'switchport_mode_private_vlan_host_promiscous', 'switchport_mode_private_vlan_trunk'
   * 'switchport_private_vlan_association_trunk', 'switchport_private_vlan_mapping_trunk'
+  * 'private_vlan_mapping'
 * Extend vpc with vpc+ attributes on Nexus 5k/6k/7k:
   * `fabricpath_emulated_switch_id`
   * `fabricpath_multicast_load_balance` (only on Nexus 7k)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,7 +50,7 @@ Changelog
   * `stp_mst_port_priority`, `stp_port_priority`, `stp_port_type`, `stp_vlan_cost`, `stp_vlan_port_priority`
   * `switchport_private_vlan_trunk_allowed_vlan`, `switchport_private_vlan_trunk_native_vlan`
   * `switchport_mode_private_vlan_host`, `switchport_mode_private_vlan_host_association`
-  * `switchport_mode_private_vlan_host_promiscous`, `switchport_mode_private_vlan_trunk`
+  * `switchport_mode_private_vlan_host_promiscous`, `switchport_mode_private_vlan_trunk_promiscous`, `switchport_mode_private_vlan_trunk_secondary`
   * `switchport_private_vlan_association_trunk`, `switchport_private_vlan_mapping_trunk`
   * `private_vlan_mapping`
 * Extend vpc with vpc+ attributes on Nexus 5k/6k/7k:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,11 +48,11 @@ Changelog
   * `ipv4_forwarding`
   * `stp_bpdufilter`, `stp_bpduguard`, `stp_cost`, `stp_guard`, `stp_link_type`, `stp_mst_cost`
   * `stp_mst_port_priority`, `stp_port_priority`, `stp_port_type`, `stp_vlan_cost`, `stp_vlan_port_priority`
-  * 'switchport_private_vlan_trunk_allowed_vlan', 'switchport_private_vlan_trunk_native_vlan'
-  * 'switchport_mode_private_vlan_host', 'switchport_mode_private_vlan_host_association'
-  * 'switchport_mode_private_vlan_host_promiscous', 'switchport_mode_private_vlan_trunk'
-  * 'switchport_private_vlan_association_trunk', 'switchport_private_vlan_mapping_trunk'
-  * 'private_vlan_mapping'
+  * `switchport_private_vlan_trunk_allowed_vlan`, `switchport_private_vlan_trunk_native_vlan`
+  * `switchport_mode_private_vlan_host`, `switchport_mode_private_vlan_host_association`
+  * `switchport_mode_private_vlan_host_promiscous`, `switchport_mode_private_vlan_trunk`
+  * `switchport_private_vlan_association_trunk`, `switchport_private_vlan_mapping_trunk`
+  * `private_vlan_mapping`
 * Extend vpc with vpc+ attributes on Nexus 5k/6k/7k:
   * `fabricpath_emulated_switch_id`
   * `fabricpath_multicast_load_balance` (only on Nexus 7k)
--- a/lib/cisco_node_utils/cmd_ref/interface.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface.yaml
@@ -412,6 +412,20 @@ switchport_mode_private_vlan_trunk:
   set_value: "<state> switchport mode private-vlan trunk <mode>"
   default_value: ""
 
+switchport_private_vlan_trunk_allowed_vlan:
+  _exclude: [ios_xr, N8k]
+  multiple: true
+  get_value: '/^switchport private-vlan trunk allowed vlan (.*)$/'
+  set_value: "<state> switchport private-vlan trunk allowed vlan <oper> <vlans>"
+  default_value: []
+
+switchport_private_vlan_trunk_native_vlan:
+  _exclude: [ios_xr, N8k]
+  multiple: true
+  get_value: '/^switchport private-vlan trunk native vlan (.*)$/'
+  set_value: "<state> switchport private-vlan trunk native vlan <vlan>"
+  default_value: ''
+
 switchport_trunk_allowed_vlan:
   _exclude: [ios_xr]
   get_value: '/^switchport trunk allowed vlan (.*)$/'
--- a/lib/cisco_node_utils/cmd_ref/interface.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface.yaml
@@ -385,6 +385,18 @@ switchport_mode_port_channel:
   set_value: "<state> switchport mode <mode>"
   default_value: ""
 
+switchport_mode_private_vlan_host:
+  _exclude: [ios_xr, N8k]
+  get_value: '/^switchport mode private-vlan (.*)$/'
+  set_value: "<state> switchport mode private-vlan <mode>"
+  default_value: ""
+
+switchport_mode_private_vlan_trunk:
+  _exclude: [ios_xr, N8k]
+  get_value: '/^switchport mode private-vlan trunk (.*)$/'
+  set_value: "<state> switchport mode private-vlan trunk <mode>"
+  default_value: ""
+
 switchport_trunk_allowed_vlan:
   _exclude: [ios_xr]
   get_value: '/^switchport trunk allowed vlan (.*)$/'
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -48,6 +48,11 @@ Changelog
   * `ipv4_forwarding`
   * `stp_bpdufilter`, `stp_bpduguard`, `stp_cost`, `stp_guard`, `stp_link_type`, `stp_mst_cost`
   * `stp_mst_port_priority`, `stp_port_priority`, `stp_port_type`, `stp_vlan_cost`, `stp_vlan_port_priority`
+  * 'switchport_private_vlan_trunk_allowed_vlan', 'switchport_private_vlan_trunk_native_vlan'
```
chrisvanheuveln: These should be backticks not quotes.

lib/cisco_node_utils/cmd_ref/interface.yaml
```diff
@@ -210,6 +210,13 @@ negotiate_auto_portchannel:
   set_value: "<state> negotiate auto"
   default_value: true
 
+private_vlan_mapping:
+  _exclude: [ios_xr, N8k]
```
chrisvanheuveln, line 218: Are these new properties really not supported on n8k or are you excluding because they still need to be tested on n8k?

lib/cisco_node_utils/cmd_ref/interface.yaml
```diff
@@ -383,6 +390,61 @@ switchport_mode_port_channel:
...
+  get_value: '/^switchport private-vlan trunk allowed vlan (.*)$/'
+  set_value: "<state> switchport private-vlan trunk allowed vlan <oper> <vlans>"
+  default_value: []
+
+switchport_private_vlan_trunk_native_vlan:
+  _exclude: [ios_xr, N8k]
+  multiple: true
+  get_value: '/^switchport private-vlan trunk native vlan (.*)$/'
+  set_value: "<state> switchport private-vlan trunk native vlan <vlan>"
+  default_value: ''
```
chrisvanheuveln: I think the default native vlan is normally 1 ?

lib/cisco_node_utils/interface.rb
```diff
@@ -903,6 +914,444 @@ def switchport_trunk_native_vlan=(val)
       end
     end
 
+    # Interface private vlan configuration properties
+    # private vlan mode (host, promisc, trunk promisc , trunk sec)
+    # private vlan mapping
+    # private vlan native vlan
+    # private vlan allow vlan
+
+    def cli_error_check(result)
```
chrisvanheuveln, line 937: Let's not duplicate this here. Better to either:

(a) Move it to cisco_cmn_utils.rb
-or-
(b) Make the cli_error_check in vlan.rb a class method

My vote is (a).

lib/cisco_node_utils/interface.rb
```diff
@@ -903,6 +914,444 @@ def switchport_trunk_native_vlan=(val)
...
+
+    def set_args_keys_default
+      keys = { name: @name }
+      @set_args = keys
+    end
+
+    def set_args_keys(hash={}) # rubocop:disable Style/AccessorMethodName
+      set_args_keys_default
+      @set_args = @set_args.merge!(hash) unless hash.empty?
+    end
```
chrisvanheuveln: I'm normally a fan of these `set_args_` methods but in this case the only value they add is not having to specify `:name`. There's no other context. Also, the rest of the methods in this very large provider do not use these helpers. It's better to code to the surrounding code. 

If this was a new provider and all the set methods used `set_args_` then I would probably overlook it. In this case it's simpler and more obvious to just specify `:name` and the other values directly in the `config_set` call, so I'd prefer to have them removed.

lib/cisco_node_utils/interface.rb
```diff
@@ -903,6 +914,444 @@ def switchport_trunk_native_vlan=(val)
...
+
+    def switchport_mode_private_vlan_host_lookup_string
+      case @name
+      when ETHERNET
+        return 'switchport_mode_private_vlan_host'
+      end
```
chrisvanheuveln: 1) Are you expecting more name types to match against? Just wondering why you used `case` instead of `if`
2) Why uppercase ETHERNET? We downcase it in the initialize method. How does this match anything?
3) I don't understand the benefit of having these separate methods that all either return a single string name or nil. I don't think they're needed.

lib/cisco_node_utils/interface.rb
```diff
@@ -903,6 +914,444 @@ def switchport_trunk_native_vlan=(val)
...
+        set_args_keys(state: '', mode: IF_SWITCHPORT_MODE[mode_set])
+        config_set('interface', switchport_mode_private_vlan_host_lookup_string,
+                   @set_args)
+      when :disabled
+        set_args_keys(state: 'no', mode: '')
+        config_set('interface', switchport_mode_private_vlan_host_lookup_string,
+                   @set_args)
+
+      end
+    end
```
chrisvanheuveln: This can be simplified to just:

```
    def switchport_enable_and_mode_private_vlan_host(mode_set)
      switchport_enable unless switchport
      if mode_set[/(host|promiscuous)/]
        state = ''
        mode = IF_SWITCHPORT_MODE[mode_set]
      else
        state = 'no'
        mode = ''
      end
      config_set('interface', 'switchport_mode_private_vlan_host',
                  name: @name, state: state, mode: mode)
    end
```

chrisvanheuveln: ditto for similar methods

lib/cisco_node_utils/cmd_ref/interface.yaml
davidefdl, line 218: Yes there are not supported for private vlan

lib/cisco_node_utils/cmd_ref/interface.yaml
davidefdl: yes done

lib/cisco_node_utils/interface.rb
davidefdl, line 937: The two methods are not actually same. For interface the error is in body[1] for vln in body[2] and for interfaces we do not have a warning msg

lib/cisco_node_utils/interface.rb
davidefdl: done

lib/cisco_node_utils/interface.rb
davidefdl: yes in the feature I need to add port channel anf fex

CHANGELOG.md
davidefdl: done

lib/cisco_node_utils/interface.rb
davidefdl: For the ETHERNET I was simply using the existing code that is present in interface.rb.
Plus I believe this is just a regex having all the types
ETHERNET = Regexp.new('(Ethernet|GigE|MgmtEth)', Regexp::IGNORECASE)
so it's not actually the name.name property

davidefdl: For your third comment again since I'm updating the interface.rb I was just following some existing code inside the file...just trying to follow the guideline present in the file.

lib/cisco_node_utils/interface.rb
davidefdl: Done

lib/cisco_node_utils/interface.rb
chrisvanheuveln: Okay, ETHERNET makes sense. 
It's still not clear if these name string lookups are going to be different for each interface type. I also have some other questions. So let's have a phone call after scrum today and walk through this review.

tests/test_interface_private_vlan.rb
```diff
@@ -0,0 +1,836 @@
...
+    config('feature private-vlan', 'no feature vtp')
+    @interface = Interface.new(interfaces[0])
+  end
+
+  def teardown
+    config("default interface ethernet #{interfaces_id[0]}")
+    config('no feature private-vlan')
+    super
+  end
+end
```
chrisvanheuveln, line 57: Why the separate class here? Also, you setup & teardown should have similar cleanups.

edit: Okay I see what you're doing here. It's fine. But you still need to fix setup / teardown.

Ref:
https://github.com/cisco/cisco-network-node-utils/blob/develop/tests/test_ace.rb#L22

```
def setup
  super
 @interface = Interface.new(interfaces[0])
  cleanup if @@pre_clean_needed
  @@pre_clean_needed = false
end

def teardown
  cleanup
  config("default interface ethernet #{interfaces_id[0]}")
  super
end

def cleanup
  # ADD: remove  pvlans 42 & 43 from each interface
  #   Interface.all_pvlan_intfs.each { |i|      config("default interface ethernet #{i}")  }

  # ADD: remove all svi 42 & 43
  #   Interface.all_interfaces.each { |i|   config("no interface #{i}" if i[/vlan/] }

  remove_vlan_vlans
  config('no feature private-vlan', 'no feature vtp')
end
```

tests/test_interface_private_vlan.rb
```diff
@@ -0,0 +1,836 @@
...
+class TestSwitchport < TestInterfaceSwitchport
+  def test_interface_switchport_private_host_mode
+    if validate_property_excluded?('interface',
+                                   'switchport_mode_private_vlan_host')
+      assert_nil(interface.switchport_mode_private_vlan_host)
+      assert_raises(Cisco::UnsupportedError) do
+        interface.switchport_mode_private_vlan_host = :host
+      end
+      return
+    else
```
chrisvanheuveln: This should be `end` instead of `else` since you've got a return at the end of the block.
Ditto for all of your methods.

tests/test_interface_private_vlan.rb
```diff
@@ -0,0 +1,836 @@
...
+        :host,
+        :promiscuous,
+      ]
+
+      switchport_modes.each do |start|
+        begin
+          interface.switchport_mode_private_vlan_host = start
+          assert_equal(start, interface.switchport_mode_private_vlan_host,
+                       "Err: Switchport mode, #{start}, not as expected")
+        rescue Cisco::CliError
```
chrisvanheuveln: Why the rescue? We should not be silently ignoring CliErrors.
If you expect a CliError then you should explicitly test for it with `assert_raise`. If it's a problem with unsupported line cards then you can use a skip check like this: https://github.com/cisco/cisco-network-node-utils/blob/develop/tests/ciscotest.rb#L130
...in which case you would also want to split the test for :host and :promiscuous into two separate test cases.

tests/test_interface_private_vlan.rb
```diff
@@ -0,0 +1,836 @@
...
+        interface.switchport_mode_private_vlan_host = :host
+      end
+      return
+    else
+      v1 = Vlan.new(10)
+      pv_type = 'primary'
+      v1.private_vlan_type = pv_type
+      v2 = Vlan.new(11)
+      pv_type = 'community'
+      v2.private_vlan_type = pv_type
```
chrisvanheuveln: nit: the `pv_type` variable is unnecessary.

tests/test_interface_private_vlan.rb
```diff
@@ -0,0 +1,836 @@
...
+      v1.private_vlan_association = ['11']
+      input = %w(10 11)
+      result = %w(10 11)
```
chrisvanheuveln: nit: you don't need separate variables for `input` and `result` since your input value is your expected value. Just use one, e.g.  `assoc = %w(10 11)`

tests/test_interface_private_vlan.rb
```diff
@@ -0,0 +1,836 @@
...
+      v1 = Vlan.new(10)
+      pv_type = 'primary'
+      v1.private_vlan_type = pv_type
+      v2 = Vlan.new(11)
+      pv_type = 'community'
+      v2.private_vlan_type = pv_type
+      v1.private_vlan_association = ['11']
+
+      input = %w(10 11)
+      result = %w(10 11)
```
chrisvanheuveln: ditto nits: unnecessary `pv_type`, `input`, `result` throughout this file.

edit: for array inputs use:   `result = input.clone`

davidefdl: done

tests/test_interface_private_vlan.rb
davidefdl: done

tests/test_interface_private_vlan.rb
davidefdl: done

tests/test_interface_private_vlan.rb
davidefdl: done

tests/test_interface_private_vlan.rb
davidefdl: done

tests/test_interface_private_vlan.rb
davidefdl, line 57: done

lib/cisco_node_utils/interface.rb
```diff
@@ -58,18 +69,32 @@ def to_s
...
         hash[id] = Interface.new(id, false)
       end
       hash
     end
 
+    # General-purpose filter for Interface.interfaces().
+    #    opt: Identifies the filter. This may be overloaded in the future to
+    #         allow a hash of filter conditions.
+    #     id: The interface name
+    # Return: true if the interface should be filtered out, false to keep it.
```
chrisvanheuveln, line 89: I like this change. :-)

lib/cisco_node_utils/interface.rb
```diff
@@ -903,6 +928,442 @@ def switchport_trunk_native_vlan=(val)
...
+
+    def switchport_enable_and_mode_private_vlan_host(mode_set)
+      switchport_enable unless switchport
+      if mode_set[/(host|promiscuous)/]
+        config_set('interface', 'switchport_mode_private_vlan_host',
+                   name: @name, state: '', mode: IF_SWITCHPORT_MODE[mode_set])
+      else
+        config_set('interface', 'switchport_mode_private_vlan_host',
+                   name: @name, state: 'no', mode: IF_SWITCHPORT_MODE[mode_set])
+      end
```
chrisvanheuveln, line 964: I still prefer having a single `config_set` and using variables for `state` and `mode`. 
But I grudgingly accept your approach.

lib/cisco_node_utils/interface.rb
```diff
@@ -903,6 +928,442 @@ def switchport_trunk_native_vlan=(val)
...
+      is_list = prepare_array(switchport_private_vlan_trunk_allowed_vlan)
+      configure_private_vlan_host_property(:allow_vlan, vlans,
+                                           is_list, '')
+    end
+
+    def switchport_private_vlan_trunk_native_vlan
+      result = config_get('interface',
+                          'switchport_private_vlan_trunk_native_vlan',
+                          name: @name)
+      result
```
chrisvanheuveln: This assignment to `result` is unnecessary

tests/test_interface_private_vlan.rb
```diff
@@ -21,19 +21,43 @@
...
+  def remove_svis
+    Interface.interfaces.each do |name, _i|
+      next unless name[/vlan/]
+      if name.match(/\bvlan1\b/)
+        next
+      else
+        config("no interface #{name}")
+      end
+    end
+  end
```
chrisvanheuveln, line 51: This is cleaner:

```
  def remove_svis
    Interface.interfaces.each do |name, _i|
      next if name.match(/^vlan1$/) || !name[/vlan/]
      config("no interface #{name}")
    end
  end
```

lib/cisco_node_utils/interface.rb
davidefdl: done

tests/test_interface_private_vlan.rb
davidefdl, line 51: done

PR #326, opened 2016-03-30
aaa_authorization_service test fix

robert-w-gries:
## Changes
- Properly clean up after tests
- Use `config` to clean up aaa configurations since `nxapi` cleanup is prohibited by nature of the provider
- Make tacacs_group method for retrieving group names
- Split up tests so that the configs do not overlap with each other
- Allow for single group configuration in group_method_set
## Test Results

|  | N3k | N5k | N6k | N7k | N8k | N9k I2 | N9k I3 | XR |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| test_aaa_authorization | ✅ | ✅ | ✅ | ✅ | ? | ✅ | 6 F<br>12 E | Skipped |

### DIFF
--- a/tests/test_aaa_authorization_service.rb
+++ b/tests/test_aaa_authorization_service.rb
@@ -25,8 +25,27 @@ class TestAaaAuthorizationService < CiscoTestCase
   end
 
   def teardown
+    cleanup_aaa
     feature_tacacs(false)
     super
+  end
+
+  def cleanup_aaa
+    cmds = config('show run aaa').scan(/^aaa auth.*/)
+    cmds.each do |cmd|
+      config("no #{cmd}")
+    end
+  end
+
+  def config_tacacs_servers(servers)
+    servers.each do |server|
+      config("aaa group server tacacs+ #{server}")
+    end
+  end
+
+  def feature_tacacs(feature=true)
+    state = feature ? '' : 'no'
+    config("#{state} feature tacacs+")
   end
 
   # Method to pre-configure a valid tacacs server and aaa group.  This
@@ -46,28 +65,12 @@ class TestAaaAuthorizationService < CiscoTestCase
     end
   end
 
-  def feature_tacacs(feature=true)
-    if feature
-      config('feature tacacs')
-    else
-      config('no feature tacacs',
-             'no aaa authentication login ascii-authentication')
-    end
-  end
-
-  def config_tacacs_servers(servers)
-    config('feature tacacs+')
-    servers.each do |server|
-      config("aaa group server tacacs+ #{server}")
-    end
+  def prefix
+    'aaa authorization'
   end
 
   def show_cmd
     'show run aaa all | no-more'
-  end
-
-  def prefix
-    'aaa authorization'
   end
 
   def test_create_unsupported_type
--- a/tests/test_aaa_authorization_service.rb
+++ b/tests/test_aaa_authorization_service.rb
@@ -18,9 +18,12 @@ require_relative '../lib/cisco_node_utils/aaa_authorization_service'
 # TestAaaAuthorizationService - Minitest for AaaAuthorizationService util
 class TestAaaAuthorizationService < CiscoTestCase
   @skip_unless_supported = 'aaa_authorization_service'
+  @@pre_clean_needed = true # rubocop:disable Style/ClassVars
 
   def setup
     super
+    cleanup_aaa if @@pre_clean_needed
+    @@pre_clean_needed = false # rubocop:disable Style/ClassVars
     feature_tacacs
     preconfig_tacacs_server_access(tacacs_groups[0])
     config_tacacs_servers(tacacs_groups[1..3])
--- a/tests/test_aaa_authorization_service.rb
+++ b/tests/test_aaa_authorization_service.rb
@@ -22,6 +22,8 @@ class TestAaaAuthorizationService < CiscoTestCase
   def setup
     super
     feature_tacacs
+    preconfig_tacacs_server_access(tacacs_groups[0])
+    config_tacacs_servers(tacacs_groups[1..3])
   end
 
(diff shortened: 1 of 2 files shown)


tests/test_aaa_authorization_service.rb
```diff
@@ -22,54 +22,72 @@ class TestAaaAuthorizationService < CiscoTestCase
   def setup
     super
     feature_tacacs
```
chrisvanheuveln, line 27: Your setup needs an initial cleanup as well. We've been using `@@pre_cleanup_needed` as a flag to avoid having setup repeat cleanups after teardown runs, but you still need that very first pre clean. See: https://github.com/cisco/cisco-network-node-utils/blob/develop/tests/test_ace.rb#L22

tests/test_aaa_authorization_service.rb
```diff
@@ -22,54 +22,72 @@ class TestAaaAuthorizationService < CiscoTestCase
...
+    preconfig_tacacs_server_access(tacacs_groups[0])
+    config_tacacs_servers(tacacs_groups[1..3])
   end
 
   def teardown
+    cleanup_aaa
     feature_tacacs(false)
```
chrisvanheuveln, line 34: Can you move the feature disablement into `cleanup_aaa` ?

tests/test_aaa_authorization_service.rb
```diff
@@ -22,54 +22,72 @@ class TestAaaAuthorizationService < CiscoTestCase
...
+    p << ' ' + cmd_type
+    p << ' ' + service
+    p << ' group ' + groups unless groups.empty?
+    p << ' ' + method unless method.empty?
+    Regexp.new(p)
+  end
+
+  # Method to pre-configure a valid tacacs server and aaa group.  This
+  # group can be included in the testing such access to the device
+  # never is compromised.
```
chrisvanheuveln, line 71: nit: This comment is Yoda very.

:neckbeard:

tests/test_aaa_authorization_service.rb
```diff
@@ -22,54 +22,72 @@ class TestAaaAuthorizationService < CiscoTestCase
   def setup
     super
     feature_tacacs
+    preconfig_tacacs_server_access(tacacs_groups[0])
```
mikewiebe, line 28: Can you just combine `preconfig_tacacs_server_access` and `config_tacacs_servers` into a single method?  What is the rational for separating this out?

tests/test_aaa_authorization_service.rb
```diff
@@ -632,240 +537,144 @@ def test_get_default_groups
...
-    method = :local
-    groups = [group0, group1, group3]
-    aaa_a_service.groups_method_set(groups, method)
+  def test_commands_default_local_multi
+    aaa_a_service = AaaAuthorizationService.new(:commands, 'default')
+    aaa_a_service.groups_method_set(tacacs_groups[0..2], :local)
 
-    group_str = "group #{group0} #{group1} #{group3}"
-    p = /#{prefix} #{type_str} #{service} #{group_str} local/
+    p = get_pattern(:commands, 'default', tacacs_groups[0..2], :local)
```
mikewiebe, line 563: Is this right?  The original code used group 0, 1 and 3.

tests/test_aaa_authorization_service.rb
robert-w-gries, line 28: I did not write the original method, but it seems that the difference is that `preconfig_tacacs_server_access` creates a `tacacs-server` that the `aaa group server` has access to.  Not sure why it's necessary, but I kept it in.

tests/test_aaa_authorization_service.rb
robert-w-gries, line 27: Added

tests/test_aaa_authorization_service.rb
robert-w-gries, line 34: Is this a necessary change?  I'm not sure what the purpose is.

tests/test_aaa_authorization_service.rb
robert-w-gries, line 563: I changed it to use groups 0, 1, and 2 because `[0..2]` was faster than typing each individual group :-)

tests/test_aaa_authorization_service.rb
mikewiebe, line 28: The creation of the tacacs-server is needed for the remaining tests, but my point is that it should be fairly trivial to combine these two into a single setup method.

robert-w-gries, line 28: I prefer having the helper methods separated to keep them smaller and single-purpose

tests/test_aaa_authorization_service.rb
chrisvanheuveln, line 34: You call cleanup_aaa from both setup and teardown. But you only do the feature disable from teardown? If it makes sense for teardown then it should also occur for setup. You can't assume that every testbed starts out clean.

Unless you plan to use cleanup_aaa for something beyond setup & teardown then it makes sense to include the feature disable inside cleanup_aaa.

robert-w-gries, line 34: I think I'll remove the feature disablement instead since I don't think it's helpful in cleanup.

Removing it decreases test time from 228.7 seconds to 92.8 seconds.

tests/test_aaa_authorization_service.rb
mikewiebe, line 28: As apposed to the single purpose of setting up the test?  They are quite small already but not that big a deal! :smile:

mikewiebe:
:+1:

tests/test_aaa_authorization_service.rb
chrisvanheuveln, line 34: Fine with me.

PR #330, opened 2016-04-04
Added IOS XR support for tacacs_server_group

jonnytdevops:


jonnytdevops:
```
TestTacacsServerGroup#test_create_destroy_multiple = 36.14 s = .
TestTacacsServerGroup#test_create_destroy_single = 23.79 s = .
TestTacacsServerGroup#test_create_destroy_single_ipv6 = 20.24 s = .

Finished in 83.396458s, 0.0360 runs/s, 0.4796 assertions/s.

3 runs, 40 assertions, 0 failures, 0 errors, 0 skips
```

chburket:
<removed>

jonnytdevops:
@chburket I forgot to upload the new style of test (which is what these results refer to). Going to do that now.

jonnytdevops:
@chburket new test style uploaded, thanks

tests/test_tacacs_server_group_style2.rb
```diff
@@ -0,0 +1,151 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require File.expand_path('../ciscotest', __FILE__)
+require File.expand_path('../../lib/cisco_node_utils/tacacs_server_group', \
+                         __FILE__)
+
+# TestTacacsServerGroup - Minitest for TacacsServerGroup node utility.
+class TestTacacsServerGroup < CiscoTestCase
```
chburket: Same comment as the other PR:

Can you rename this class to something unique (e.g. TestTacacsServerGroupStyle2)? Having the class named the same as the class in test_tacacs_server_group.rb is a bit confusing (difficult to tell which test was run by looking at the output).

jonnytdevops:
@chburket Code updated, I'll squash after a +1, thanks

chburket:
I don't see any changes to test_tacacs_server_group.rb.  Does that pass/skip properly on XR?

jonnytdevops:
@chburket Nope, it doesn't. We're using the new style test for IOS XR now (which should also pass on Nexus). Thanks

jonnytdevops:
```
Finished in 386.256338s, 0.0518 runs/s, 0.1243 assertions/s.

20 runs, 48 assertions, 0 failures, 0 errors, 0 skips
```

jonnytdevops:
@chburket @chrisvanheuveln @mikewiebe Am I able to squash this for merge please? Thanks!

tests/test_tacacs_server_group.rb
```diff
@@ -231,11 +241,14 @@ def test_remove_server_twice_tacacs
...
 
     detach_aaaservergroup(aaa_group)
-    detach_tacacsserverhost(server1)
-    detach_tacacsserverhost(server2)
+    detach_tacacsserverhost(server_name1)
+    detach_tacacsserverhost(server_name2)
   end
 
   def test_get_vrf_tacacs
+    # Not part of NetDev development
```
mikewiebe: These are added because the netdev type does not support these properties correct?  If so, I would prefer one of the following options (in order of preference).

1) Add an `exclude: [ios_xr]` for these properties in the yaml and then use the `validate_property_excluded` to do the `assert_nil` and assert_raises(Cisco::UnsupportedError)` checks
- https://github.com/cisco/cisco-network-node-utils/blob/56982bca060bcab5a13dcb2124dce4a02ebe54d9/tests/test_vlan_private.rb#L76

2) Change one is simple, but based on our agreement we will accept the PR without this change, but please add a TODO reminder in the comments to validate this and we will make the updates.

jonnytdevops: @mikewiebe I'll add the excludes to the YAML file and return based on validate_property_excluded. Does this sound ok? Not sure if the negative testing makes sense for all non-netdev method as some properties have multiple methods. Thanks!

mikewiebe: @jonnytpuppet  That's fine for now.  We will add any needed negative test support.

jonnytdevops:
```
Finished in 397.913187s, 0.0503 runs/s, 0.1206 assertions/s.

20 runs, 48 assertions, 0 failures, 0 errors, 0 skips
```

Tests and YAML updated

@mikewiebe @chburket @chrisvanheuveln Should I squash this now for merge? Thanks!

chburket:
:+1: other than fixing the rubocop errors

jonnytdevops:
@chburket Rubocop updates pushed :)

mikewiebe:
:+1:  @jonnytpuppet go ahead and squash for merge.

jonnytdevops:
@mikewiebe @chburket PR squashed :) Thanks!

PR #331, opened 2016-04-04
Configure feature tacacs once in aaa test

robert-w-gries:
- Decreases test time by ~100 seconds.
- Test passes on all platforms.

### DIFF
--- a/tests/test_aaa_authorization_service.rb
+++ b/tests/test_aaa_authorization_service.rb
@@ -22,16 +22,17 @@ class TestAaaAuthorizationService < CiscoTestCase
 
   def setup
     super
-    cleanup_aaa if @@pre_clean_needed
-    @@pre_clean_needed = false # rubocop:disable Style/ClassVars
-    feature_tacacs
+    if @@pre_clean_needed
+      cleanup_aaa
+      config('feature tacacs+')
+      @@pre_clean_needed = false # rubocop:disable Style/ClassVars
+    end
     preconfig_tacacs_server_access(tacacs_groups[0])
     config_tacacs_servers(tacacs_groups[1..3])
   end
 
   def teardown
     cleanup_aaa
-    feature_tacacs(false)
     super
   end
 
@@ -46,11 +47,6 @@ class TestAaaAuthorizationService < CiscoTestCase
     servers.each do |server|
       config("aaa group server tacacs+ #{server}")
     end
-  end
-
-  def feature_tacacs(feature=true)
-    state = feature ? '' : 'no'
-    config("#{state} feature tacacs+")
   end
 
   # Helper method to get regexp for aaa authorization commands


tests/test_aaa_authorization_service.rb
```diff
@@ -22,16 +22,17 @@ class TestAaaAuthorizationService < CiscoTestCase
...
+      config('feature tacacs+')
+      @@pre_clean_needed = false # rubocop:disable Style/ClassVars
+    end
     preconfig_tacacs_server_access(tacacs_groups[0])
     config_tacacs_servers(tacacs_groups[1..3])
   end
 
   def teardown
     cleanup_aaa
-    feature_tacacs(false)
```
mikewiebe, line 34: Is the main rational for removing this teardown step to save runtime?

tests/test_aaa_authorization_service.rb
```diff
@@ -22,16 +22,17 @@ class TestAaaAuthorizationService < CiscoTestCase
 
   def setup
     super
-    cleanup_aaa if @@pre_clean_needed
-    @@pre_clean_needed = false # rubocop:disable Style/ClassVars
-    feature_tacacs
+    if @@pre_clean_needed
```
chrisvanheuveln, line 25: nit: `return if @@pre_clean_needed`

:+1:

tests/test_aaa_authorization_service.rb
robert-w-gries, line 34: Main rational is that configuring the feature to be on/off manually doesn't improve the test and wastes time.  Would rather set it once and have quicker tests since the test quality isn't affected by this change.

tests/test_aaa_authorization_service.rb
robert-w-gries, line 25: After looking at it more, it doesn't seem that the guard clause applies here.  The `setup` method would skip the actual setup if `@@pre_clean_needed` is set to `true`

chrisvanheuveln, line 25: sorry... `return unless @@pre_clean_needed`

tests/test_aaa_authorization_service.rb
mikewiebe, line 34: We should be restoring the testbed to a clean state.  Any features/config we apply for the test need to be removed when testing is complete so they don't impact tests that follow.

robert-w-gries, line 34: I thought our stance on feature disablement was that it had no impact on other features/tests so there was no purpose in implementing it (with a [caveat for vtp](https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/feature.rb#L200-L205)).  We've made changes to the feature provider to reflect this stance.

As far as I know and per previous discussions, `feature tacacs+` has no effect on other tests so making sure to remove the feature is unnecessary, especially since adding `no feature tacacs+` more than doubles the test time.

mikewiebe, line 34: Our stance is for the provider object itself but IMO, does not apply to test cases.  Any configuration needed by a test or set of tests should be removed at the end.  While there _may_ not be any perceptible impact now, this could certainly change in the future as if/when changes are made to the tacacs feature.  If you are worried about time, I came across the `after_tests()` method that might allow us to just cleanup at the end after all of the tests have completed.
- https://github.com/seattlerb/minitest/search?utf8=%E2%9C%93&q=after_tests

chrisvanheuveln, line 34: `after_run`  (`after_tests` is deprecated)

robert-w-gries, line 34: `Minitest.after_run` doesn't fit in nicely with our current test layout.  It's possible to add in our own method of pre-suite and post-suite cleanup but it's out of scope and will need to be investigated later.

PR #332, opened 2016-04-04
wait_for_initialize for slow-start bgp

chrisvanheuveln:
- 5/6k take a second or two to fully initialize; if our setters are called too quickly they'll fail with e.code 400 e.clierror {}
- Easy to repro in the sandbox (or cli) by doing:  'router bgp 1 ; router bgp 1 ; event cli size small' to simulate NU doing bgp create followed immediately by event_history_cli=
- Fixed by checking for the bgp process init state via 'show ip bgp'
- This is more noticeable with our 6k but the 5k has similar slow start problems so I kept 5k in the include list

### DIFF
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -102,14 +102,15 @@ module Cisco
     def wait_for_process_initialized
       return unless node.product_id[/N(5|6)/]
 
-      # Hack for slow-start platforms which may have setter failures if the
-      # bgp instance is still initializing when they process. To see this
-      # problem in a sandbox do 'router bgp 1 ; router bgp 1 ; shutdown'.
+      # Hack for slow-start platforms which will have setter failures if the
+      # bgp instance is still initializing. To see this problem in a sandbox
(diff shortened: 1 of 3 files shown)


lib/cisco_node_utils/bgp.rb
```diff
@@ -94,6 +95,23 @@ def destroy
...
+    end
+
+    def wait_for_process_initialized
+      return unless node.product_id[/N(5|6)/]
+
+      # Hack for slow-start platforms which may have setter failures if the
+      # bgp instance is still initializing when they process. To see this
+      # problem in a sandbox do 'router bgp 1 ; router bgp 1 ; shutdown'.
+      4.times do
+        break if process_initialized?
```
mikewiebe: I believe you mentioned that the error message is not very meaningful if the process is not initialized?  If that is indeed the case, I would like to see a meaningful message raised in the event that the process does not initialize after 4 attempts.

chrisvanheuveln: Added.

PR #341, opened 2016-04-08
MT-Lite support for 5k/6k

anramgop:
Added support for 5k,6k in the yaml files
Added support for N7700
Ran vxlan_vtep+vxlan_vtep_vni minitests on the 6k/7k to validate changes.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/feature.yaml
+++ b/lib/cisco_node_utils/cmd_ref/feature.yaml
@@ -33,13 +33,16 @@ itd:
   set_value: "feature itd"
 
 nv_overlay:
+  # N3k - native vxlan support in a future release 
+  # N6k - vxlan supported on 56xx only as of today
   _exclude: [N3k]
   kind: boolean
   get_value: '/^feature nv overlay$/'
   set_value: "feature nv overlay"
 
 nv_overlay_evpn:
-  # TBD: Exclude all n3k even though N31xx in 'switchmode n9k' supports evpn
+  # N3k - native vxlan support in a future release 
+  # N6k - vxlan supported on 56xx only as of today
   _exclude: [N3k]
   kind: boolean
   get_command: "show running | section nv"
@@ -71,8 +74,8 @@ vn_segment_vlan_based:
     get_value: '/^feature vn-segment-vlan-based$/'
     set_value: 'feature vn-segment-vlan-based'
     default_value: false
-  N8k: *vn_segment_vlan_based_mt_lite
   N9k: *vn_segment_vlan_based_mt_lite
+  N8k: *vn_segment_vlan_based_mt_lite
   N6k: *vn_segment_vlan_based_mt_lite
   N5k: *vn_segment_vlan_based_mt_lite
 
--- a/lib/cisco_node_utils/cmd_ref/feature.yaml
+++ b/lib/cisco_node_utils/cmd_ref/feature.yaml
@@ -33,7 +33,7 @@ itd:
   set_value: "feature itd"
 
 nv_overlay:
-  # N3k - native vxlan support in a future release 
+  # N3k - native vxlan support in a future release
   # N6k - vxlan supported on 56xx only as of today
   _exclude: [N3k]
   kind: boolean
@@ -41,7 +41,7 @@ nv_overlay:
   set_value: "feature nv overlay"
 
 nv_overlay_evpn:
-  # N3k - native vxlan support in a future release 
+  # N3k - native vxlan support in a future release
   # N6k - vxlan supported on 56xx only as of today
   _exclude: [N3k]
   kind: boolean
--- a/lib/cisco_node_utils/cmd_ref/feature.yaml
+++ b/lib/cisco_node_utils/cmd_ref/feature.yaml
@@ -34,7 +34,6 @@ itd:
 
 nv_overlay:
   # N3k - native vxlan support in a future release
-  # N6k - vxlan supported on 56xx only as of today
   _exclude: [N3k]
   kind: boolean
   get_value: '/^feature nv overlay$/'
--- a/lib/cisco_node_utils/cmd_ref/feature.yaml
+++ b/lib/cisco_node_utils/cmd_ref/feature.yaml
@@ -33,7 +33,7 @@ itd:
   set_value: "feature itd"
 
 nv_overlay:
-  _exclude: [N3k, N5k, N6k]
+  _exclude: [N3k]
   kind: boolean
   get_value: '/^feature nv overlay$/'
   set_value: "feature nv overlay"
@@ -73,6 +73,8 @@ vn_segment_vlan_based:
     default_value: false
   N8k: *vn_segment_vlan_based_mt_lite
   N9k: *vn_segment_vlan_based_mt_lite
+  N6k: *vn_segment_vlan_based_mt_lite
+  N5k: *vn_segment_vlan_based_mt_lite
 
 vni:
   kind: boolean
--- a/lib/cisco_node_utils/cmd_ref/feature.yaml
+++ b/lib/cisco_node_utils/cmd_ref/feature.yaml
@@ -37,11 +37,10 @@ nv_overlay:
   _exclude: [N3k]
   kind: boolean
   get_value: '/^feature nv overlay$/'
-  set_value: "feature nv overlay"
+  set_value: "<state> feature nv overlay"
 
 nv_overlay_evpn:
   # N3k - native vxlan support in a future release
-  # N6k - vxlan supported on 56xx only as of today
   _exclude: [N3k]
   kind: boolean
   get_command: "show running | section nv"
--- a/lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
+++ b/lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
@@ -29,14 +29,11 @@ mt_full_support:
 mt_lite_support:
   # This is only used for determining support for Multi-Tenancy Lite
   kind: boolean
-  N9k:
+  N9k: &mt_lite_default
     default_only: true
-  N8k:
-    default_only: true
-  N6k:
-    default_only: true
-  N5k:
-    default_only: true
+  N8k: *mt_lite_default
+  N6k: *mt_lite_default
+  N5k: *mt_lite_default
   else:
     # this feature is always off on these platforms and cannot be changed
     default_only: false
@@ -53,7 +50,7 @@ source_intf:
   default_value: ''
 
 source_intf_hold_down_time:
-  _exclude: [N7k]
+  _exclude: [N5k, N6k, N7k, N8k]
   kind: int
   get_value: '/^source-interface hold-down-time (\d+)$/'
   set_value: '<state> source-interface hold-down-time <time>'
--- a/lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
+++ b/lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
@@ -1,6 +1,6 @@
 # vxlan_vtep
 ---
-_exclude: [N3k, N5k, N6k, N7k, ios_xr]
+_exclude: [N3k, ios_xr]
 
 _template:
   get_command: "show running-config interface all | section 'interface nve'"
@@ -30,6 +30,12 @@ mt_lite_support:
   # This is only used for determining support for Multi-Tenancy Lite
   kind: boolean
   N9k:
+    default_only: true
+  N8k:
+    default_only: true
+  N6k:
+    default_only: true
+  N5k:
     default_only: true
   else:
     # this feature is always off on these platforms and cannot be changed
--- a/lib/cisco_node_utils/feature.rb
+++ b/lib/cisco_node_utils/feature.rb
@@ -112,6 +112,13 @@ module Cisco
       config_set('feature', 'nv_overlay')
     end
 
+    def self.nv_overlay_disable
+      # Note: vdc platforms restrict this feature to F3 or newer linecards
+      # Note: this is for test purposes only
+      return unless nv_overlay_enabled?
+      config_set('feature', 'nv_overlay', state: 'no')
+    end
+
     def self.nv_overlay_enabled?
       config_get('feature', 'nv_overlay')
     rescue Cisco::CliError => e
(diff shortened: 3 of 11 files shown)


lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
```diff
@@ -31,6 +31,12 @@ mt_lite_support:
   kind: boolean
   N9k:
     default_only: true
+  N8k:
+    default_only: true
+  N6k:
+    default_only: true
+  N5k:
+    default_only: true
```
robert-w-gries, line 33: Using an anchor will trim this down:

``` diff
-  N9k:
-     default_only: true
-  N8k:
-    default_only: true
-  N6k:
-    default_only: true
-  N5k:
-    default_only: true
+  N9k: &mt_lite_support_true
+    default_only: true
+  N8k: *mt_lite_support_true
+  N6k: *mt_lite_support_true
+  N5k: *mt_lite_support_true
```

tests/ciscotest.rb
```diff
@@ -236,7 +236,9 @@ def mt_full_interface?
     # we will provide an appropriate interface name if the linecard is present.
     # Example 'show mod' output to match against:
     #   '9  12  10/40 Gbps Ethernet Module  N7K-F312FQ-25 ok'
-    sh_mod = @device.cmd("sh mod | i '^[0-9]+.*N7K-F3'")[/^(\d+)\s.*N7K-F3/]
+    #   '9  12  10/40 Gbps Ethernet Module  N77-F312FQ-25 ok'
```
robert-w-gries, line 244: Unintentionally copied line?

robert-w-gries, line 244: Ah, never mind I see that `N77` is the difference

lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
```diff
@@ -1,6 +1,6 @@
 # vxlan_vtep
 ---
-_exclude: [N3k, N5k, N6k, N7k, ios_xr]
```
robert-w-gries, line 3: Since N7k is now an included platform, do you need to add an N7k entry to the `mt_lite_support` property?

chrisvanheuveln, line 3: @robert-w-gries fyi N7k's do mt-full, not mt-lite.

lib/cisco_node_utils/cmd_ref/feature.yaml
```diff
@@ -33,7 +33,7 @@ itd:
   set_value: "feature itd"
 
 nv_overlay:
-  _exclude: [N3k, N5k, N6k]
+  _exclude: [N3k]
```
chrisvanheuveln, line 37: What are the restrictions for this on n5k and n6k? i.e. This is going to be a pain to troubleshoot without some guidance so I'd like to see some comments in the yaml file; i.e. does this only work on specific models? does it require a specific OS version? 

Notes:
- this doesn't work at all on my 6k:

```
n6k-92(config)# feature nv ov
NVE Feature NOT supported on this Platform 
n6k-92(config)# sh ver | i system
  system:    version 7.3(0)N1(1)
```
- I have to disable fabricpath on my n5k to use this:

```
n56-136(config)# feature nv over
Please disable fabricpath before enabling 'nv overlay'
n56-136(config)# no feature-set fabricpath 
Feature-set Operation may take up to 95 minutes depending on the size of configuration.
n56-136(config)# feature nv over
n56-136(config)# 
```

chrisvanheuveln:
I suspect the minitests will fail for this on my n5k and n6k. Please run the following tests on my test hardware ( `n5k-136` and `n6k-92`) : 

```
ruby tests/test_feature.rb -n test_nv_overlay
ruby tests/test_feature.rb -n test_nv_overlay_evpn
ruby tests/test_overlay_global.rb
```

lib/cisco_node_utils/cmd_ref/feature.yaml
```diff
@@ -33,13 +33,16 @@ itd:
   set_value: "feature itd"
 
 nv_overlay:
-  _exclude: [N3k, N5k, N6k]
+  # N3k - native vxlan support in a future release
+  # N6k - vxlan supported on 56xx only as of today
+  _exclude: [N3k]
```
robert-w-gries, line 37: Does this mean n6k should be excluded as well?

tests/test_vxlan_vtep.rb
```diff
@@ -202,6 +202,9 @@ def test_source_interface
   end
 
   def test_source_interface_hold_down_time
+    skip('source_intf_hold_down_time not supported on this Platform') unless
+    node.cmd_ref.supports?('vxlan_vtep', 'source_intf_hold_down_time')
+
```
robert-w-gries: [`validate_property_excluded?` does the same check](https://github.com/cisco/cisco-network-node-utils/blob/develop/tests/ciscotest.rb#L142-L144). Please remove this skip and use the `validate_property_excluded` method instead

tests/test_vxlan_vtep_vni.rb
```diff
@@ -93,6 +93,8 @@ def test_vnis
   end
 
   def test_ingress_replication
+    skip('Platform does not support ingress_replication') unless
+    node.cmd_ref.supports?('vxlan_vtep_vni', 'ingress_replication')
```
robert-w-gries: Remove

tests/test_vxlan_vtep_vni.rb
```diff
@@ -153,6 +155,8 @@ def test_multicast_group
   end
 
   def test_peer_list
+    skip('Platform does not support peer-list') unless
+    node.cmd_ref.supports?('vxlan_vtep_vni', 'peer_list')
     vni = VxlanVtepVni.new('nve1', '6000')
     if validate_property_excluded?('vxlan_vtep_vni', 'ingress_replication')
```
robert-w-gries, line 174: I believe this is a copy paste error?  Argument should be `peer_list` instead of `ingress_replication`

tests/test_vxlan_vtep_vni.rb
```diff
@@ -210,4 +214,23 @@ def test_suppress_arp
     vni.suppress_arp = vni.default_suppress_arp
     refute(vni.suppress_arp, 'suppress_arp should be disabled')
   end
+
+  def test_suppress_uuc
+    skip('Platform does not support suppress-unknown-unicast') unless
+    node.cmd_ref.supports?('vxlan_vtep_vni', 'suppress_uuc')
```
robert-w-gries: Replace with:

```
if validate_property_excluded?('vxlan_vtep_vni', 'suppress_uuc')
  assert_nil(vni.suppress_uuc)
  assert_nil(vni.default_suppress_uuc)
  assert_raises(Cisco::UnsupportedError) { vni.suppress_uuc = true }
end
```

tests/test_vxlan_vtep_vni.rb
anramgop, line 174: Not a typo. peer-ip can be configured only when IR is configured.

tests/test_vxlan_vtep_vni.rb
anramgop: I tried this and I'm not seeing any unsupported errors on unsupported platforms. The skips work correctly on all platforms. Can we stick with my code till I identify the reason for this?

chrisvanheuveln: Maybe I misunderstand your comment but...  you shouldn't see "unsupported errors" if you coded it as Rob suggested. You should see three successful asserts. 
The asserts will only fail if you do _not_ get an UnsupportedError exception from a platform that _should_ be unsupported.

anramgop: Well - I'm not getting an "UnsupportedError exception" from a platform that should be unsupported.

lib/cisco_node_utils/vxlan_vtep_vni.rb
```diff
@@ -236,5 +236,27 @@ def suppress_arp=(state)
...
+
+    def suppress_uuc
+      config_get('vxlan_vtep_vni', 'suppress_uuc', @get_args)
+    end
+
+    def suppress_uuc=(state)
+      if state
+        set_args_keys(state: '')
+        # Host reachability must be enabled for this property
+        VxlanVtep.new(@name).host_reachability = 'evpn'
```
chrisvanheuveln: @mikewiebe @robert-w-gries I'm a little concerned about having this external dependency call; i.e. the fact that this property is going to flip-flop on each run when the user sets this to a different value using the vxlan_vtep provider. 
I suggest changing this to:

```
      raise RuntimeError, "Dependency: vxlan_vtep host_reachability must be 'evpn'" unless
          VxlanVtep.new(@name).host_reachability == 'evpn'
```

Thoughts?

mikewiebe: Agreed.  The manifest entry for `vxlan_vtep` should take care of this dependency.

tests/test_vxlan_vtep_vni.rb
chrisvanheuveln: You're not giving us a lot to go on here...

In any case, I pulled a copy of your fork and it works for me with these changes:  (i.e. I get 3 asserts and a pass for this testcase)

```
   def test_suppress_uuc
-    skip('Platform does not support suppress-unknown-unicast') unless
-    node.cmd_ref.supports?('vxlan_vtep_vni', 'suppress_uuc')
     vni = VxlanVtepVni.new('nve1', '6000')
+    if validate_property_excluded?('vxlan_vtep_vni', 'suppress_uuc')
+      assert_nil(vni.suppress_uuc)
+      assert_nil(vni.default_suppress_uuc)
+      assert_raises(Cisco::UnsupportedError) { vni.suppress_uuc = true }
+      return
+    end

     # Test: Check suppress_uuc is not configured.
```

I also made some changes to setup and teardown to make this run cleaner:

```
 class TestVxlanVtepVni < CiscoTestCase
   @skip_unless_supported = 'vxlan_vtep_vni'
+  @@pre_clean_needed = true # rubocop:disable Style/ClassVars

   def setup
     super
-    config('no feature nv overlay')
-    config('feature nv overlay')
-    config('interface nve1')
+    return unless @@pre_clean_needed
+
+    # Disable features on initial cleanup only as it's too expensive
+    # to do a disable for every testcase.
+    config_no_warn('no feature nv overlay')
+    config_no_warn('no feature vn-segment-vlan-based') if
+      VxlanVtep.mt_lite_support
+    config_no_warn('feature nv overlay')
+    # nv overlay is a slow start on some platforms
+    sleep 1
+    @@pre_clean_needed = false
   end

   def teardown
-    config('no feature nv overlay')
+    config_no_warn('no interface nve1')
     super
   end
```

I only ran `test_suppress_uuc` so you may have a little more work to do on the other test cases but this should get you close.

lib/cisco_node_utils/vxlan_vtep_vni.rb
anramgop: I just re-used the suppress-arp minitest. Will replace both of them.

chrisvanheuveln:
@anramgop Looks like you never responded to Rob's comment: https://github.com/cisco/cisco-network-node-utils/pull/341#discussion-diff-59757632R37

If n6k is supported then please remove that comment line; otherwise add an appropriate exclude.

👍  After you address that last issue.

lib/cisco_node_utils/feature.rb
```diff
@@ -112,6 +112,13 @@ def self.nv_overlay_enable
       config_set('feature', 'nv_overlay')
     end
 
+    def self.nv_overlay_disable
+      # Note: vdc platforms restrict this feature to F3 or newer linecards
+      # Note: this is for test purposes only
+      return unless nv_overlay_enabled?
+      config_set('feature', 'nv_overlay', state: 'no')
```
robert-w-gries, line 119: `nv_overlay` does not currently support the `state` keyword.

Add `<state>` [to the `set value` entry](https://github.com/cisco/cisco-network-node-utils/pull/341/files#diff-96c7c3f23af838d91015cff4d91c8501L39)

robert-w-gries:
My n7k fails every VxlanVtepVni test with this message:

```
  1) Error:
TestVxlanVtepVni#test_suppress_uuc:
Cisco::CliError: [#<Cisco::VxlanVtepVni:0x00000001b671b8>] The command 'interface nve1 ;  member vni 6000 ' was rejected with error:
Invalid interface format
```

tests/test_vxlan_vtep_vni.rb
```diff
@@ -192,6 +203,7 @@ def test_peer_list
 
   def test_suppress_arp
     vni = VxlanVtepVni.new('nve1', '6000')
+    VxlanVtep.new('nve1').host_reachability = 'evpn'
```
robert-w-gries, line 212: I suspect that this will cause failures on certain platforms that do not support `VxlanVtep`.  Can you post the test results of N3k, N5k, N6k, N7k (F2 card), and N9k?

anramgop:
Here is a quick recap of the changes I’m addressing based on your comments:
1. Included state for nv_overlay
2. A sleep after disabling nve as I’ve found that it is quite slow and erratic without this.
3. Included a vxlan_linecard? test under ciscotest to test for vxlan supported line cards. (Currently, the 6k line card supported is 6004 and 5k is quite a few including 5672. 7k and 77 support F3 card presently)

robert-w-gries:
@anramgop It doesn't look like you pushed your latest changes?  Also, you will need to merge the upstream `release_1.3.0` branch into your fork in order to resolve the merge conflicts.

tests/ciscotest.rb
```diff
@@ -244,4 +244,24 @@ def mt_full_interface?
...
+    # Example 'show mod' output to match against:
+    #   '9  12  10/40 Gbps Ethernet Module  N7K-F312FQ-25 ok'
+    #   '9  12  10/40 Gbps Ethernet Module  N77-F312FQ-25 ok'
+    #   '2   6  Nexus 6xQSFP Ethernet Module  N5K-C5672UP-M6Q ok'
+    #   '2   6  Nexus xxQSFP Ethernet Module  N6K-C6004-96Q/EF ok'
+    if node.product_id[/N(5|6)K/]
+      sh_mod_string = @device.cmd("sh mod | i '^[0-9]+.*N[56]K-C[56]'")
+      sh_mod = sh_mod_string[/^(\d+)\s.*N[56]K-C(56|6004)/]
+      slot = sh_mod.nil? ? nil : Regexp.last_match[1]
+      skip('Unable to find compatible interface in chassis') if slot.nil?
```
robert-w-gries: Nit: unnecessary line

``` diff
-      slot = sh_mod.nil? ? nil : Regexp.last_match[1]
-      skip('Unable to find compatible interface in chassis') if slot.nil?
+      skip('Unable to find compatible interface in chassis') if sh_mod.nil?
```

PR #342, opened 2016-04-08
pvlan fixes to support manifest

davidefdl:


lib/cisco_node_utils/vlan.rb
```diff
@@ -100,19 +100,19 @@ def fabricpath_feature_set(fabricpath_set)
...
       else
-        if 'fabricpath' == str
+        if 'fabricpath' == str[0]
           fabricpath_feature_set(:enabled) unless
             :enabled == fabricpath_feature
         end
-        config_set('vlan', 'mode', @vlan_id, '', str)
+        config_set('vlan', 'mode', @vlan_id, '', str[0])
       end
     end
```
chrisvanheuveln, line 121: I'm confused about why `str` is now an array? Can you explain? 

The getter returns a string and the minitest that tests this setter passes a string (`v.mode = 'fabricpath'`) . This seems wrong and I'm guessing the minitest won't pass now. Have you run the minitests that validate these APIs ?

chrisvanheuveln:
In the future:
- Please include a description with your code reviews. It's not terribly clear (from the review itself) why you're making these changes. If I look back at this commit 2 months from now will it be obvious what this change was intended to fix?
- If you change an API you must run the appropriate minitests to validate it. In the review description include a statement to the effect that you '_ran test foo on platform bar_' so that we know you did the basic due diligence.
- Read this if you haven't already: https://github.com/cisco/cisco-network-node-utils/blob/develop/docs/README-develop-best-practices.md

lib/cisco_node_utils/vlan.rb
davidefdl, line 121: Hi. Yes you are right. The setter input should be a string. Changed the code to reflect this.
I missed that there was a minitest for vlan_mt_full. I tried this morning but I do not have a 7k with
a F3 LC. That seems to be the only LC where this tests can be run.
The fix is temporary. There is a nxapi bug when a vlan is configured as private vlan.
In the returned body of show vlan, the output is displayed twice causing the rejection of the manifest, since vlan.mode expects just one value.

davidefdl:
This code change is a temporary fix to address a nxapi bug. Without this, private vlan type and provider can not proceed. Once the nxapi bug is fixed we can revert back to the original code. @chrisvanheuveln

lib/cisco_node_utils/vlan.rb
chrisvanheuveln, line 121: Okay, so let's just add a comment to the getter to explain why this is odd; i.e.:

```
     def mode
        result = config_get('vlan', 'mode', @vlan_id)
        return default_mode if result.nil?
+        # Note: The yaml definition for this property uses 'multiple' as a workaround for a bug
+        # in the N7k nxapi code which displays the 'show vlan' output twice.
+       result[0].downcase! if result[0][/FABRICPATH/]
+       result[0]
      end
```

Also, I ran the vlan_mt_full minitest on my n7k (which has an F3 card) and it works, so you're good there.

davidefdl, line 121: Added the comment for the temp bug fix

PR #344, opened 2016-04-08
Feature/vlan fabric control

rkorlepa:
New property fabric-control for vlan used in MT-FULL fabricpath cases.


tests/test_vlan.rb
```diff
@@ -316,4 +316,29 @@ def test_mapped_vnis
...
+    assert(vlan.fabric_control)
+    another_vlan = Vlan.new(101)
+
+    assert_raises(RuntimeError,
+                  'VLAN misconfig did not raise CliError') do
+      another_vlan.fabric_control = true
+    end
+    vlan.destroy
+    another_vlan.destroy
+  end
```
chrisvanheuveln, line 342: 1. Your tests are okay but the second tests does the same as the first one so I would just remove the first one
2. fabric-control is not supported on some platforms so you need to explicit test for the exclusions

e.g.:

```
-  def test_vlan_fabric_control
-  def test_another_vlan_as_fabric_control

+  def test_fabric_control
    if validate_property_excluded?('vlan', 'fabric_control')
      assert_raises(Cisco::UnsupportedError) { Vlan.new('100').fabric_control = true }
      return
    end

    v100 = Vlan.new('100')
    assert_equal(v100.default_fabric_control, v100.fabric_control)
    v100.fabric_control = true
    assert(v100.fabric_control)

    v101 = Vlan.new(101)
    assert_raises(RuntimeError,
                  'VLAN misconfig did not raise CliError') do
      v101.fabric_control = true
    end
    v100.destroy
    v101.destroy
  end
```

rkorlepa, line 342: Remove the redundant test case for fabric_control and added the explicit exclude for the test cases.
Also tested this on all the platforms for skip on n3k, n5k, n6k, n9k, n8k and pass/run on n7k

chrisvanheuveln:
:+1: Please update your fork and resolve the merge conflicts. I'll merge after that. Thanks

lib/cisco_node_utils/cmd_ref/tacacs_server_group.yaml
```diff
@@ -1,9 +1,13 @@
 # tacacs_server_group
 ---
-_exclude: [ios_xr]
+_template:
```
mikewiebe: @rkorlepa Looks like you merged latest `develop` branch into your workspace instead of latest `release_1.3.0`. Please revert this merge and merge latest release instead.

rkorlepa: Fixed it. Can you please check and merge the changes.

PR #349, opened 2016-04-13
Skip mode test for 9k and 3k platform

davidefdl:
One of the minitest for test_vlan.rb is failing for 9k and 3k. Reason is the mode property is excluded
from these two platform. The mode is supported in both platform but since the mode property was introduced for fabricpath ( not supported in 9k and 3k) the property was excluded from 9k and 3k.
To avoid the failure, I'm proposing for now to return when the platform is either 3k or 9k.

### DIFF
--- a/tests/test_vlan.rb
+++ b/tests/test_vlan.rb
@@ -318,6 +318,7 @@ class TestVlan < CiscoTestCase
   end
 
   def test_mode_with_pvlan
+    return if node.product_id[/N9K|N3K/]
     v = Vlan.new(1000)
     result = 'CE'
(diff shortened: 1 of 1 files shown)


chrisvanheuveln:
Mike already beat you to this fix yesterday. See: https://github.com/cisco/cisco-network-node-utils/commit/90dfd31c975a8c373f7834c26aa623ab38f98739#diff-a870d284ff290dbf87d6416cf7ddb7a8R321

Also note that the test should use `validate_property_excluded?()` instead of checking explicit platforms.

PR #351, opened 2016-04-14
Add nil check to interface test

robert-w-gries:
Add nil check to prevent stack trace when `capabilities` on switch cannot be found

### DIFF
--- a/tests/test_interface.rb
+++ b/tests/test_interface.rb
@@ -576,7 +576,7 @@ class TestInterface < CiscoTestCase
 
     # Test up to two non-default values
     speed_values = capable_speed_values(interface)
-    warn('No compatible speeds found on interface') if speed_values.nil?
+    warn("No valid speeds found on #{interface.name}") if speed_values.empty?
     successful_runs = 0
     speed_values.each do |value|
       break if successful_runs >= 2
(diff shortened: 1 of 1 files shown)


tests/test_interface.rb
```diff
@@ -141,15 +141,15 @@ def interface_count
   def capable_speed_values(interface)
     capabilities = config("show interface #{interface.name} capabilities")
     speed_capa = capabilities.match(/Speed:\s+(\S+)/)
-    return [] if speed_capa[1].nil?
+    return [] if speed_capa.nil? || speed_capa[1].nil?
```
mikewiebe, line 144: I think it might be helpful to log a warning when speed and duplex capabilities cannot be detected.

robert-w-gries, line 144: Done

tests/test_interface.rb
```diff
@@ -576,6 +576,7 @@ def test_speed
 
     # Test up to two non-default values
     speed_values = capable_speed_values(interface)
+    warn('No compatible speeds found on interface') if speed_values.nil?
```
mikewiebe: **Nit** Include the interface name.. otherwise :+1:

robert-w-gries: Done.  I also corrected the method to be `empty?` instead of `nil?`

PR #356, opened 2016-04-18
Add Interface.capabilities helper

chrisvanheuveln:
- Minitest and beaker need an easy way to determine interface capabilities for properties like speed, duplex, etc
- show interface capabilities supplies this info for NX platforms
- `capable_speed_values` and `capable_duplex_values` already does this in minitest using config() so I just took this a step further and created a class method getter in Interface that converts the output into hashed values; it also has an option to return the raw output instead since some of the lesser-used values are multi-line and don't convert easily to key-values.
- Minitest tested with n3,5,6,7,9k. I intend to use this in beaker next.

### DIFF
--- a/lib/cisco_node_utils/interface.rb
+++ b/lib/cisco_node_utils/interface.rb
@@ -95,6 +95,38 @@ module Cisco
       true
     end
 
+    # 'capabilities' is a getter-only helper for minitest and beaker.
+    # mode values:
+    #   :hash = Transform the output into a hash
+    #   :raw  = The raw output from 'show int capabilities'. Some multi-line
+    #           values do not translate easily so this option allows the
+    #           caller to extract the data it needs.
+    #
+    # Sample cli output:
+    #    Model:                 N7K-M132XP-12L
+    #    Type (SFP capable):    10Gbase-SR
+    #    Speed:                 10,100,1000
+    #
+    # Sample hash output:
+    # {"Model"=>"N7K-M132XP-12L", "Type"=>"10Gbase-SR", "Speed"=>"10,100,1000"}
+    #
+    def self.capabilities(intf, mode=:hash)
+      array = config_get('interface', 'capabilities', name: intf)
+      return array if mode == :raw
+      hash = {}
+      if array
+        array.delete('')
+        array.each do |line|
+          k, v = line.split(':')
+          next if k.nil? || v.nil?
+          k.gsub!(/ \(.*\)/, '')
+          v.gsub!(/\s/, '')
+          hash[k] = v
+        end
+      end
+      hash
+    end
+
     def create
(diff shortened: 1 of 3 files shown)


tests/test_interface.rb
```diff
@@ -1344,6 +1351,7 @@ def test_interface_ipv4_all_interfaces
 
     # pre-configure
     begin
+      interface_ethernet_default(interfaces[1])
```
robert-w-gries, line 1361: Why is this change necessary/relevant?

lib/cisco_node_utils/interface.rb
```diff
@@ -95,6 +95,38 @@ def self.filter(opt, id)
...
+      array = config_get('interface', 'capabilities', name: intf)
+      return array if mode == :raw
+      hash = {}
+      if array
+        array.delete('')
+        array.each do |line|
+          k, v = line.split(':')
+          next if k.nil? || v.nil?
+          k.gsub!(/ \(.*\)/, '')
+          v.gsub!(/\s/, '')
```
robert-w-gries: What is the intent of these substitutions?  If you only want to remove leading/trailing whitespace, use `.strip`

chrisvanheuveln: `k` removes parenthesized text from the key: `"Type"` vs `"Type (SFP capable)"` vs `"Type (Non-SFP capable)"`.

`v` was intended to remove all whitespace, not just leading/trailing. But on second check I don't see any of those cases so I'll change it.

tests/test_interface.rb
```diff
@@ -137,20 +137,27 @@ def interface_count
     arr.count
   end
 
+  def test_interface_capabilities
+    if validate_property_excluded?('interface', 'capabilities')
+      assert_empty(Interface.capabilities(interfaces[0]))
+    else
+      refute_nil(Interface.capabilities(interfaces[0]),
+                 'Interface.capabilities should not return an empty hash')
```
robert-w-gries: Please add asserts for getting the `Model` value from both `raw` and `hash` modes.

tests/test_interface.rb
chrisvanheuveln, line 1361: - Line 1355 is a pre-config dependency for several tests that follow.
- The pre-config failed when I ran the full test because some previous, unrelated test left some vni configs on that interface that interfered with setting channel-group. 
- Notice that it's a different interface than the one that's used for all other tests in this massive test file.

Since this is the only place it's used it's better to just do it here than in `setup`, which would slow down every test case.

robert-w-gries, line 1361: Ack

tests/test_interface.rb
chrisvanheuveln: done

robert-w-gries:
👍

PR #358, opened 2016-04-20
netdev tacacs_server_host fixes

mikewiebe:
**Summary of changes:**
- Added tacacs to `feature.rb`
- Fixed bug in `tacacs_server_host.yaml` which referenced the wrong key.
- `port` attribute incorrectly returned an array of ports instead of a single port.

Tests Pass On: n3k, n5k, n6k, n7k, n8k, n9k(I2 and I3)

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/feature.yaml
+++ b/lib/cisco_node_utils/cmd_ref/feature.yaml
@@ -63,6 +63,12 @@ private_vlan:
   get_value: '/^feature private-vlan$/'
   set_value: "feature private-vlan"
 
+tacacs:
+  kind: boolean
+  get_command: "show running | i ^feature"
+  get_value: '/^feature tacacs\+$/'
+  set_value: "feature tacacs"
+
 vn_segment_vlan_based:
   # MT-lite only
   N3k: &vn_segment_vlan_based_mt_lite
--- a/lib/cisco_node_utils/cmd_ref/feature.yaml
+++ b/lib/cisco_node_utils/cmd_ref/feature.yaml
@@ -59,13 +59,11 @@ pim:
 private_vlan:
   _exclude: [N8k]
   kind: boolean
-  get_command: "show running | i ^feature"
   get_value: '/^feature private-vlan$/'
   set_value: "feature private-vlan"
 
 tacacs:
   kind: boolean
-  get_command: "show running | i ^feature"
   get_value: '/^feature tacacs\+$/'
   set_value: "feature tacacs"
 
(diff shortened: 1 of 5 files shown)


lib/cisco_node_utils/cmd_ref/feature.yaml
```diff
@@ -63,6 +63,12 @@ private_vlan:
   get_value: '/^feature private-vlan$/'
   set_value: "feature private-vlan"
 
+tacacs:
+  kind: boolean
+  get_command: "show running | i ^feature"
+  get_value: '/^feature tacacs\+$/'
+  set_value: "feature tacacs"
```
chrisvanheuveln, line 68: Thanks for moving this into Feature. With luck we'll get the rest moved by 1.4.0.

lib/cisco_node_utils/cmd_ref/feature.yaml
```diff
@@ -63,6 +63,12 @@ private_vlan:
   get_value: '/^feature private-vlan$/'
   set_value: "feature private-vlan"
 
+tacacs:
+  kind: boolean
+  get_command: "show running | i ^feature"
```
chrisvanheuveln: `get_command:` is inherited from the template so you don't need this.

lib/cisco_node_utils/feature.rb
```diff
@@ -169,6 +169,16 @@ def self.private_vlan_enabled?
     end
 
     # ---------------------------
+    def self.tacacs_enable
+      return if tacacs_enabled?
```
chrisvanheuveln: What do you think of this?

```
-    return if tacacs_enabled?
+    return if tacacs_enabled? || platform == :ios_xr
```

...what you have is okay, I'm just wondering if the platform checks would be better in the feature enable methods themselves instead of sprinkling plat checks in each of the callers. 

Maybe leave a TBD comment if you don't want to address it now.

lib/cisco_node_utils/tacacs_server_host.rb
```diff
@@ -70,6 +71,7 @@ def self.hosts
 
     def create
       destroy if platform == :ios_xr
+      Feature.tacacs_enable unless platform == :ios_xr
```
chrisvanheuveln: See plat check comment in feature.rb...

lib/cisco_node_utils/cmd_ref/feature.yaml
mikewiebe: True.. forgot that was changed.

lib/cisco_node_utils/feature.rb
mikewiebe: Yup. Like it.

PR #361, opened 2016-04-22
Fix N3k minitest and fir prepare array api to incorporate comma separ…

davidefdl:
Fix N3k minitest, excluding pvlan trunk properties in the yaml file since they are not supported.
In test_interface_private_vlan.rb make sure to default the interfaces before removing the vlans since N3k does not allowed this.
In interface.rb fix the prepare_array utility, since the comma separated vlan were not included.

Run test against N3k(l2), N9k, N6k, N5k and N7k

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/interface.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface.yaml
@@ -421,28 +421,29 @@ switchport_mode_private_vlan_host_promiscous:
 
 
 switchport_mode_private_vlan_trunk_promiscuous:
-  _exclude: [ios_xr, N8k]
+  _exclude: [ios_xr, N3k, N8k]
   kind: boolean
   get_value: '/^switchport mode private-vlan trunk promiscuous$/'
   set_value: "<state> switchport mode private-vlan trunk promiscuous"
   default_value: false
 
 switchport_mode_private_vlan_trunk_secondary:
-  _exclude: [ios_xr, N8k]
+  _exclude: [ios_xr, N3k, N8k]
   kind: boolean
   get_value: '/^switchport mode private-vlan trunk secondary$/'
   set_value: "<state> switchport mode private-vlan trunk secondary"
   default_value: false
 
 switchport_private_vlan_association_trunk:
-  _exclude: [ios_xr, N8k]
+  _exclude: [ios_xr, N3k, N8k]
   multiple: true
+  #get_value: '/^switchport private-vlan association trunk (.*) (.*)$/'
   get_value: '/^switchport private-vlan association trunk (.*)$/'
   set_value: "<state> switchport private-vlan association trunk <vlan_pr> <vlan>"
   default_value: []
 
 switchport_private_vlan_mapping_trunk:
-  _exclude: [ios_xr, N8k]
+  _exclude: [ios_xr, N3k, N8k]
   multiple: true
   get_value: '/^switchport private-vlan mapping trunk (.*)$/'
   set_value: "<state> switchport private-vlan mapping trunk <vlan_pr> <vlans>"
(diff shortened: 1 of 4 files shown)


lib/cisco_node_utils/interface.rb
```diff
@@ -1019,9 +1019,19 @@ def default_switchport_mode_private_vlan_host_association
     end
 
     def prepare_array(is_list)
```
chrisvanheuveln, line 1030: It would be helpful to include some comments for this increasingly complicated method; i.e. some sample inputs and the expected result format. Also note for future methods (I'm not going to make you change it here), but please try to choose better method and variable names. `prepare_array` is fairly vague when we use arrays all over the place; `is_list` isn't really the 'is' list when you use this method to format the `should_list`; etc. Try to code for the people that will have to debug this or extend the functionality down the road.

lib/cisco_node_utils/interface.rb
```diff
@@ -1074,6 +1084,7 @@ def configure_private_vlan_trunk_property(property, should_list_new,
         is_list.each do |vlans|
           vlans = vlans.split(' ')
           if vlans[0].eql? should_list_new[0]
+            # if vlans[0].eql? pr_vlan
```
chrisvanheuveln: I would guess that this commented code should be removed?

lib/cisco_node_utils/interface.rb
```diff
@@ -1085,6 +1096,7 @@ def configure_private_vlan_trunk_property(property, should_list_new,
         end
         result = config_set('interface', PVLAN_PROPERTY[property], name: @name,
                             state: '', vlan_pr: should_list_new[0],
+                            # state: '', vlan_pr: pr_vlan,
```
chrisvanheuveln: ditto remove

chrisvanheuveln:
Please also fix test_interface_svi.rb. I partially fixed it with this:

```
+++ b/tests/test_interface_svi.rb
@@ -107,7 +107,7 @@ class TestSvi < CiscoTestCase
       return
     end
     input = %w(10 20)
-    result = %w(10 20)
+    result = %w(10,20)
```

...but it fails after that with:

```
TestSvi#test_private_vlan_mapping_bad_args [tests/test_interface_svi.rb:118]:
svi pvlan mapping did not raise RuntimeError.
[RuntimeError] exception expected, not
Class: <Cisco::CliError>
Message: <"[interface vlan23] The command ' private-vlan mapping 23' was rejected with error:\nERROR: secondary VLAN list contains primary VLAN id\n">

```

lib/cisco_node_utils/cmd_ref/interface.yaml
```diff
@@ -421,28 +421,29 @@ switchport_mode_private_vlan_host_promiscous:
 
 
 switchport_mode_private_vlan_trunk_promiscuous:
-  _exclude: [ios_xr, N8k]
+  _exclude: [ios_xr, N3k, N8k]
```
mikewiebe, line 424: Does I4 (Dublin+ support this? Why does N9k support this, but not N3k?  They are running the same image.

lib/cisco_node_utils/interface.rb
davidefdl, line 1030: Absolutely agree with you. I did not have much time so far to revisit this api and other to add the proper comment. I've added a comment now at the top the api but it will need more clean up and explanation in the feature.

lib/cisco_node_utils/interface.rb
davidefdl: done

lib/cisco_node_utils/interface.rb
davidefdl: done

lib/cisco_node_utils/cmd_ref/interface.yaml
davidefdl, line 424: Not sure why they do not support trunk mode. In the cmd file
keyword trunk "private-vlan trunk " dynamic $notvalidinn3k they specifically skip this mode

chrisvanheuveln:
👍  _IF_ the tests are now passing.

PR #367, opened 2016-04-24
Release 1.3.0

dcheriancisco:
Added 'compatible_interfaces' class method which list the compatible interfaces on a per feature basis for modular systems. This infra will be made available via 'facter' to be accessed by puppet layer.

In each feature where it is required, a property 'supported_module_pids' should be made available in its YAML file which lists the regex for the supported modules.

For example with 'fabricpath' feature on N7K:

supported_module_pids:
  _exclude: [N5k, N6k]
  default_only: 'N7K-(?:F2.*25E|F3|F4|M3)'

specifies the supported modules as F2E, F3, F4 and M3

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/fabricpath.yaml
+++ b/lib/cisco_node_utils/cmd_ref/fabricpath.yaml
@@ -144,6 +144,10 @@ mode:
   set_value: "<state> fabricpath mode transit"
   default_value: "normal"
 
+supported_module_pids:
+  _exclude: [N5k, N6k]
+  default_only: 'N7K-(?:F2.*25E|F3|F4|M3)'
+
 supported_modules:
   _exclude: [N5k, N6k]
   default_only: "f2e f3"
--- a/lib/cisco_node_utils/feature.rb
+++ b/lib/cisco_node_utils/feature.rb
@@ -237,16 +237,17 @@ module Cisco
         result.is_a?(String) &&
         /Hardware is not capable of supporting/.match(result)
     end
+
     # ---------------------------
     def self.compatible_interfaces(feature)
       # Figure out the interfaces in a modular switch that are
-      # compatible with the given feature and return an array of 
+      # compatible with the given feature and return an array of
       # such interfaces
       module_regex = Regexp.new config_get(feature, 'supported_module_pids')
       return [] if module_regex.nil?
       # first get the compatible modules present in the switch
-      slots = Platform.slots.select do |slot, filt_mod| 
-        filt_mod['pid']  =~ module_regex
+      slots = Platform.slots.select do |_slot, filt_mod|
+        filt_mod['pid'] =~ module_regex
       end
       return [] if slots.empty?
       # get the slot numbers only into filtered slots array
@@ -254,7 +255,7 @@ module Cisco
       # now filter interfaces in the vdc based on compatible slots
       vdc = Vdc.new(Vdc.default_vdc_name)
       filt_intfs = vdc.interface_membership.select do |intf|
-          filt_slots.include? intf[/\d+/]
+        filt_slots.include? intf[/\d+/]
       end
       filt_intfs
     end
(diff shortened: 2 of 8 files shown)


chrisvanheuveln:
This code is a nice innovation. The next step will be to add support on the beaker side, plus refactor the mt-full tests to use it. 

Nit: Please use a meaningful pull request title.

lib/cisco_node_utils/feature.rb
```diff
@@ -237,5 +237,27 @@ def self.cli_error_check(result)
...
+      return [] if slots.empty?
+      # get the slot numbers only into filtered slots array
+      filt_slots = slots.keys.map { |key| key[/\d+/] }
+      # now filter interfaces in the vdc based on compatible slots
+      vdc = Vdc.new(Vdc.default_vdc_name)
+      filt_intfs = vdc.interface_membership.select do |intf|
+        filt_slots.include? intf[/\d+/]
+      end
+      filt_intfs
+    end
```
chrisvanheuveln, line 262: Why did you put this in the Feature class? Seems like this is very VDC specific and would be a better fit in the Vdc class.

tests/ciscotest.rb
```diff
@@ -225,6 +225,28 @@ def interface_cleanup(intf_name)
...
   end
 
+  # setup fabricpath env if possible and populate the interfaces array
+  # otherwise cause a global skip
+  def fabricpath_testenv_setup
+    return unless node.product_id[/N7K/]
+    intf_array = Feature.compatible_interfaces('fabricpath')
+    vdc = Vdc.new(Vdc.default_vdc_name)
+    save_lr = vdc.limit_resource_module_type
+    if intf_array.empty? || save_lr != 'f2e f3'
```
chrisvanheuveln: Instead of hard-coding 'f2e f3' can you use the yaml via `FabricPath.supported_modules` ? (would require adding `self.supported_modules` to fabricpath)

lib/cisco_node_utils/cmd_ref/fabricpath.yaml
```diff
@@ -144,6 +144,10 @@ mode:
   set_value: "<state> fabricpath mode transit"
   default_value: "normal"
 
+supported_module_pids:
+  _exclude: [N5k, N6k]
+  default_only: 'N7K-(?:F2.*25E|F3|F4|M3)'
+
 supported_modules:
   _exclude: [N5k, N6k]
   default_only: "f2e f3"
```
chrisvanheuveln, line 153: Can these two tags be combined? e.g.

```
supported_modules:
  _exclude: [N5k, N6k]
  config_set: "f2e f3 f4 m3"
  default_only: 'N7K-(?:F2.*25E|F3|F4|M3)'
```

edit: I seem to recall that some vxlan features require 'f3' exclusively and fail if you do 'f2e f3' so maybe this isn't feasible.

lib/cisco_node_utils/feature.rb
dcheriancisco, line 262: I wanted to use this in the facter. There I have used the name 'feature_compatible_module_iflist' which is indexed by a feature name like 'fabricpath' or 'vpc' or 'vni' etc. Please check the puppet layer PR as well

tests/ciscotest.rb
dcheriancisco: yes good point. I was meaning to use that. Will change it.

lib/cisco_node_utils/cmd_ref/fabricpath.yaml
dcheriancisco, line 153: right let's keep it separate for now.

lib/cisco_node_utils/feature.rb
chrisvanheuveln, line 262: Got it. That's fine then.

chrisvanheuveln:
👍  After you fix the hard-coded 'f2e f3' references.

mikewiebe:
:+1:

dcheriancisco:
NOTE: I haven't added the method supported_modules in fabricpath as it is only relevant for n7k. Instead calling the node.config_get()

PR #368, opened 2016-04-25
Treat speed as string

robert-w-gries:
Fix needed to fix puppet idempotence issue:

```
Notice: /Stage[main]/Main/Node[default]/Cisco_interface[ethernet3/10]/speed: speed changed '10000' to '10000'
```


chrisvanheuveln:
Does `test_interface` minitest pass after this change?

robert-w-gries:
Platforms tested with success:
- n3k
- n5k
- n6k
- n7k
- n9k - I3

PR #386, opened 2016-04-29
For private vlan interface properties, the return value when feature …

davidefdl:
For private vlan interface properties the return value should be an empty array when the feature is not enabled since the show run interface output is available even if feature private-vlan is not configured.
Change the minitest accordingly .
Ran test towards 3k/5-6k/7k/9k

### DIFF
--- a/tests/test_interface_private_vlan.rb
+++ b/tests/test_interface_private_vlan.rb
@@ -61,8 +61,6 @@ class TestSwitchport < TestInterfaceSwitchport
   def test_interface_switchport_private_host_mode
     if validate_property_excluded?('interface',
                                    'switchport_mode_private_vlan_host')
-      assert_equal([], interface.switchport_mode_private_vlan_host,
-                   'Err: should not be supported ')
       assert_raises(Cisco::UnsupportedError) do
         interface.switchport_mode_private_vlan_host = :host
       end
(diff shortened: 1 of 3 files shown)


tests/test_interface_private_vlan.rb
```diff
@@ -61,7 +61,8 @@ class TestSwitchport < TestInterfaceSwitchport
   def test_interface_switchport_private_host_mode
     if validate_property_excluded?('interface',
                                    'switchport_mode_private_vlan_host')
-      assert_nil(interface.switchport_mode_private_vlan_host)
+      assert_equal([], interface.switchport_mode_private_vlan_host,
+                   'Err: should not be supported ')
```
chrisvanheuveln: These asserts don't make sense. The getters are going to return [] even if the properties are supported. 
Just remove them altogether -- the `assert_raise` checks should be sufficient for these properties.

davidefdl: Done

tests/test_interface_svi.rb
```diff
@@ -101,7 +103,9 @@ def test_private_vlan_mapping
   def test_private_vlan_mapping_bad_args
     if validate_property_excluded?('interface',
                                    'private_vlan_mapping')
-      assert_nil(svi.private_vlan_mapping)
+      assert_equal([], svi.private_vlan_mapping,
+                   'Err: should not be supported ')
```
chrisvanheuveln: missed this one

davidefdl: Done

mikewiebe:
@davidefdl Will patch my changes in that we discussed and open the new PR shortly.

PR #388, opened 2016-04-29
Refactored private vlan getters

mikewiebe:
**Summary:**
Refactored getters in `interface.rb` to use default yaml entries as single source of truth.

`tests/test_interface_private_vlan.rb` and `test_interface_svi.rb` tests pass on:
 `n3k(I2, I3), n5k, n6k, n7k, n8k, n9k(I2, I3)`


mikewiebe:
@davidefdl Here is the refactored code.

PR #389, opened 2016-04-30
Feature/vxlan fixes

robert-w-gries:
### Changes
- Explicitly include n5k and n6k in yaml
- Add delay after `nv_overlay_enable` to prevent crash on N5k
- Add check for vxlan linecard
### Testing
- Validated against vxlan-supported platforms: N5k and N7k
- Validated against non-supported platforms (skipped)


chrisvanheuveln:
👍

PR #385, opened 2016-04-29
N7K F3 errors

anramgop:
The following minitests are fixed using the function - vdc_lc_state.
- TestOverlayGlobal##test_anycast_gateway_mac
- TestOverlayGlobal##test_dup_host_ip_addr_detection
- TestFeature##test_nv_overlay
- TestVrfAf#test_route_target
- TestVrf#test_route_distinguisher
  Here are the minitest runs:

```
###ruby tests/test_vrf_af.rb -e n7k -v -n test_route_target
Run options: -e n7k -v -n test_route_target --seed 52231

Running:


Node under test:
  - name  - core
  - type  - N7F-C7018
  - image - bootflash:///aneesh-s.bin

TestVrfAf#test_route_target = 81.66 s = .

Finished in 81.665799s, 0.0122 runs/s, 0.9306 assertions/s.

1 runs, 76 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for MiniTest to /nobackup/anramgop/puppet/cisco-network-node-utils/coverage. 1339 / 2725 LOC (49.14%) covered.
ruby tests/test_vrf_af.rb -e lab7k -v -n test_route_target
Run options: -e lab7k -v -n test_route_target --seed 62205

Running:


Node under test:
  - name  - n7k-j
  - type  - N7K-C7009
  - image - bootflash:///n7000-s2-dk9.7.3.1.D1.0.48.bin

TestVrfAf#test_route_target = 8.85 s = S

Finished in 8.851257s, 0.1130 runs/s, 0.2260 assertions/s.

  1) Skipped:
TestVrfAf#test_route_target [/nobackup/anramgop/puppet/cisco-network-node-utils/tests/ciscotest.rb:279]:
Unable to find compatible interface in chassis

1 runs, 2 assertions, 0 failures, 0 errors, 1 skips
Coverage report generated for MiniTest to /nobackup/anramgop/puppet/cisco-network-node-utils/coverage. 1220 / 2725 LOC (44.77%) covered.
###ruby tests/test_vrf.rb -e n7k -v -n test_route_distinguisher
Run options: -e n7k -v -n test_route_distinguisher --seed 49327

Running:


Node under test:
  - name  - core
  - type  - N7F-C7018
  - image - bootflash:///aneesh-s.bin

TestVrf#test_route_distinguisher = 11.97 s = .

Finished in 11.973159s, 0.0835 runs/s, 0.4176 assertions/s.

[...]

PR #391, opened 2016-05-02
vxlan/f3 test dependencies

chrisvanheuveln:
- Committing on behalf on #385, which had merge conflicts and we wanted this fix prior to releasing 1.3.0
- One issue with #385 was the route_target tests in vrf_af: Aneesh had placed a check for vxlan_linecard there, which solves the problem with platforms that don't have the requisite linecards to do evpn; but that call will also doa skip which in this case would skip all tests instead of just the evpn ones. I added a TBD breadcrumb there for now. We will ultimately need to either add a non-skip check or move the evpn tests out of that testcase.
- Tested on n3,n5(LC capable),n6(non-LC cap),n7(LC cap),n7(non-LC cap),n9

### DIFF
--- a/tests/ciscotest.rb
+++ b/tests/ciscotest.rb
@@ -329,4 +329,18 @@ class CiscoTestCase < TestCase
     shell_command("sudo cp #{filename} /etc/resolv.conf", context)
     shell_command("rm #{filename}", context)
   end
+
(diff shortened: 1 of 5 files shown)


chrisvanheuveln:
Merging my own PR as most of this change has already been reviewed via #385 and we are pulling 1.3.0 shortly...

PR #394, opened 2016-05-05
ios xr implementation of snmpserver

tphoney:
snmp_server for IOS_XR, contact and location

### DIFF
--- a/tests/test_snmpserver.rb
+++ b/tests/test_snmpserver.rb
@@ -32,16 +32,22 @@ class TestSnmpServer < CiscoTestCase
     @default_show_command = 'show run snmp all | no-more'
     # set all commands to defaults
     s = SnmpServer.new
-    s.aaa_user_cache_timeout = s.default_aaa_user_cache_timeout
-    s.contact = s.default_contact
-    s.global_enforce_priv = s.default_global_enforce_priv
-    s.location = s.default_location
-    s.packet_size = 0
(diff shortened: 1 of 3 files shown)


tests/test_snmpserver.rb
```diff
@@ -145,23 +176,34 @@ def test_snmpserver_sys_contact_set_special_chars
...
-    contact = line.to_s.gsub('sys contact:', '').strip unless line.nil?
+    if platform == :nexus
+      line = assert_show_match(command: 'show snmp | no-more',
+                               pattern: /sys contact: [^\n\r]*/)
+      contact = ''
+      contact = line.to_s.gsub('sys contact:', '').strip unless line.nil?
+    else
+      refute_show_match(command: 'show running-config snmp-server',
+                        pattern: /snmp-server contact [^\n\r]+/)
+      contact = ''
```
chburket, line 187: Is there supposed to be a `contact = line.to_s.gsub`... here?

tphoney, line 187: @chburket the iOS_xr does not display contact if it is not set. Therefore there is nothing to gsub.

chburket, line 187: Just a nit, but doesn't that mean you are assuming a default value of ''?  If the default value in the yaml ever changes then this test will fail.  Not going to happen, so I'm fine with it as is.

chburket:
:+1:

PR #397, opened 2016-05-09
fix vni for all supported platforms

mikewiebe:
The `mapped_vni` property should be supported on all nexus platforms except n7k.  Previously we only had support for n3/n9k.  This update fixes that.

Tested on all nxos platforms.

The show command for `n5k` and `n6k` differs in that you have to specify the vlan range to complete the command :(


chrisvanheuveln:
👍

PR #399, opened 2016-05-10
Negotiate auto test fix

robert-w-gries:
### Test Results

`test_duplex`, `test_negotiate_auto_ethernet`, and `test_negotiate_auto_portchannel` pass for the following platforms:
1. n3k
2. n5k
3. n6k
4. n7k
5. n8k
   - physical
   - virtual
6. n9k
   - camden - physical
   - camden - virtual
   - dublin - physical
### Portchannel Error Fix

The [unnecessary XR `UnsupportedError` check](https://github.com/cisco/cisco-network-node-utils/compare/develop...robert-w-gries:feature/negotiate-auto-fix#diff-cf2422f8b2cb8aa033a2f17f39d7354fL768) was causing an issue when re-running the test:

``` bash
  1) Error:
TestInterface#test_negotiate_auto_portchannel:
Cisco::CliError: [interface_channel_group ethernet1/1] The command ' channel-group 10 force' was rejected with error:
command failed: port not compatible [speed]
```

### DIFF
--- a/tests/test_interface.rb
+++ b/tests/test_interface.rb
@@ -756,7 +756,6 @@ class TestInterface < CiscoTestCase
 
       # Platforms raise error unless speed properly configured first
       speeds = valid_speeds(interface)
-      puts "\n\nSPEEDS = #{speeds}\n\n"
       negotiate_auto_helper(interface, 'auto') if speeds.delete('auto')
 
       non_auto = speeds.shift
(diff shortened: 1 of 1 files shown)


mikewiebe:
@robert-w-gries Are these fixes we should be making in our `release_1.3.1` branch?

robert-w-gries:
Yes, I think that's a good idea

PR #400, opened 2016-05-10
Utils.normalize_range_array

chrisvanheuveln:
- The motivation for this change was finding yet another array normalizer in the private vlan code so these changes are initially meant to address fixes there. IMO our range summarize methods were getting a bit long and difficult to maintain; these new methods will hopefully simplify things somewhat.
- The main problem these methods are trying to solve is reconciling manifest inputs with getter results when there are overlapping ranges; e.g.:
  - '2-5, 9, 4-6'     needs to be ['2-6', '9']
  - ['2', '3', '4-6'] needs to be ['2-6']
  - ['2', '3', '4']   needs to be ['2-4']
- It's fairly easy to merge the ranges if they are actual ruby ranges instead of the dash-syntax ranges, so the new methods just convert to ruby ranges, merge, and convert back.
- There's also a method that breaks the ranges down into individual elements so that delta_add_remove can add or remove specific elements on the switch.
- I have a related commit for Puppet on the way.
- Tested on n9k

### DIFF
--- a/lib/cisco_node_utils/cisco_cmn_utils.rb
+++ b/lib/cisco_node_utils/cisco_cmn_utils.rb
@@ -224,5 +224,127 @@ module Cisco
       end
       lranges.to_s.gsub('..', '-').delete('[').delete(']').delete(' ')
     end
+
+    # normalize_range_array
+    #
+    # Given a list of ranges, merge any overlapping ranges and normalize the
+    # them as a string that can be used directly on the switch.
+    #
+    # Note: The ranges are converted to ruby ranges for easy merging,
+    # then converted back to a cli-syntax ranges.
+    #
+    # Accepts an array or string:
+    #   ["2-5", "9", "4-6"]  -or-  '2-5, 9, 4-6'  -or-  ["2-5, 9, 4-6"]
+    # Returns a merged and ordered range:
+    #   ["2-6", "9"]
+    #
+    def self.normalize_range_array(range)
+      return range if range.nil? || range.empty?
+
+      # Handle string within an array: ["2-5, 9, 4-6"] to '2-5, 9, 4-6'
+      range = range.shift if range.is_a?(Array) && range.length == 1
+
+      # Handle string only: '2-5, 9, 4-6' to ["2-5", "9", "4-6"]
+      range = range.split(',') if range.is_a?(String)
+
+      # Convert to ruby-syntax ranges
+      range = dash_range_to_ruby_range(range)
+
+      # Sort & Merge
+      merged = merge_range(range)
+
+      # Convert back to cli dash-syntax
+      ruby_range_to_dash_range(merged)
+    end
+
+    # Convert a cli-dash-syntax range to ruby-range. This is useful for
+    # preparing inputs to merge_range().
+    #
+    # Inputs an array or string of dash-syntax ranges -> returns an array
+    # of ruby ranges.
+    #
+    # Accepts an array or string: ["2-5", "9", "4-6"] or '2-5, 9, 4-6'
+    # Returns an array of ranges: [2..5, 9..9, 4..6]
+    #
+    def self.dash_range_to_ruby_range(range)
+      range = range.split(',') if range.is_a?(String)
+      range.map! do |rng|
+        if rng[/-/]
+          # '2-5' -> 2..5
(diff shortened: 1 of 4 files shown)


lib/cisco_node_utils/cisco_cmn_utils.rb
```diff
@@ -224,5 +224,127 @@ def self.array_to_str(array, sort=true)
...
       lranges.to_s.gsub('..', '-').delete('[').delete(']').delete(' ')
     end
+
+    # normalize_range_array
+    #
+    # Given a list of ranges, merge any overlapping ranges and normalize the
+    # them as a string that can be used directly on the switch.
+    #
+    # Note: The ranges are converted to ruby ranges for easy merging,
+    # then converted back to a cli-syntax ranges.
```
mikewiebe: _nit_ `converted back to a cli-syntax range` or `converted back to cli-syntax rages`

lib/cisco_node_utils/cisco_cmn_utils.rb
```diff
@@ -224,5 +224,127 @@ def self.array_to_str(array, sort=true)
...
+    #   ["2-6", "9"]
+    #
+    def self.normalize_range_array(range)
+      return range if range.nil? || range.empty?
+
+      # Handle string within an array: ["2-5, 9, 4-6"] to '2-5, 9, 4-6'
+      range = range.shift if range.is_a?(Array) && range.length == 1
+
+      # Handle string only: '2-5, 9, 4-6' to ["2-5", "9", "4-6"]
+      range = range.split(',') if range.is_a?(String)
```
mikewiebe, line 248: Technically speaking, I don't think you need the `if range.is_a?(String)` check.  Does not hurt anything, but range should always be a `String` at this point right?

lib/cisco_node_utils/cisco_cmn_utils.rb
```diff
@@ -224,5 +224,127 @@ def self.array_to_str(array, sort=true)
...
+    # This is useful for preparing inputs to delta_add_remove().
+    #
+    # Inputs an array or string of dash-ranges:
+    #  ["2-5", "9", "4-6"] or '2-5, 9, 4-6' or ['2-5, 9, 4-6']
+    # Returns an array of range elements:
+    #  ["2", "3", "4", "5", "6", "9"]
+    #
+    def self.dash_range_to_elements(range)
+      range = range.shift if range.is_a?(Array) && range.length == 1
+      range = range.split(',') if range.is_a?(String)
```
mikewiebe, line 317: Same thing here.. `if range.is_a?(String)` should not be needed.

tests/test_cmn_utils.rb
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env ruby
+# cisco_cmn_utils Unit Tests
```
mikewiebe, line 2: Should we add a TBD to add testcases for all cmn_utils methods?

lib/cisco_node_utils/cisco_cmn_utils.rb
chrisvanheuveln: _nit nit_ `/rages/ranges/` 😃 

fixed

lib/cisco_node_utils/cisco_cmn_utils.rb
chrisvanheuveln, line 248: Actually no. There are three possible input formats: the third one is the multi-element array; e.g. `['2-5', '9', '4-6']`

lib/cisco_node_utils/cisco_cmn_utils.rb
chrisvanheuveln, line 317: ditto non-issue

tests/test_cmn_utils.rb
chrisvanheuveln, line 2: yeah, had the same thought. I'll add it.

lib/cisco_node_utils/cisco_cmn_utils.rb
mikewiebe, line 248: Gotcha.. ok.

lib/cisco_node_utils/cisco_cmn_utils.rb
mikewiebe, line 317: Ack.

PR #406, opened 2016-05-13
interface: private_vlan refactor

chrisvanheuveln:
- There were a few errors with the original implementation. As I worked through this I found that it made sense to rewrite many of the methods to use the new range conversion utilities, and with so many functional changes we decided to take the opportunity to shorten the property names. Since the original code was already released with 1.3.0 we decided to just deprecate them.
- Created `DEPRECATED.yaml` and `interface_DEPRECATED.rb` for the original code. The setters will raise a warning if used, which includes the name of the new property to use. I ran the original test_interface_private_vlan tests against the deprecated code and they still pass, but then updated the test file to use the new methods. This means that going forward the deprecated methods will not have any tests run, and any issues found will not be fixed.
- Related puppet fixes are in the works.
- Tested on n3-I4,n5,n6,n7,n8,n9-I2,n9-I3

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/DEPRECATED.yaml
+++ b/lib/cisco_node_utils/cmd_ref/DEPRECATED.yaml
@@ -11,12 +11,12 @@ _template:
     get_command: "show running interface all"
 
 private_vlan_any:
-  # DEPRECATED (REMOVING AT 2.0.0). USE 'pvlan_any'
+  # DEPRECATED (REMOVING WITH RELEASE 2.0.0). USE 'pvlan_any'
   multiple:
   get_value: '/switchport private-vlan/'
 
 private_vlan_mapping:
-  # DEPRECATED (REMOVING AT 2.0.0). USE 'pvlan_mapping'
+  # DEPRECATED (REMOVING WITH RELEASE 2.0.0). USE 'pvlan_mapping'
   _exclude: [ios_xr, N8k]
   multiple: true
   get_value: '/^private-vlan mapping (.*)$/'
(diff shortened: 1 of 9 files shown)


lib/cisco_node_utils/cmd_ref/DEPRECATED.yaml
```diff
@@ -0,0 +1,95 @@
...
+# These yaml tags are deprecated and will be removed in future releases.
+###############################################################################
+---
+_template:
+  context: ["interface <name>"]
+  nexus:
+    get_command: "show running interface all"
+
+private_vlan_any:
+  # DEPRECATED (REMOVING AT 2.0.0). USE 'pvlan_any'
```
mikewiebe: _nit_ How about `DEPRECATED (REMOVING IN VER: 2.0.0). USE...`

chrisvanheuveln: I changed it to:
`DEPRECATED (REMOVING WITH RELEASE 2.0.0)`

lib/cisco_node_utils/cmd_ref/DEPRECATED.yaml
```diff
@@ -0,0 +1,95 @@
+# DEPRECATED
```
mikewiebe, line 1: Feels a little too general.  Should we use something a bit more descriptive here like you did for `interface_DEPRECATED.yaml` ?

chrisvanheuveln, line 1: It's a common deprecation file. We're going to put the cisco_vlan deprecated yaml entries in there as well. There's no need to separate them.

mikewiebe:
I looked this over pretty thoroughly and it looks good.  Just wondering if we should keep the tests that cover the deprecated api names until we remove them.  Otherwise 👍

chrisvanheuveln:
I considered keeping the old tests. The methods are now EOL though so if we find breakage down the road we won't be fixing those methods; thus the tests are somewhat pointless IMO.

PR #407, opened 2016-05-17
private_vlan refactor for vlan

saichint:
This PR is for refactoring private_vlan association and private_vlan_type to pvlan_association and pvlan_type. 

Created a new file vlan_DEPRECATED.yaml for the original code.
Changed DEPRECATED>yaml to include the 2 properties which are refactored.

ALl minitests passed.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/DEPRECATED.yaml
+++ b/lib/cisco_node_utils/cmd_ref/DEPRECATED.yaml
@@ -1,0 +1,115 @@
+# DEPRECATED
+#
+###############################################################################
+# WARNING. DO NOT USE ANY OF THE YAML TAGS IN THIS FILE.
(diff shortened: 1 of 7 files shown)


lib/cisco_node_utils/cmd_ref/DEPRECATED.yaml
```diff
@@ -15,6 +15,17 @@ private_vlan_any:
   multiple:
```
glennmatthews, line 15: Presumably this should be "vlan_DEPRECATED.yaml"?

chrisvanheuveln, line 15: No, I decided to just put all of the deprecated properties in the same yaml file since they're unique. I don't see an advantage to splitting them into separate files.

PR #409, opened 2016-05-18
feature bfd

saichint:
This PR is for enabling bfd feature which will be used in bfd_global and bfd_neighbor providers.
Tested on all nexus platforms and all pass.


lib/cisco_node_utils/feature.rb
```diff
@@ -23,6 +23,25 @@ class Feature < NodeUtil
...
     # feature disablement for cleanup purposes.
     # ---------------------------
+    def self.bfd_enable
+      return if bfd_enabled?
+      config_set('feature', 'bfd', state: '')
+    end
+
+    def self.bfd_disable
+      return unless bfd_enabled?
+      config_set('feature', 'bfd', state: 'no')
```
mikewiebe: In your email you indicated that bfd at times takes 20-30 seconds to disable.  We should keep and eye on that, especially for tests that disable bfd and then immediately try to re-enable it for the next test.  I would not add logic to handle that unless it's needed.

chrisvanheuveln: As I noted in my email: why do we need this method at all? I'd prefer to remove it to discourage its use.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ Changelog
 ### New feature support
 
 #### Cisco Resources
+* Bfd
+  * feature (@saichint)
```
chrisvanheuveln: The feature.rb work isn't really a resource, it's just an internal api.

lib/cisco_node_utils/feature.rb
mikewiebe: Agreed.  Get rid of bfd_disable.

saichint: I checked this and it is not really taking 25 seconds. The message says it coul;d take upto 25 sec but in minitest everything seems fast. So I did not add any code. For the method, why it is needed at all, I thought that this may be needed in the destroy method of node utils. I was looking at the fabricpath_gobal.rb
https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/fabricpath_global.rb#L101

If I get rid of this disabled method, then do we need destroy method at all? Let me know.

CHANGELOG.md
saichint: Ok, then what should I put here? Can you suggest an alternate comment?

lib/cisco_node_utils/feature.rb
chrisvanheuveln: The reasoning in the FP case was that there was no other meaningful way to turn off FP other than explicitly setting each FP property back to its default (which I still advocate). PIM is another example of this but in the PIM case it just defaults its properties (just one at the moment).

My take on this is that if BFD needs a destroy method then it should follow the PIM model and just set all properties to default; in which case the bfd object will still exist after a destroy but it will be more like an interface object after a destroy -- which may be a bad example because it doesn't reset any properties... though it probably should at least do a 'default interface'... but that's another discussion 😃

CHANGELOG.md
chrisvanheuveln:
```
#### Cisco Resources
* Bidirectional Forwarding Detection
  * bfd (@saichint)
```

saichint:
Ok. Did the changes and tested and all pass.

PR #410, opened 2016-05-19
Remove hardcoded tacacs server reference

robert-w-gries:
Use yaml to define the tacacs server used in the `test_aaa_authorization_service` test.

`test_aaa_authorization_service` will skip if the `tacacs_server.yaml` file is missing, empty, or missing one of the four keys: `host`, `key`, `vrf`, `intf`.

### DIFF
--- a/tests/tacacs_server.yaml.example
+++ b/tests/tacacs_server.yaml.example
@@ -1,0 +1,6 @@
+# Tacacs server configuration for test_aaa_authorization_service.rb
+# Tacacs server must be set up with external access for test to run
+host: '10.122.197.197'
+key:  'testing123'
+vrf:  'management'
+intf: 'mgmt0'
--- a/tests/test_aaa_authorization_service.rb
+++ b/tests/test_aaa_authorization_service.rb
@@ -71,7 +71,7 @@ class TestAaaAuthorizationService < CiscoTestCase
   # Method to pre-configure the user-defined tacacs server in tacacs_server.yaml
   def preconfig_tacacs_server_access(group_name)
     path = File.expand_path('../tacacs_server.yaml', __FILE__)
-    skip('Cannot find tacacs_server.yaml') unless File.file?(path)
+    skip('Cannot find tests/tacacs_server.yaml') unless File.file?(path)
     cfg = YAML.load(File.read(path))
     valid_cfg?(cfg)
     config("tacacs-server key #{cfg['key']}",
@@ -96,8 +96,8 @@ class TestAaaAuthorizationService < CiscoTestCase
   end
 
   def valid_cfg?(cfg)
-    skip('tacacs_server.yaml file is empty') unless cfg
-    msg = 'Missing key in tacacs_server.yaml'
+    skip('tests/tacacs_server.yaml file is empty') unless cfg
+    msg = 'Missing key in tests/tacacs_server.yaml'
     %w(host key vrf intf).each do |key|
       skip("#{msg}: #{key}") if cfg[key].nil?
     end
--- a/tests/test_aaa_authorization_service.rb
+++ b/tests/test_aaa_authorization_service.rb
@@ -107,7 +107,8 @@ class TestAaaAuthorizationService < CiscoTestCase
     test_aaa = config("test aaa server tacacs+ #{host} test test")
     # Valid tacacs server will return message regarding user authentication
     valid = test_aaa[/^user has \S+ authenticat(ed|ion)/]
-    fail("Host '#{host}' is not a valid tacacs server") unless valid
+    fail "Host '#{host}' is either not a valid tacacs server " \
+          'or not reachable' unless valid
   end
 
   def test_create_unsupported_type
--- a/tests/test_aaa_authorization_service.rb
+++ b/tests/test_aaa_authorization_service.rb
@@ -74,13 +74,13 @@ class TestAaaAuthorizationService < CiscoTestCase
     skip('Cannot find tests/tacacs_server.yaml') unless File.file?(path)
     cfg = YAML.load(File.read(path))
     valid_cfg?(cfg)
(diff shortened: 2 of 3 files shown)


tests/tacacs_server.yaml.example
```diff
@@ -0,0 +1,6 @@
+# Tacacs server configuration for test_aaa_authorization_service.rb
+# Tacacs server must be set up with external access for test to run
+host: '10.122.197.197'
+key:  'testing123'
+vrf:  'management'
+intf: 'mgmt0'
```
chrisvanheuveln, line 6: I'm thinking it might be handier to make this the _actual_ `tacacs_server.yaml` file instead of `.example`, and just comment out the keys. Not a big deal if you want to keep it as is.

👍  on the rest

tests/test_aaa_authorization_service.rb
```diff
@@ -92,6 +95,14 @@ def tacacs_groups
     %w(tac_group bxb100 sjc200 rtp10)
   end
 
+  def valid_cfg?(cfg)
```
mikewiebe, line 98: I would like for us to take this one step further and actually verify reach-ability to the tacacs server and skip the aaa tests if it's not reachable.  Ruby has a builtin ping that looks simple enough.

```
irb(main):001:0> require 'ping'
=> true
irb(main):002:0> Ping.pingecho('dt-n9k5-1', 10)  <---- Valid Host
=> true
irb(main):003:0> Ping.pingecho('dt-n9k5-2', 10) <---- Invalid Host
=> false
irb(main):004:0> 
```

tests/test_aaa_authorization_service.rb
```diff
@@ -92,6 +95,14 @@ def tacacs_groups
...
+    skip('tacacs_server.yaml file is empty') unless cfg
+    msg = 'Missing key in tacacs_server.yaml'
```
mikewiebe: Also, please specify the full relative path to the `tacacs_server.yaml` file so users of the tests know exactly where it is to quickly and easily populate it.

tests/tacacs_server.yaml.example
robert-w-gries, line 6: If I made it the actual `tacacs_server.yaml` file, git would want to track the file any time we updated the server info.  It would make it easy to accidentally overwrite the file with a commit of someone's server info.

tests/test_aaa_authorization_service.rb
robert-w-gries, line 98: Working on it.

tests/test_aaa_authorization_service.rb
robert-w-gries: Ack

tests/test_aaa_authorization_service.rb
robert-w-gries, line 98: @mikewiebe I found a way to verify the server without using ping.  It involves [using the `test` command to see if the server is reachable and a tacacs server](http://www.cisco.com/c/en/us/td/docs/switches/datacenter/sw/4_1/nx-os/security/configuration/guide/sec_nx-os-cfg/sec_tacacsplus.html#wpxref30015)

robert-w-gries:
Added logic to test if the user-specified tacacs server is reachable and valid.

#### Valid tacacs server

``` bash
34 runs, 129 assertions, 0 failures, 0 errors, 0 skips
```

#### Unreachable host

``` bash
  1) Error:
TestAaaAuthorizationService#test_type_commands_default_console_group:
RuntimeError: Host '10.10.10.1' is not a valid tacacs server
    tests/test_aaa_authorization_service.rb:110:in `valid_server?'
    tests/test_aaa_authorization_service.rb:83:in `preconfig_tacacs_server_access'
    tests/test_aaa_authorization_service.rb:30:in `setup'
```

#### Invalid host

``` bash
  1) Error:
TestAaaAuthorizationService#test_config_commands_console_unselected_multi:
RuntimeError: Host '10.10.1.1' is not a valid tacacs server
    tests/test_aaa_authorization_service.rb:110:in `valid_server?'
    tests/test_aaa_authorization_service.rb:83:in `preconfig_tacacs_server_access'
    tests/test_aaa_authorization_service.rb:30:in `setup'
```

tests/tacacs_server.yaml.example
chrisvanheuveln, line 6: Good point.

tests/test_aaa_authorization_service.rb
chrisvanheuveln, line 98: Nice.

mikewiebe, line 98: Nice find.  Does it work for all nexus platforms?

mikewiebe:
I would prefer we add a bit more granularity to the error message.  Something along the lines of:
`RuntimeError: Host '10.10.10.1' is either not valid or not reachable.`

tests/test_aaa_authorization_service.rb
robert-w-gries, line 98: Tested platforms:
- N3k
- N5k
- N6k
- N7k
- N8kv
- N9kv

tests/test_aaa_authorization_service.rb
```diff
@@ -92,6 +95,22 @@ def tacacs_groups
...
+      skip("#{msg}: #{key}") if cfg[key].nil?
+    end
+  end
+
+  def valid_server?(host)
+    test_aaa = config("test aaa server tacacs+ #{host} test test")
+    # Valid tacacs server will return message regarding user authentication
+    valid = test_aaa[/^user has \S+ authenticat(ed|ion)/]
+    fail "Host '#{host}' is either not a valid tacacs server " \
+          'or not reachable' unless valid
```
robert-w-gries, line 111: @mikewiebe Done.

mikewiebe:
👍

PR #413, opened 2016-05-23
yum package manager fixes

mikewiebe:
**Summary:**
Previously we used to decompose an rpm file name and use the various bits of information to verify that the rpm package was actually installed.  This was prone to error because of all the different naming conventions used for rpm filenames.  This change instead uses the `show install packages` command to gather the information and then this is compared to the rpm file name.

**Testing:**
- Camden (I2) Release Image
- Deblin (I3) Release Image
- Fretta Camden recent nightly image.

### DIFF
--- a/lib/cisco_node_utils/yum.rb
+++ b/lib/cisco_node_utils/yum.rb
@@ -34,7 +34,6 @@ module Cisco
         version = arch = '' if name_arch == pkg
         # Check for match
         if pkg.match(name_arch) && pkg.match(version) && pkg.match(arch)
-          puts "Found package: #{name_arch}, #{version}, #{arch}"
           return true
         end
       end
--- a/lib/cisco_node_utils/yum.rb
+++ b/lib/cisco_node_utils/yum.rb
@@ -22,38 +22,23 @@ require_relative 'node_util'
 module Cisco
   # This Yum class provides cisco package management functions through nxapi.
   class Yum < NodeUtil
-    def self.decompose_name(file_name)
-      # ex: chef-12.0.0alpha.2+20150319.git.1.b6f-1.el5.x86_64.rpm
-      name_ver_arch_regex = /^([\w\-\+]+)-(\d+\..*)\.(\w{4,})(?:\.rpm)?$/
-
-      # ex n9000_sample-1.0.0-7.0.3.x86_64.rpm
-      name_ver_arch_regex_nx = /^(.*)-([\d\.]+-[\d\.]+)\.(\w{4,})\.rpm$/
-
-      # ex: b+z-ip2.x64_64
-      name_arch_regex = /^([\w\-\+]+)\.(\w+)$/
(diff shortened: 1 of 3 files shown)


lib/cisco_node_utils/yum.rb
```diff
@@ -22,45 +22,31 @@
...
+      patch_data = config_get('yum', 'query_all')
+      patch_data.each do |name_arch, version, _state|
+        # Separate name and architecture
+        next if name_arch.rindex('.').nil?
+        arch = name_arch.slice!(name_arch.rindex('.')..-1).delete('.')
+        # Version/Architecture info not available when only pkg name specified.
+        version = arch = '' if name_arch == pkg
+        # Check for match
+        if pkg.match(name_arch) && pkg.match(version) && pkg.match(arch)
+          puts "Found package: #{name_arch}, #{version}, #{arch}"
```
mikewiebe: Just removed this puts debug line, but github is slow today

lib/cisco_node_utils/yum.rb
```diff
@@ -22,45 +22,30 @@
...
-        else
-          query_name = "#{pkg_info[1]}.#{pkg_info[3]}"
+    def self.validate_installed(pkg)
+      # Sample data returned from config_get('yum', 'query_all')
+      # ["nxos.sample-n8k_EOR.lib32_nxos", "1.0.0-7.0.3.F1.1", "@patching"],
+      patch_data = config_get('yum', 'query_all')
+      patch_data.each do |name_arch, version, _state|
+        # Separate name and architecture
+        next if name_arch.rindex('.').nil?
+        arch = name_arch.slice!(name_arch.rindex('.')..-1).delete('.')
```
chrisvanheuveln, line 32: cleaner:

```
-        next if name_arch.rindex('.').nil?
-        arch = name_arch.slice!(name_arch.rindex('.')..-1).delete('.')
+        arch = name_arch.split('.').last
+        next if arch.nil?
```

mikewiebe, line 32: While it looks nice it's not functionally equivalent.  By using `slice!` I can store the architecture in `arch` and the name without the architecture in `name_arch` without an extra line.

With your code, `name_arch` still contains the entire original name.

```
2.2.1 :001 > name_arch = "nxos.sample-n8k_EOR.lib32_nxos"
 => "nxos.sample-n8k_EOR.lib32_nxos" 
2.2.1 :002 > arch = name_arch.split('.').last
 => "lib32_nxos" 
2.2.1 :003 > name_arch
 => "nxos.sample-n8k_EOR.lib32_nxos" 
```

chrisvanheuveln, line 32: Ah, I missed what you were doing with that. 

`name,sep, arch = name_arch.rpartition('.')` is another way to do it but what you have is fine.

PR #414, opened 2016-05-23
Bfd global

saichint:
This PR is for bfd_global provider NU.
Only a partial list of attributes are implemented in this PR. 
All tests passed on all platforms.


lib/cisco_node_utils/bfd_global.rb
```diff
@@ -0,0 +1,121 @@
...
+
+require_relative 'node_util'
+
+module Cisco
+  # node_utils class for bfd_global
+  class BfdGlobal < NodeUtil
+    attr_reader :name
+
+    def initialize(name, instantiate=true)
+      fail ArgumentError unless name.to_s == 'default'
```
mikewiebe: Why do we need a `name` parameter?  Personally, I would treat this like vtp.

https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/vtp.rb#L28

lib/cisco_node_utils/bfd_global.rb
```diff
@@ -0,0 +1,121 @@
...
+      set_args_keys_default
+
+      Feature.bfd_enable if instantiate
+    end
+
+    def to_s
+      "bfd_global #{name}"
+    end
+
+    def self.globals
```
mikewiebe: Don't think we need `self.globals` either since there are no instances to collect.

lib/cisco_node_utils/bfd_global.rb
```diff
@@ -0,0 +1,121 @@
...
+      hash = {}
+      hash['default'] = BfdGlobal.new('default', false) if
+        Feature.bfd_enabled?
+      hash
+    end
+
+    # Reset everything back to default
```
mikewiebe, line 33: Add `return unless Feature.bfd_enabled?` at the beginning of the `destroy` method.

lib/cisco_node_utils/bfd_global.rb
```diff
@@ -0,0 +1,121 @@
...
+
+    def echo_interface=(val)
+      if val
+        @set_args[:intf] = val
+      else
+        @set_args[:state] = 'no'
+        @set_args[:intf] = echo_interface
+      end
+      config_set('bfd_global', 'echo_interface', @set_args) if
+      @set_args[:intf]
```
mikewiebe: Check indention for `@set_args[:intf]`

lib/cisco_node_utils/bfd_global.rb
```diff
@@ -0,0 +1,121 @@
...
+    def destroy
+      self.echo_interface = default_echo_interface
+      self.fabricpath_vlan = default_fabricpath_vlan if fabricpath_vlan
+      self.startup_timer = default_startup_timer if startup_timer
+      @name = nil
+      set_args_keys_default
+    end
+
+    # Helper method to delete @set_args hash keys
+    def set_args_keys_default
```
mikewiebe, line 43: We have a slightly more efficient way to do this now.
1) Create a `set_args_keys` method (Example): https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/bgp_neighbor.rb#L86
2) Call it in your setters.  This way you don't need to explicitly call `set_args_keys_default` at the end of each setter.  https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/bgp_neighbor.rb#L94

saichint, line 43: Done

saichint:
Made all the changes and retested

lib/cisco_node_utils/bfd_global.rb
```diff
@@ -0,0 +1,101 @@
...
+
+    def default_startup_timer
+      config_get_default('bfd_global', 'startup_timer')
+    end
+
+    def fabricpath_vlan
+      config_get('bfd_global', 'fabricpath_vlan', @get_args)
+    end
+
+    def fabricpath_vlan=(val)
```
chrisvanheuveln, line 91: Does this property require a `Feature.fabricpath_enable` call?

lib/cisco_node_utils/cmd_ref/bfd_global.yaml
```diff
@@ -0,0 +1,27 @@
...
+_exclude: [ios_xr]
+
+_template:
+  get_command: "show running-config bfd all"
+  context: ~
+
+echo_interface:
+  kind: int
+  get_value: '/^bfd echo-interface loopback(\d+)$/'
+  set_value: '<state> bfd echo-interface loopback <intf>'
```
chrisvanheuveln: I see where you're going with this but it seems wrong to specify the interface as just a number; i.e.

```
  echo_interface => '10'          # Bad
  echo_interface => 'loopback10'  # Good
```

tests/test_bfd_global.rb
```diff
@@ -0,0 +1,88 @@
...
+      bg.startup_timer
+  end
+
+  def test_echo_interface
+    bg = BfdGlobal.new
+    config 'interface loopback 10'
+    bg.echo_interface = 10
+    assert_equal(10, bg.echo_interface)
+    bg.echo_interface = bg.default_echo_interface
+    assert_equal(bg.default_echo_interface, bg.echo_interface)
```
chrisvanheuveln: The approach we've been trying to move to is to first test for the default state, then change & test a non-default, then change & test with default:

```
  def test_echo_interface
    config 'interface loopback 10'
    bg = BfdGlobal.new
    default = bg.default_echo_interface
    assert_equal(default, bg.echo_interface)

    bg.echo_interface = 10
    assert_equal(10, bg.echo_interface)

    bg.echo_interface = default
    assert_equal(default, bg.echo_interface)
    config 'no interface loopback 10'
end
```

lib/cisco_node_utils/bfd_global.rb
saichint, line 91: I spoke with Deepak about this. He said the fabricpath need to be enabled for the bfd fabrcipath CLIs to take effect but the user should do it separately in a different manifest since this is bfd global. He said, we should just mention this requirement in the README.md for bfd global. We can discuss further with Deepak if you still think otherwise.

PR #418, opened 2016-05-26
Bfd global

saichint:
This PR is for the remaining properties of the bfd global provider. 
All tests pass on all nexus platforms.
The 'interval' property is excluded on n8k and n9k due to platform bugs. They will be included once the bugs are resolved.

### DIFF
--- a/lib/cisco_node_utils/bfd_global.rb
+++ b/lib/cisco_node_utils/bfd_global.rb
@@ -56,14 +56,11 @@ module Cisco
       self.fabricpath_slow_timer = default_fabricpath_slow_timer if
       fabricpath_slow_timer
       self.startup_timer = default_startup_timer if startup_timer
-      config_set('bfd_global', 'interval', state: 'no',
-                 intv: 50, mrx: 50, mult: 3)
-      config_set('bfd_global', 'ipv4_interval', state: 'no',
-                 intv: 50, mrx: 50, mult: 3) if ipv4_interval
-      config_set('bfd_global', 'ipv6_interval', state: 'no',
-                 intv: 50, mrx: 50, mult: 3) if ipv6_interval
-      config_set('bfd_global', 'fabricpath_interval', state: 'no',
-                 intv: 50, mrx: 50, mult: 3) if fabricpath_interval
+      self.interval = default_interval
+      self.ipv4_interval = default_ipv4_interval if ipv4_interval
+      self.ipv6_interval = default_ipv6_interval if ipv6_interval
+      self.fabricpath_interval = default_fabricpath_interval if
+      fabricpath_interval
       @name = nil
       set_args_keys_default
     end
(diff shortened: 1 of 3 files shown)


lib/cisco_node_utils/bfd_global.rb
```diff
@@ -33,9 +33,24 @@ def initialize(instantiate=true)
...
+      self.ipv4_slow_timer = default_ipv4_slow_timer if ipv4_slow_timer
+      self.ipv6_slow_timer = default_ipv6_slow_timer if ipv6_slow_timer
+      self.fabricpath_slow_timer = default_fabricpath_slow_timer if
+        fabricpath_slow_timer
       self.startup_timer = default_startup_timer if startup_timer
+      self.ipv4_interval = default_ipv4_interval if ipv4_interval
+      self.ipv6_interval = default_ipv6_interval if ipv6_interval
+      self.fabricpath_interval = default_fabricpath_interval if
+        fabricpath_interval
       set_args_keys_default
```
chrisvanheuveln, line 54: There's a lot of code duplication here. I think it would be cleaner to do it this way:

```
    def destroy
      return unless Feature.bfd_enabled?
      [:interval,
       :echo_interface,
       :echo_rx_interval,
      ].each do |prop|
        # Reset each property to its default state
        self.send("#{prop}=", self.send("default_#{prop}")) } if
          self.send prop
      end
    end
```

lib/cisco_node_utils/bfd_global.rb
```diff
@@ -97,5 +210,87 @@ def fabricpath_vlan=(val)
...
+    # CLI: bfd fabricpath interval 100 min_rx 100 multiplier 25
+    def fabricpath_interval
+      config_get('bfd_global', 'fabricpath_interval', @get_args)
+    end
+
+    # interval is an array of interval, min_rx and multiplier
+    # ex: ['100', '100', '25']
+    # CLI: bfd interval 100 min_rx 100 multiplier 25
+    def interval=(arr)
+      set_args_keys(intv: arr[0], mrx: arr[1], mult: arr[2],
```
chrisvanheuveln: nit: I'm all for short variable names but they should still be meaningful, i.e. not require a comment to know what they are. These fields aren't that long so I would just keep the full name for each variable. I'm also a fan of multi-variable assignment, e.g.:

```
  interval, min_rx, multiplier = arr
  set_args_keys(interval: interval, min_rx: min_rx, multiplier: multiplier,
    ...
```

...but again these are just nits so your call.

lib/cisco_node_utils/cmd_ref/bfd_global.yaml
```diff
@@ -12,13 +12,117 @@ echo_interface:
...
+  _exclude: [N5k, N6k]
+  kind: int
+  get_value: '/^bfd echo-rx-interval (\d+)$/'
+  set_value: '<state> bfd echo-rx-interval <rxi>'
+  N3k: &rx_interval_250
+    default_value: 250
+  N7k: &rx_interval_50
+    default_value: 50
+  N8k: *rx_interval_50
+  N9k: *rx_interval_50
```
chrisvanheuveln: This is okay but you can shorten it to:

```
N3k:
    default_value: 250
else:
    default_value: 50
```

tests/test_bfd_global.rb
```diff
@@ -39,11 +39,31 @@ def test_create_destroy
 
     # destroy
     bg.destroy
+    assert_equal(bg.interval, bg.default_interval) if bg.interval
```
chrisvanheuveln: Why the `if` checks on all of these asserts? Shouldn't these all be equal to default regardless?

tests/test_bfd_global.rb
```diff
@@ -89,4 +109,177 @@ def test_startup_timer
...
+  def test_echo_rx_interval
+    bg = BfdGlobal.new
+    if validate_property_excluded?('bfd_global', 'echo_rx_interval')
+      assert_nil(bg.echo_rx_interval)
+      assert_raises(Cisco::UnsupportedError) do
+        bg.echo_rx_interval = 300
+      end
+      return
+    end
+    assert_equal(bg.echo_rx_interval, bg.echo_rx_interval)
```
chrisvanheuveln:
```
-    assert_equal(bg.echo_rx_interval, bg.echo_rx_interval)
+    assert_equal(bg.default_echo_rx_interval, bg.echo_rx_interval)
```

tests/test_bfd_global.rb
```diff
@@ -89,4 +109,177 @@ def test_startup_timer
...
+    bg = BfdGlobal.new
+    arr = %w(200 200 50)
+    if validate_property_excluded?('bfd_global', 'ipv4_interval')
+      assert_nil(bg.ipv4_interval)
+      assert_raises(Cisco::UnsupportedError) do
+        bg.ipv4_interval = arr
+      end
+      return
+    end
+    assert_equal(bg.ipv4_interval, bg.default_ipv4_interval)
```
chrisvanheuveln: Vars are swapped. You want to specify *expected* first:

```
-    assert_equal(bg.ipv4_interval, bg.default_ipv4_interval)
+    assert_equal(bg.default_ipv4_interval, bg.ipv4_interval)
```

tests/test_bfd_global.rb
```diff
@@ -89,4 +109,177 @@ def test_startup_timer
...
+    bg = BfdGlobal.new
+    arr = %w(100 100 25)
+    if validate_property_excluded?('bfd_global', 'interval')
+      assert_nil(bg.interval)
+      assert_raises(Cisco::UnsupportedError) do
+        bg.interval = arr
+      end
+      return
+    end
+    assert_equal(bg.interval, bg.default_interval)
```
chrisvanheuveln: Vars are swapped. You want to specify _expected_ first:

```
-    assert_equal(bg.interval, bg.default_interval)
+    assert_equal(bg.default_interval, bg.interval)
```

Ditto below and also in the ipv4 & ipv6 interval tests.

saichint:
All review comments taken care. Retested on all platforms and all tests pass

PR #419, opened 2016-06-02
Bfd for bgp_neighbor

saichint:
This PR is for adding a new boolean property bfd to bgp_neighbor provider.
Test passed for all nexus platforms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ Changelog
 * Bidirectional Forwarding Detection
   * bfd (@saichint)
 
+### Added
+
+#### Cisco Resources
+* Added a new property bfd for cisco_bgp_neighbor
+
 ## [v1.3.1]
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,8 @@ Changelog
 
 ### Added
 
-#### Cisco Resources
-* Added a new property bfd for cisco_bgp_neighbor
+* Extend cisco_bgp_neighbor with attributes:
+   * `bfd`
 
 ## [v1.3.1]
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ Changelog
 * Bidirectional Forwarding Detection
   * bfd (@saichint)
 
+### Added
+
+#### Cisco Resources
+* Added a new property bfd for cisco_bgp_neighbor
```
chrisvanheuveln: We use this format for new properties on existing providers:

```
### Added

* Extend cisco_bgp_neighbor with attributes:
   * `bfd`
```

tests/test_bgp_neighbor.rb
```diff
@@ -265,6 +265,26 @@ def test_dynamic_capability
     end
   end
 
+  def test_bfd
+    %w(default test_vrf).each do |vrf|
+      neighbor = create_neighbor(vrf)
+      if platform == :ios_xr
```
chrisvanheuveln: This shouldn't be a platform check, it should rely on the yaml exclude; i.e.

```
-      if platform == :ios_xr
+      if validate_property_excluded?('bgp_neighbor', 'bfd')
```

saichint: Will do. The reason I did it this way was because the test_dynamic_capability was written this way before by someone and I just wanted to keep this new property the same way. May be this file is very old and it does not use validate_property_excluded method. 
Anyway, will change it

chrisvanheuveln: Yup, there are probably a few more of those old platform checks floating about. We'll weed them out someday.

PR #422, opened 2016-06-07
Ospf Area

saichint:
This PR is for a new provider ospf_area. All properties except nssa based properties are implemented. virtual-link properties will be implemented in a separate provider.

Due to some "no" cmd not working properly for area range based configs on the nx platforms, some workarounds are added.

All tests have passed on all nexus platforms.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
+++ b/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
@@ -35,7 +35,7 @@ filter_list_out:
 
 range:
   multiple: true
-  get_value: '/^area <area> range (\S+)\s*(?:(not-advertise))?\s*(?:cost (\d+))?$/'
+  get_value: '/^area <area> range (.*)$/'
   set_value: '<state> area <area> range <ip> <not_advertise> <cost> <value>'
   default_value: []
 
--- a/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
+++ b/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
@@ -1,0 +1,45 @@
+# ospf_area
+---
+_exclude: [ios_xr]
+
+_template:
+  get_command: "show running ospf | include area"
+  context:
+    - "router ospf <name>"
+    - "(?)vrf <vrf>"
+
+areas:
+  multiple: true
+  get_value: '/^area\s+(\S+)/'
+
+authentication:
+  get_value: '/^area <area> authentication ?(.*)$/'
+  set_value: '<state> area <area> authentication <auth>'
+  default_value: false
+
+default_cost:
+  kind: int
+  get_value: '/^area <area> default-cost (\d+)$/'
+  set_value: '<state> area <area> default-cost <cost>'
+  default_value: false
+
+filter_list_in:
+  get_value: '/^area <area> filter-list route-map (\S+) in$/'
+  set_value: '<state> area <area> filter-list route-map <route_map> in'
+  default_value: false
+
+filter_list_out:
+  get_value: '/^area <area> filter-list route-map (\S+) out$/'
+  set_value: '<state> area <area> filter-list route-map <route_map> out'
+  default_value: false
+
+range:
+  multiple: true
+  get_value: '/^area <area> range .+$/'
+  set_value: '<state> area <area> range <ip> <not_advertise> <cost>'
+  default_value: []
+
+stub:
+  get_value: '/^area <area> stub ?(.*)$/'
+  set_value: '<state> area <area> stub <stub>'
+  default_value: false
--- a/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
+++ b/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
@@ -45,6 +45,13 @@ range:
   default_value: []
 
 stub:
-  get_value: '/^area <area> stub(?:\s+\S+)?$/'
-  set_value: '<state> area <area> stub <stub>'
+  kind: boolean
+  get_value: '/^area <area> stub/'
+  set_value: '<state> area <area> stub'
+  default_value: false
+
+stub_no_summary:
+  kind: boolean
+  get_value: '/^area <area> stub no-summary$/'
+  set_value: '<state> area <area> stub no-summary'
   default_value: false
--- a/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
+++ b/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
@@ -13,6 +13,11 @@ areas:
   get_value: '/^area\s+(\S+)/'
 
 authentication:
+  # the config can be 'area <area> authentication' or
+  # 'area <area> authentication message-digest' or
+  # no authentication config at all so need to grab the
+  # optional match to get the whole config for checking
+  # the mode
   get_value: '/^area <area> authentication(?:\s+\S+)?$/'
   set_value: '<state> area <area> authentication <auth>'
   default_value: false
--- a/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
+++ b/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
@@ -21,7 +21,7 @@ default_cost:
   kind: int
   get_value: '/^area <area> default-cost (\d+)$/'
   set_value: '<state> area <area> default-cost <cost>'
-  default_value: false
+  default_value: 1
 
 filter_list_in:
   get_value: '/^area <area> filter-list route-map (\S+) in$/'
--- a/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
+++ b/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
@@ -3,7 +3,7 @@
 _exclude: [ios_xr]
 
 _template:
-  get_command: "show running ospf | include area"
+  get_command: "show running ospf | no-more"
   context:
     - "router ospf <name>"
     - "(?)vrf <vrf>"
@@ -13,7 +13,7 @@ areas:
   get_value: '/^area\s+(\S+)/'
 
 authentication:
-  get_value: '/^area <area> authentication ?(.*)$/'
+  get_value: '/^area <area> authentication(?:\s+\S+)?$/'
   set_value: '<state> area <area> authentication <auth>'
   default_value: false
 
@@ -40,6 +40,6 @@ range:
   default_value: []
 
 stub:
-  get_value: '/^area <area> stub ?(.*)$/'
+  get_value: '/^area <area> stub(?:\s+\S+)?$/'
   set_value: '<state> area <area> stub <stub>'
   default_value: false
--- a/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
+++ b/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
@@ -35,8 +35,8 @@ filter_list_out:
 
 range:
   multiple: true
-  get_value: '/^area <area> range .+$/'
-  set_value: '<state> area <area> range <ip> <not_advertise> <cost>'
+  get_value: '/^area <area> range (\S+)\s*(?:(not-advertise))?\s*(?:cost (\d+))?$/'
+  set_value: '<state> area <area> range <ip> <not_advertise> <cost> <value>'
   default_value: []
 
 stub:
--- a/lib/cisco_node_utils/router_ospf_area.rb
+++ b/lib/cisco_node_utils/router_ospf_area.rb
@@ -55,6 +55,7 @@ module Cisco
       hash_final = {}
       RouterOspf.routers.each do |instance|
         name = instance[0]
+        # get all area ids under default vrf
         area_ids = config_get('ospf_area', 'areas', name: name)
         unless area_ids.nil?
           area_ids.uniq.each do |area|
@@ -65,6 +66,7 @@ module Cisco
         vrf_ids = config_get('ospf', 'vrf', name: name)
         next if vrf_ids.nil?
         vrf_ids.each do |vrf|
+          # get all area ids under each vrf
           area_ids = config_get('ospf_area', 'areas', name: name, vrf: vrf)
           next if area_ids.nil?
           area_ids.uniq.each do |area|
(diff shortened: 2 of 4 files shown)


lib/cisco_node_utils/cmd_ref/ospf_area.yaml
```diff
@@ -0,0 +1,45 @@
...
+  context:
+    - "router ospf <name>"
+    - "(?)vrf <vrf>"
+
+areas:
+  multiple: true
+  get_value: '/^area\s+(\S+)/'
+
+authentication:
+  get_value: '/^area <area> authentication(?:\s+\S+)?$/'
```
mikewiebe, line 21: Please include a comment here to indicate why the optional match is needed.

lib/cisco_node_utils/router_ospf_area.rb
```diff
@@ -0,0 +1,290 @@
...
+    ########################################################
+    #                      PROPERTIES                      #
+    ########################################################
+
+    # CLI can be either of the following or none
+    # area 1.1.1.1 authentication
+    # area 1.1.1.1 authentication message-digest
+    def authentication
+      auth = config_get('ospf_area', 'authentication', @get_args)
+      return auth unless auth
```
mikewiebe: I think it would be clearer to `return default_authentication unless auth`.

lib/cisco_node_utils/router_ospf_area.rb
```diff
@@ -0,0 +1,290 @@
...
+
+    def create
+      # create RouterOspfVrf only
+      # area_id is used at each config cli, not as a context
+      # ex: area 1.1.1.1 default-cost 1000
+      # area 1.1.1.1 authentication message-digest
+      RouterOspfVrf.new(@router, @vrf)
+    end
+
+    def destroy
```
mikewiebe, line 89: We can talk about this more in the parking lot today, but I would have expected `destroy` to remove the area, not just reset the area properties back to default, but this could be tricky since the properties themselves would have to be namevars.

saichint, line 89: @mikewiebe 
As discussed, removed the create method for this and leave the rest. Tested self.areas method as well (using puppet module code). No minitest yet for this. Will add it after I get an example.

lib/cisco_node_utils/router_ospf_area.rb
```diff
@@ -0,0 +1,285 @@
...
+      fail ArgumentError unless IPAddr.new(area_id).ipv4?
+
+      @router = ospf_router
+      @vrf = vrf_name
+      @area_id = area_id
+
+      set_args_keys_default
+    end
+
+    def self.areas
```
chrisvanheuveln, line 46: Try this instead:

```
    def self.areas
      hash = {}
      RouterOspf.routers.each do |name, _obj|
        hash[name] = {}

        area_ids = config_get('ospf_area', 'areas', name: name)
        if area_ids
          hash[name]['default'] = {}
          area_ids.uniq.each do |area|
            hash[name]['default'][area] =
              RouterOspfArea.new(name, 'default', area, false)  # HAD A TYPO HERE EARLIER
          end
        end

        vrf_ids = config_get('ospf', 'vrf', name: name)
        next if vrf_ids.nil?
        vrf_ids.each do |vrf|
          hash[name][vrf] = {}

          area_ids = config_get('ospf_area', 'areas', name: name, vrf: vrf)
          next if area_ids.nil?
          area_ids.uniq.each do |area|
            hash[name][vrf][area] = RouterOspfArea.new(name, vrf, area, false)
          end
        end
      end
      hash
    end
```

saichint, line 46: If there are no area objects at all under the given ospf router, the code I put before returns 
hash = {}, which is what we expect. However, this code, returns hash[name] = {} even if there is nothing under the router. I don't know if that would be correct, any thoughts?

lib/cisco_node_utils/router_ospf_area.rb
```diff
@@ -0,0 +1,285 @@
...
+      fail ArgumentError unless vrf_name.length > 0
+      fail ArgumentError unless area_id.length > 0
+
+      # check the area_id is a proper ipv4 address
+      fail ArgumentError if (begin
+                               IPAddr.new(area_id)
+                             rescue
+                               nil
+                             end).nil?
+      fail ArgumentError unless IPAddr.new(area_id).ipv4?
```
chrisvanheuveln: Area IDs can be a simple Integer so these IPAddr calls are wrong.

I would just to_s all of the vars above and get rid of the IPAddr checks, i.e.:

```
-      fail TypeError unless ospf_router.is_a?(String)
-      fail TypeError unless vrf_name.is_a?(String)
-      fail TypeError unless area_id.is_a?(String)
-      fail ArgumentError unless ospf_router.length > 0
-      fail ArgumentError unless vrf_name.length > 0
-      fail ArgumentError unless area_id.length > 0
-      # check the area_id is a proper ipv4 address
-      fail ArgumentError if (begin
-                               IPAddr.new(area_id)
-                             rescue
-                               nil
-                             end).nil?
-      fail ArgumentError unless IPAddr.new(area_id).ipv4?
-      @router = ospf_router
-      @vrf = vrf_name
-      @area_id = area_id
+      @router = ospf_router.to_s
+      @vrf = vrf_name.to_s
+      @area_id = area_id.to_s
+      fail ArgumentError if @router.empty? || @vrf.empty? || @area_id.empty?
```

lib/cisco_node_utils/router_ospf_area.rb
```diff
@@ -0,0 +1,285 @@
...
+        end
+      end
+      hash
+    end
+
+    # Helper method to delete @set_args hash keys
+    def set_args_keys_default
+      @set_args = { name: @router }
+      @set_args[:vrf] = @vrf unless @vrf == 'default'
+      @set_args[:area] = @area_id
```
chrisvanheuveln:
```
-      @set_args = { name: @router }
+      @set_args = { name: @router, area: @area_id }
+      @set_args[:vrf] = @vrf unless @vrf == 'default'
-      @set_args[:area] = @area_id
```

lib/cisco_node_utils/router_ospf_area.rb
saichint: Area IDs can be simple integers, however, the backend platform code converts the intgeter to ipaddr and the show command only shows them as ipv4 address. 
for ex:
agent-lab13-nx(config)# router ospf abcd
agent-lab13-nx(config-router)# area 123 stub
agent-lab13-nx(config-router)# sh running-config ospf

!Command: show running-config ospf
!Time: Wed Jun  8 10:50:52 2016

version 7.0(3)I3(1)
feature ospf

router ospf abcd
  area 0.0.0.123 stub

If we do allow, we need to keep converting the integers to ipv4 addresses and use them internally. I thought it is unnecessary to allow 2 types of inputs when the output is always ipv4. So basically forcing the user to give us the actual ipv4 address. Let me know if you disagree.

lib/cisco_node_utils/router_ospf_area.rb
```diff
@@ -0,0 +1,285 @@
...
+    end
+
+    # rubocop:disable Style/AccessorMethodName
+    def set_args_keys(hash={})
+      set_args_keys_default
+      @set_args = @get_args.merge!(hash) unless hash.empty?
+    end
+
+    def destroy
+      return unless Feature.ospf_enabled?
```
chrisvanheuveln, line 90: This probably needs a second check... I'm thinking about the case where I've just created this object but I haven't set any properties yet, but then follow immediately with a destroy. The result on the device will be a be a router and vrf configuration where maybe there was none before. 

edit: A solution is to just call your `areas` method and check to see if the vrf has an empty hash. And with this check you won't need to check for Feature.ospf_enabled:

```
-      return unless Feature.ospf_enabled?
+      current = RouterOspfArea.areas
+      return unless current[@router] && current[@router][@vrf]

```

lib/cisco_node_utils/router_ospf_area.rb
chrisvanheuveln, line 46: It's okay in that case but you can do this if you want:

```
      RouterOspf.routers.each do |name, obj|
-      hash[name] = {}

        area_ids = config_get('ospf_area', 'areas', name: name)
        if area_ids
+         hash[name] = {}
          hash[name]['default'] = {}
...
        vrf_ids.each do |vrf|
+         hash[name] ||= {}
```

lib/cisco_node_utils/router_ospf_area.rb
```diff
@@ -0,0 +1,285 @@
...
+    # area 1.1.1.1 range 10.3.0.0/16 not-advertise cost 23
+    # sometimes the not-advertise and cost are reversed in the
+    # show command for reasons unknown!
+    # area 1.1.1.1 range 10.3.0.0/16 cost 23 not-advertise
+    # use positional way of getting the values as it is
+    # simple and there are only two properties which are
+    # optional. ip is mandatory
+    # the return list is of the form [ip, not-advertise, cost]
+    # ex: [['10.3.0.0/16', true, '23'], ['10.3.0.0/32', true, false],
+    #      ['10.3.0.1/32', false, false], ['10.3.3.0/24', false, '450']]
```
chrisvanheuveln: While I have no doubt that this works, I have reservations for making the user follow this clunky positional syntax in a manifest. IMO, this should be broken down into three properties:

```
  range => '10.3.0.0/16',
  range_advertise => false,    # TBD: false == not-advertise ?
  range_cost => 23,
```

...then use a single setter for all three.

lib/cisco_node_utils/router_ospf_area.rb
```diff
@@ -0,0 +1,285 @@
...
+      end
+    end
+
+    def default_range
+      config_get_default('ospf_area', 'range')
+    end
+
+    # CLI can be either of the following or none
+    # area 1.1.1.1 stub
+    # area 1.1.1.1 stub no-summary
```
chrisvanheuveln: What will the manifest look like for this one? The yaml shows the default as `false` but the getter returns `no_summary` or `summary`. 

I _think_ this should be:

```
  stub => 'summary',
  stub => 'no-summary',
  stub => 'default',     # defaults to 'summary'
```

...or else you should consider a second property, but I find that a little confusing:

```
  stub => true,
  stub_summary => false,
```

tests/test_router_ospf_area.rb
```diff
@@ -0,0 +1,172 @@
...
+    super
+  end
+
+  def create_routerospfarea_default(router='Wolfpack', name='default',
+                                    area_id='1.1.1.1')
+    RouterOspfArea.new(router, name, area_id)
+  end
+
+  def create_routerospfarea_vrf(router='Wolfpack', name='blue',
+                                area_id='1.1.1.1')
```
chrisvanheuveln: Change area_id to a simple integer so that you cover that case too (without having to add an explicit test case).

lib/cisco_node_utils/router_ospf_area.rb
saichint: Actually this is a multiple array setter.
So the manifest is like:
[['10.3.0.0/16', true, '23'], ['10.3.0.0/32', true, false], ['10.3.0.1/32', false, false]]

The one you gave me above works only for one array but not for arrays within array. I agree the way I have done is not the best but it is simple. Also there is a show command issue where the it is shifting the properties for some unknown reason randomly.

lib/cisco_node_utils/router_ospf_area.rb
saichint: it is a tristate variable.
stub => 'summary',
  stub => 'no-summary',
  stub => 'default',     # this removes the config entirely using 'no' command.
This is very similar to authentication property with small difference.

tests/test_router_ospf_area.rb
saichint: please look at the comment I gave for area_id integer above. I think we have some confusion about this PR. Can we discuss tomorrow as there are lot of comments here which I am not sure how to go about?

lib/cisco_node_utils/router_ospf_area.rb
chrisvanheuveln: We should allow the user to enter either type but we need to normalize it to v4 syntax.

How about this:

```
+      @area_id = area_id.to_s
+      raise ArgumentError if @area_id.empty?
+      # Convert to dot-notation
+      @area_id = IPAddr.new(area_id.to_i,Socket::AF_INET) unless @area_id[/\./]
```

lib/cisco_node_utils/router_ospf_area.rb
chrisvanheuveln: Ah, I missed the obvious :-(

Okay, let's try to come up with a cleaner way to do this with nested arrays -- maybe use more meaningful values (true / false are not that helpful) and allow for defaults; e.g.:

```
  range => [
            ['10.3.0.0/16', 'not-advertise', 23],
            ['10.3.0.0/32', 'not-advertise'],
            ['10.3.0.1/32'],
           ]
```

We do something similar with `redistribute` statements, where some have route-maps and others don't. In this case each value is easy for our code to distinguish between a range address, an advertise flag, and a cost value.

tests/test_router_ospf_area.rb
chrisvanheuveln: My comment still holds - we need to support integers in the manifest. OSPF used integer notation from day one, the dotted notation came later. We can have a call tomorrow to discuss in more detail.

lib/cisco_node_utils/router_ospf_area.rb
chrisvanheuveln: It's similar but it doesn't work for me in this case, since the default stub behavior is to summarize,  which will be confusing in a manifest. Let's reconsider this one.

lib/cisco_node_utils/router_ospf_area.rb
saichint, line 46: Done

lib/cisco_node_utils/router_ospf_area.rb
saichint: Done

lib/cisco_node_utils/router_ospf_area.rb
saichint, line 90: Done in a different way

saichint:
Fixed all the review comments. Retested on all platforms. On n8k, there is a cosmetic issue ->
If we set any property on the area, default-cost gets shown up in the running_config.
for ex:
n9k-157(config-router)# area 1 stub
n9k-157(config-router)# sh running-config ospf

!Command: show running-config ospf
!Time: Thu Jun  9 14:54:46 2016

version 7.0(3)F1(1)
feature ospf

router ospf a
  area 0.0.0.1 stub
  area 0.0.0.1 default-cost 1 ==> this cannot be removed until the ospf router is removed. This is only cosmetic as the default value for default_cost is 1 only, though it does not show up on any other platform.

So one of the tests need to be skipped partially on n8k only.

PR #423, opened 2016-06-07
Switchport trunk allowed vlan now accepts 'vlan add' lines

robert-w-gries:
Fix for cisco/cisco-network-puppet-module#344
### Description

This PR fixes the regexp for `switchport trunk allowed vlan` and `switchport private-vlan trunk allowed vlan` properties such that the only matched values are `all`, `none`, or `[vlan_ids]`.

This PR also extends the testcases for these properties.  Now, the tests will manually add an extra vlan to an existing range and ensure that our getter will retrieve the proper vlan range.

``` bash
(config-if)# switchport trunk allowed vlan 500-528,530,532,534,587,590-593
(config-if)# switchport trunk allowed vlan add 597
```

Note:  Behavior for these config commands differs between images, which needed to be handled in the testcase.
#### N9k I2

``` bash
interface Ethernet1/2
  switchport
  switchport trunk allowed vlan 500-528,530,532,534,587,590-593,597
```
#### N9k I3

``` bash
interface Ethernet1/2
  switchport
  switchport trunk allowed vlan 500-528,530,532,534,587,590-593
  switchport trunk allowed vlan add 597
```
### Tested Platforms
- n3k - I2
- n5k
- n6k
- n7k
- n8kv
- n9k - I3
- n9kv - I2

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/interface.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface.yaml
@@ -446,7 +446,7 @@ switchport_pvlan_promiscuous:
 
 switchport_pvlan_trunk_allowed_vlan:
   _exclude: [ios_xr, N8k]
-  get_value: '/^switchport private-vlan trunk allowed vlan (.*)$/'
+  get_value: '/^switchport private-vlan trunk allowed vlan (\S+)$/'
   set_value: "switchport private-vlan trunk allowed vlan <range>"
   default_value: 'none'
 
@@ -479,7 +479,7 @@ switchport_pvlan_trunk_secondary:
 
 switchport_trunk_allowed_vlan:
   _exclude: [ios_xr]
-  get_value: '/^switchport trunk allowed vlan (.*)$/'
+  get_value: '/^switchport trunk allowed vlan (\S+)$/'
   set_value: "<state> switchport trunk allowed vlan <vlan>"
   default_value: "1-4094"
   # TBD: default should be 'none'; no need for state. See pvlan version.
--- a/lib/cisco_node_utils/cmd_ref/interface.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface.yaml
@@ -446,7 +446,8 @@ switchport_pvlan_promiscuous:
 
 switchport_pvlan_trunk_allowed_vlan:
   _exclude: [ios_xr, N8k]
-  get_value: '/^switchport private-vlan trunk allowed vlan (\S+)$/'
+  multiple: true
+  get_value: '/^switchport private-vlan trunk allowed vlan (?:add )?(\S+)$/'
   set_value: "switchport private-vlan trunk allowed vlan <range>"
   default_value: 'none'
 
@@ -479,12 +480,10 @@ switchport_pvlan_trunk_secondary:
 
 switchport_trunk_allowed_vlan:
   _exclude: [ios_xr]
-  get_value: '/^switchport trunk allowed vlan (\S+)$/'
+  multiple: true
+  get_value: '/^switchport trunk allowed vlan (?:add )?(\S+)$/'
   set_value: "<state> switchport trunk allowed vlan <vlan>"
-  default_value: "1-4094"
-  # TBD: default should be 'none'; no need for state. See pvlan version.
-  # set_value: "switchport trunk allowed vlan <range>"
-  # default_value: 'none'
+  default_value: '1-4094'
 
 switchport_trunk_native_vlan:
   _exclude: [ios_xr]
--- a/lib/cisco_node_utils/interface.rb
+++ b/lib/cisco_node_utils/interface.rb
@@ -933,7 +933,11 @@ module Cisco
     end
 
     def switchport_trunk_allowed_vlan
-      config_get('interface', 'switchport_trunk_allowed_vlan', name: @name)
+      vlans = config_get('interface', 'switchport_trunk_allowed_vlan',
+                         name: @name)
+      vlans = vlans.join(',') if vlans.is_a?(Array)
+      vlans = Utils.normalize_range_array(vlans, :string) unless vlans == 'none'
+      vlans
     end
 
     def switchport_trunk_allowed_vlan=(val)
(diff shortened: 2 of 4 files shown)


mikewiebe:
@robert-w-gries I think we should put this into the release branch.

lib/cisco_node_utils/cmd_ref/interface.yaml
```diff
@@ -479,7 +479,7 @@ switchport_pvlan_trunk_secondary:
 
 switchport_trunk_allowed_vlan:
   _exclude: [ios_xr]
-  get_value: '/^switchport trunk allowed vlan (.*)$/'
+  get_value: '/^switchport trunk allowed vlan (\S+)$/'
```
mikewiebe: This does not look quite right to me.

```
2.2.1 :008 >   line = 'switchport trunk allowed vlan add 597'
 => "switchport trunk allowed vlan add 597" 
2.2.1 :009 > line.match(/^switchport trunk allowed vlan (\S+)$/)
 => nil 
2.2.1 :010 > 
```

We need get all instances of allowed vlans, even if they span multiple config lines right?

**Update:** Chris and I talked and see that this is actually expected behavior.  As soon as the range gets too long it gets appended on the next line with the `add` keyword.
- Make `add` optional in the yaml regexp
- Make it a `multiple:` in the yaml block.
- Merge all lines you find for the range.

robert-w-gries:
@mikewiebe @chrisvanheuveln 

Tested the following testcases across all platforms:
- `tests/test_interface_private_vlan.rb`
  - `test_switchport_pvlan_trunk_allowed_vlan`
- `tests/test_interface_switchport.rb`
  - `test_sw_trunk_allowed_vlan`

tests/test_interface_private_vlan.rb
```diff
@@ -323,10 +323,12 @@ def test_switchport_pvlan_trunk_allowed_vlan
     assert_equal(default, i.switchport_pvlan_trunk_allowed_vlan)
 
     vlans = '500-528,530,532,534,587,590-593'
+    all_vlans = '500-528,530,532,534,587,590-593,597-598'
     config("interface #{i.name}",
            "switchport private-vlan trunk allowed vlan #{vlans}",
-           'switchport private-vlan trunk allowed vlan add 597')
-    assert_equal(vlans + ',597', i.switchport_pvlan_trunk_allowed_vlan)
+           'switchport private-vlan trunk allowed vlan add 597',
+           'switchport private-vlan trunk allowed vlan add 598')
```
chrisvanheuveln, line 330: nit: you can just do this instead of two configs:

```
+           'switchport private-vlan trunk allowed vlan add 597-598')
```

...or even just get rid of `vlans` and just use `all_vlans`, since that will create the `add 597-598` config:

```
     config("interface #{i.name}",
             "switchport private-vlan trunk allowed vlan #{all_vlans}")
```

chrisvanheuveln:
👍   The nits are your call.

lib/cisco_node_utils/interface.rb
```diff
@@ -933,7 +933,11 @@ def default_switchport_mode
     end
 
     def switchport_trunk_allowed_vlan
-      config_get('interface', 'switchport_trunk_allowed_vlan', name: @name)
+      vlans = config_get('interface', 'switchport_trunk_allowed_vlan',
+                         name: @name)
+      vlans = vlans.join(',') if vlans.is_a?(Array)
+      vlans = Utils.normalize_range_array(vlans, :string) unless vlans == 'none'
```
mikewiebe, line 939: When would `vlans == 'none'` for this property?  The default value is it not `none` for this prop.

tests/test_interface_private_vlan.rb
```diff
@@ -321,6 +321,14 @@ def test_switchport_pvlan_trunk_allowed_vlan
 
     i.switchport_pvlan_trunk_allowed_vlan = default
     assert_equal(default, i.switchport_pvlan_trunk_allowed_vlan)
+
+    vlans = '500-528,530,532,534,587,590-593'
+    all_vlans = '500-528,530,532,534,587,590-593,597-598'
+    config("interface #{i.name}",
+           "switchport private-vlan trunk allowed vlan #{vlans}",
```
mikewiebe, line 328: Does this test result in multiple add lines on all platforms?  If not please make sure to modify it so that it does.

tests/test_interface_switchport.rb
```diff
@@ -280,6 +279,16 @@ def test_sw_trunk_allowed_vlan
 
       interface.switchport_trunk_allowed_vlan = '20, 30'
       assert_equal('20,30', interface.switchport_trunk_allowed_vlan)
+
+      # Some images have behavior where 'vlan add' is separate line
+      # This behavior is triggered for vlan ranges that exceed character limit
+      vlans = '500-528,530,532,534,587,590-593'
```
mikewiebe, line 285: Same comment here for the multiple add lines.

lib/cisco_node_utils/interface.rb
robert-w-gries, line 939: I think you were looking at the wrong method?  This is the getter method

chrisvanheuveln, line 939: It is and it isn't. :-(

pvlan correctly uses none:
https://github.com/cisco/cisco-network-node-utils/blob/release_1.3.1/lib/cisco_node_utils/cmd_ref/interface.yaml#L447

non-pvlan is wrong, has a TBD to fix it like pvlan:
https://github.com/cisco/cisco-network-node-utils/blob/release_1.3.1/lib/cisco_node_utils/cmd_ref/interface.yaml#L480

edit: nm, Rob's fix is correct.

tests/test_interface_private_vlan.rb
robert-w-gries, line 328: As far as I can tell, `switchport private_vlan trunk allowed vlan` does not have the multi-line behavior that `switchport trunk allowed vlan` does, but I added in the logic to handle it if the behavior is added in later on.

tests/test_interface_switchport.rb
robert-w-gries, line 285: This testcase results in multiple `vlan` lines for `n7k` and `n9k-I3`:

``` bash
n9k-110(config-if)# switchport trunk allowed vlan 500-528,530,532,534,587,590-593
n9k-110(config-if)# switchport trunk allowed vlan add 597
n9k-110(config-if)# switchport trunk allowed vlan add 598
n9k-110(config-if)# show running-config interface ethernet 1/5

!Command: show running-config interface Ethernet1/5
!Time: Mon Jun 13 14:11:51 2016

version 7.0(3)I3(1)

interface Ethernet1/5
  switchport
  switchport trunk allowed vlan 500-528,530,532,534,587,590-593
  switchport trunk allowed vlan add 597-598
```

tests/test_interface_private_vlan.rb
robert-w-gries, line 330: 👍 Testing the change then pushing it

chrisvanheuveln:
👍

lib/cisco_node_utils/interface.rb
mikewiebe, line 939: I was assuming `none` was coming back from the yaml in the default case.  I did not realize you could actually configure a value of `none` on the device.  All is good then 👍

PR #432, opened 2016-06-16
Ospf_area with nssa

saichint:
This PR is for router_ospf_area provider which includes nssa properties.
All tests pass on on platforms except on n8k.
On n8k, commands like "area <area> nssa default-information-originate" and "area <area> nssa default-information-originate route-map <route-map>" are not getting configured properly after we reset nssa and due to this the relevant tests failed. Bug is filed on nxos platform and skipping these tests only on n8k.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
+++ b/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
@@ -35,7 +35,7 @@ filter_list_out:
 
 range:
   multiple: true
-  get_value: '/^area <area> range (\S+)\s*(?:(not-advertise))?\s*(?:cost (\d+))?$/'
+  get_value: '/^area <area> range (.*)$/'
   set_value: '<state> area <area> range <ip> <not_advertise> <cost> <value>'
   default_value: []
 
--- a/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
+++ b/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
@@ -1,0 +1,45 @@
+# ospf_area
+---
+_exclude: [ios_xr]
+
+_template:
+  get_command: "show running ospf | include area"
+  context:
+    - "router ospf <name>"
+    - "(?)vrf <vrf>"
+
+areas:
+  multiple: true
+  get_value: '/^area\s+(\S+)/'
+
+authentication:
+  get_value: '/^area <area> authentication ?(.*)$/'
+  set_value: '<state> area <area> authentication <auth>'
+  default_value: false
+
+default_cost:
+  kind: int
+  get_value: '/^area <area> default-cost (\d+)$/'
+  set_value: '<state> area <area> default-cost <cost>'
+  default_value: false
+
+filter_list_in:
+  get_value: '/^area <area> filter-list route-map (\S+) in$/'
+  set_value: '<state> area <area> filter-list route-map <route_map> in'
+  default_value: false
+
+filter_list_out:
+  get_value: '/^area <area> filter-list route-map (\S+) out$/'
+  set_value: '<state> area <area> filter-list route-map <route_map> out'
+  default_value: false
+
+range:
+  multiple: true
+  get_value: '/^area <area> range .+$/'
+  set_value: '<state> area <area> range <ip> <not_advertise> <cost>'
+  default_value: []
+
+stub:
+  get_value: '/^area <area> stub ?(.*)$/'
+  set_value: '<state> area <area> stub <stub>'
+  default_value: false
--- a/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
+++ b/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
@@ -45,6 +45,13 @@ range:
   default_value: []
 
 stub:
-  get_value: '/^area <area> stub(?:\s+\S+)?$/'
-  set_value: '<state> area <area> stub <stub>'
+  kind: boolean
+  get_value: '/^area <area> stub/'
+  set_value: '<state> area <area> stub'
+  default_value: false
+
+stub_no_summary:
+  kind: boolean
+  get_value: '/^area <area> stub no-summary$/'
+  set_value: '<state> area <area> stub no-summary'
   default_value: false
--- a/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
+++ b/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
@@ -13,6 +13,11 @@ areas:
   get_value: '/^area\s+(\S+)/'
 
 authentication:
+  # the config can be 'area <area> authentication' or
+  # 'area <area> authentication message-digest' or
+  # no authentication config at all so need to grab the
+  # optional match to get the whole config for checking
+  # the mode
   get_value: '/^area <area> authentication(?:\s+\S+)?$/'
   set_value: '<state> area <area> authentication <auth>'
   default_value: false
--- a/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
+++ b/lib/cisco_node_utils/cmd_ref/ospf_area.yaml
@@ -21,7 +21,7 @@ default_cost:
   kind: int
   get_value: '/^area <area> default-cost (\d+)$/'
   set_value: '<state> area <area> default-cost <cost>'
-  default_value: false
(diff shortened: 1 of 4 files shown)


lib/cisco_node_utils/cmd_ref/ospf_area.yaml
```diff
@@ -38,6 +38,35 @@ filter_list_out:
   set_value: '<state> area <area> filter-list route-map <route_map> out'
   default_value: false
 
+nssa:
+  multiple: true
+  get_value: '/^area <area> nssa(.*)$/'
+  set_value: '<state> area <area> nssa <no_summary> <no_redistribution> <default_information_originate> <route_map> <rm>'
+
+nssa_def_info_originate:
```
chrisvanheuveln: `/nssa_def_info_originate/nssa_default_originate/`

We already have bgp properties that follow this format so let's not deviate.

lib/cisco_node_utils/router_ospf_area.rb
```diff
@@ -185,6 +187,149 @@ def default_filter_list_out
...
+    # no-redistribution and/or default-information-originate.
+    # route-map <map> can be appended with default-information-originate
+    # Basically, every property this CLI configures is optional
+    def nssa
+      hash = {}
+      output = config_get('ospf_area', 'nssa', @get_args)
+      return hash if output.nil?
+      output.each do |line|
+        next if line.include?('translate')
+        hash[:nssa_enable] = true
```
chrisvanheuveln: I don't quite understand why you need this `nssa_enable`. Is this an explicit property that will show up in the manifest? It would help if you could include a sample manifest for these changes.

lib/cisco_node_utils/router_ospf_area.rb
```diff
@@ -185,6 +187,149 @@ def default_filter_list_out
...
+          hash[:route_map] = params[params.index('route-map') + 1]
+        end
+      end
+      hash
+    end
+
+    def nssa_enable
+      hash = nssa
+      ne = hash[:nssa_enable] ? hash[:nssa_enable] : default_nssa_enable
+      ne
```
chrisvanheuveln: This `ne` assignment isn't needed.

lib/cisco_node_utils/router_ospf_area.rb
```diff
@@ -185,6 +187,149 @@ def default_filter_list_out
...
+          params = line.split
+          hash[:route_map] = params[params.index('route-map') + 1]
+        end
+      end
+      hash
+    end
+
+    def nssa_enable
+      hash = nssa
+      ne = hash[:nssa_enable] ? hash[:nssa_enable] : default_nssa_enable
```
chrisvanheuveln: I would keep this simple and just do:

```
    def nssa_enable
      nssa[:nssa_enable].nil? ? false : true
    end
```

...but I still want to know why it's needed at all.

lib/cisco_node_utils/router_ospf_area.rb
```diff
@@ -185,6 +187,149 @@ def default_filter_list_out
...
+
+    def default_nssa_enable
+      config_get_default('ospf_area', 'nssa_enable')
+    end
+
+    def nssa_def_info_originate
+      hash = nssa
+      ndio = default_nssa_def_info_originate
+      ndio = hash[:def_info_origin] if hash[:def_info_origin]
+      ndio
```
chrisvanheuveln: Similar comment here and in the other methods...

```
    def nssa_def_info_originate
      val = nssa[:def_info_origin]
      val.nil? ? default_nssa_def_info_originate : val
```

lib/cisco_node_utils/cmd_ref/ospf_area.yaml
```diff
@@ -38,6 +38,35 @@ filter_list_out:
...
+  kind: boolean
+  default_value: false
+
+nssa_enable:
+  kind: boolean
+  default_value: false
+
+nssa_no_redistribution:
```
chrisvanheuveln, line 56: I dislike using 'no' properties that disable behavior when they're set to true. It's confusing just to write that.

In other words, I prefer:

| behavior | cli | better | worse |
| :-- | :-- | :-- | :-- |
| disable<br>redistribution | `nssa no-redistribute` | `redistribution => false` | `no_redistribution => true` |
| redistribution<br> by default | `nssa` | `redistribution => <true|default>` | `no_redistribution => false` |

...but this is a personal preference. I'd like to get some feedback from Mike and Rob as well.

lib/cisco_node_utils/router_ospf_area.rb
```diff
@@ -185,6 +187,149 @@ def default_filter_list_out
...
+      nrm = hash[:route_map] if hash[:route_map]
+      nrm
+    end
+
+    def default_nssa_route_map
+      config_get_default('ospf_area', 'nssa_route_map')
+    end
+
+    def nssa_set(enable, def_info_originate, no_redistribution,
+                 no_summary, route_map)
```
chrisvanheuveln: This would be easier to process if you passed these vars in as a hash.

lib/cisco_node_utils/router_ospf_area.rb
```diff
@@ -185,6 +187,149 @@ def default_filter_list_out
...
+      if nssa_enable
```
chrisvanheuveln: This is a bit messy and doesn't indicate why it's necessary.

edit: Sorry, this comment was a little half-baked. Take another look at my suggested code:

```
# new yaml tag
nssa_area_destroy:
  set_value: 'no area <area> nssa'
```

```
  def nssa_set(hash)
    # The nssa cli is additive so it must be removed altogether when making changes.
    config_set('ospf_area', 'nssa_destroy', @set_args)) if nssa_enable
    return if hash.empty?

    # Process each nssa property
    @set_args[:state] = ''
    hash.keys.each do |k|
      @set_args[k] = hash[k]
    end
    config_set('ospf_area', 'nssa', @set_args)
  end
```

lib/cisco_node_utils/router_ospf_area.rb
```diff
@@ -185,6 +187,149 @@ def default_filter_list_out
...
+
+    # CLI can be the following or none
+    # area 1.1.1.1 nssa translate type7 always
+    # area 1.1.1.1 nssa translate type7 always supress-fa
+    # area 1.1.1.1 nssa translate type7 never
+    # area 1.1.1.1 nssa translate type7 supress-fa
+    def nssa_translate_type7
+      str = config_get('ospf_area', 'nssa_translate_type7', @get_args)
+      str = 'always_supress_fa' if str == 'always supress-fa'
+      str = 'supress_fa' if str == 'supress-fa'
```
chrisvanheuveln, line 283: Why convert these if you're just converting them back again in the setter?

tests/test_router_ospf_area.rb
```diff
@@ -188,6 +188,273 @@ def test_range
...
+    # "area <area> nssa default-information-originate",
+    # properly if we reset it first. It is only configuring nssa
+    # but not the other parameters. bug ID: CSCva11482
+    ad.nssa_set(true, true,
+                ad.default_nssa_no_redistribution,
+                ad.default_nssa_no_summary,
+                ad.default_nssa_route_map)
+    assert_equal(true, ad.nssa_enable)
+    assert_equal(true, ad.nssa_def_info_originate) unless
+      /N8/ =~ node.product_id
```
chrisvanheuveln: You shouldn't skip this for 8k - if/when they fix it we'll likely forget to update this test. Better to explicitly test for it so that it fails when they fix it; i.e.:

```
  if node.product_id[/N8/]
    refute(ad.nssa_def_info_originate)
  else
    assert(ad.nssa_def_info_originate)
  end
```

tests/test_router_ospf_area.rb
```diff
@@ -210,6 +483,8 @@ def test_destroy
               ['10.3.0.0/32', 'not_advertise'],
               ['10.3.0.1/32'],
               ['10.3.3.0/24', '450']]
+    ad.nssa_set(true, true, true, true, 'aaa')
```
chrisvanheuveln: This is a good example of why I hate positional arguments.

lib/cisco_node_utils/router_ospf_area.rb
saichint, line 283: As discussed, leaving it as it

PR #433, opened 2016-06-20
Added support for retrieving operational data via gRPC/YANG

chburket:
Also added show_running_yang utility to get XR YANG configuration from bash shell.

Currently running rake test, but initial tests look fine.

### DIFF
--- a/docs/README-utilities.md
+++ b/docs/README-utilities.md
@@ -1,0 +1,38 @@
+# Executable utilities included in this gem
+
+## check_metric_limits.rb
+
+This is a helper script for developers. If you're doing refactoring work to
+reduce the code complexity metrics, you can run this script to report the
+worst offenders for each metric and whether you've managed to improve any
+metrics compared to the baseline.  Run this script from the base
+cisco-network-node-utils directory to report metrics of code in the
+./lib and ./tests directories.
+
+```bash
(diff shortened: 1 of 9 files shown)


docs/README-utilities.md
```diff
@@ -0,0 +1,38 @@
+# Executable utilities included in this gem
+
+## check_metric_limits.rb
```
chburket, line 3: Just to be clear, I didn't create the check_metric_limits.rb utility. I just found it in the bin directory and figured I would document it along with our new show_running_yang utility.

lib/cisco_node_utils/node.rb
```diff
@@ -260,7 +265,7 @@ def product_id
           # No support for structured output for this command yet
           output = get(command:     'show inventory',
                        data_format: :cli)
-          return /NAME: "Rack 0".*\nPID: (\S+)/.match(output)[1]
+          return /NAME: .*\nPID: (\S+)/.match(output)[1]
```
chburket, line 268: Made the regex a little less specific, since we had a case where the output did not include the "Rack 0" part.

chburket:
Rake test results are consistent with previous runs.

mikewiebe:
👍 @chburket I did not conduct a deep dive on this one but mostly was looking for impacts to Nexus.  Based on your test results etc the changes look fine to me.

PR #443, opened 2016-07-15
Fix error due to time zone difference

robert-w-gries:
Check for non-nil value instead of verifying exact time


robert-w-gries:
Please hold off on merging for now. I am going to add in similar code to `test_platform.rb`

PR #444, opened 2016-07-15
Fix last_reset check for platform

robert-w-gries:
Follow-up to #443 that fixes `TestPlatform`

### DIFF
--- a/tests/test_platform.rb
+++ b/tests/test_platform.rb
@@ -116,12 +116,11 @@ class TestPlatform < CiscoTestCase
   end
 
   def test_last_reset
-    s = @device.cmd('sh ver').scan(/usecs after\s+(.*)/).flatten.first
     if Utils.nexus_i2_image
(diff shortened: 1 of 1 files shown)


mikewiebe:
Do we need this fix in `develop` also?

robert-w-gries:
Yes.  I assume the plan is to merge `release` branch into `develop` once the release is finished.

mikewiebe:
True, but Chris has been making changes in develop and then cherry picking them into the release branch since it could be quite some time before we merge back into develop.  I would suggest you do the same.

robert-w-gries:
Why don't we just merge `release_1.3.1` branch straight into `develop`? There shouldn't be a need for cherry-picking anything from release -> develop

chrisvanheuveln:
Actually I was doing it the other way - I was working out of develop and then cherry-picking into 1.3.1.

But yeah merge would work too.

PR #445, opened 2016-07-18
Fix suppress-fib-pending lib and test code

robert-w-gries:
## Lib Fixes
- Explicitly exclude `ios_xr` as a platform
- Fix incorrect default value
- Add `auto_default` attribute to handle legacy nxos code that failed to nvgen `suppress-fib-pending`
  - Handle `nil` value for `config_get`
- Fix regexp for property
## Test Fixes
- Latest nxos branch removes `suppress-fib-pending` as a property under vrf contexts. This PR updates test code to handle this change.
- Add `validate_property_excluded` method to check for unsupported platforms
- Remove `test_suppress_fib_pending_not_configured` that was actually testing the property's default value and moved the code to `test_default_suppress_fib_pending` which was essentially performing a no-op in the test.
## Platforms Tested
- n3k - `dublin`
- n5k
- n6k
- n7k
- n8kv
- n9k - `e-dev`
- n9kv - `camden`

### DIFF
--- a/tests/test_router_bgp.rb
+++ b/tests/test_router_bgp.rb
@@ -1167,13 +1167,9 @@ class TestRouterBgp < CiscoTestCase
       bgp.suppress_fib_pending = false
       refute(bgp.suppress_fib_pending,
              'bgp suppress_fib_pending should be disabled')
+      bgp.suppress_fib_pending = bgp.default_suppress_fib_pending
+      assert_equal(bgp.default_suppress_fib_pending, bgp.suppress_fib_pending)
     end
-    bgp.destroy
-  end
-
-  def test_default_suppress_fib_pending
-    bgp = setup_default
-    assert_equal(bgp.default_suppress_fib_pending, bgp.suppress_fib_pending)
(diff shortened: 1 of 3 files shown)


tests/test_router_bgp.rb
```diff
@@ -1155,39 +1155,25 @@ def test_default_shutdown
...
     bgp.destroy
   end
 
   def test_default_suppress_fib_pending
     bgp = setup_default
-    refute(bgp.default_suppress_fib_pending,
-           'bgp suppress_fib_pending default value should be false')
+    assert_equal(bgp.default_suppress_fib_pending, bgp.suppress_fib_pending)
     bgp.destroy
   end
```
chrisvanheuveln: Your changes look fine, but while you're in here can you move this default check (line 1176) into the previous test case (line 1170) and just get rid of this testcase altogether? Every testcase adds another setup/teardown, which drags this test out ever longer...

robert-w-gries: Done.

PR #450, opened 2016-07-26
Refactor aaa local auth removal

mikewiebe:
**Summary of changes:**
- Add yaml entry to indicate which platforms support removal of authentication method `local`
- Update tests to strip `local` from the config cleanup if removal of `local` is not supported.
- Renamed a few variables to make them shorter.

Tested on: n9k and n5k.


robert-w-gries:
👍

PR #451, opened 2016-07-27
New overlay_global behaviors

chrisvanheuveln:
- DME changes for l2rib properties resulted in removal of the 'default' keyword from the cli:
  
  `l2rib dup-host-mac-detection default`
- Prior to this change, 'default' was the only way to reset the host_moves & timeout properties back to default; there was/is no 'no' command, so the only way to reset it now is to re-enter the command with the default values.
- This change necessitates changes to the setter, minitest, and beaker. ~~I also changed the default values to actual integer values instead of empty strings.~~
- Tested minitest & beaker for: n9-edev,n9-I4,n5,n6. Skips for n3 (normal) and n7 (requires F3 card).

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/overlay_global.yaml
+++ b/lib/cisco_node_utils/cmd_ref/overlay_global.yaml
@@ -25,9 +25,6 @@ dup_host_mac_detection:
   get_value: '/^l2rib dup-host-mac-detection (\d+) (\d+)$/'
   set_value: "l2rib dup-host-mac-detection <host_moves> <timeout>"
 
-dup_host_mac_detection_default:
-  set_value: "l2rib dup-host-mac-detection default"
-
 dup_host_mac_detection_host_moves:
   default_value: 5
 
--- a/lib/cisco_node_utils/overlay_global.rb
+++ b/lib/cisco_node_utils/overlay_global.rb
@@ -31,7 +31,6 @@ module Cisco
 
     # dup-host-ip-addr-detection
     def dup_host_ip_addr_detection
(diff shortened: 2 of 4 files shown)


lib/cisco_node_utils/overlay_global.rb
```diff
@@ -31,7 +31,6 @@ class OverlayGlobal < NodeUtil
 
     # dup-host-ip-addr-detection
     def dup_host_ip_addr_detection
-      return nil unless Feature.nv_overlay_evpn_enabled?
```
mikewiebe, line 34: Is it safe to remove this?  I thought this was needed for differences between camden/dublin but I could be mis-remembering.  Should probably test this against I2 and I3 images.

lib/cisco_node_utils/cmd_ref/overlay_global.yaml
```diff
@@ -25,9 +25,6 @@ dup_host_mac_detection:
   get_value: '/^l2rib dup-host-mac-detection (\d+) (\d+)$/'
   set_value: "l2rib dup-host-mac-detection <host_moves> <timeout>"
 
-dup_host_mac_detection_default:
```
mikewiebe, line 28: You indicated in the summary that you changed default values.  Are those changes not in the diffs?

chrisvanheuveln, line 28: Sorry I was misremembering those default changes. They have not changed here.

lib/cisco_node_utils/overlay_global.rb
chrisvanheuveln, line 34: Tests pass on I2 and I3.

PR #457, opened 2016-08-02
Ospf bfd for interfaces

saichint:
This PR is for extending interface parameters to include bfd.
The details are:
1. cisco_interface provider extended to include bfd_echo
2. cisco_interface_ospf provider extended to include bfd and point_to_point network type.
3. cisco_interface_portchannel provider extended to include bfd per-link.
4. cisco_ospf_vrf provider extended to include bfd which enables bfd on all interfaces under this router.

All relevant min tests passed on all nexus platforms.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/interface.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface.yaml
@@ -21,6 +21,13 @@ all_interfaces:
   get_context: ~
   get_value: '/^interface (.*)/'
 
+bfd_echo:
+  _exclude: [ios_xr]
+  kind: boolean
+  get_value: '/^(?:no )?bfd echo$/'
+  set_value: '<state> bfd echo'
+  default_value: true
+
 capabilities:
   _exclude: [ios_xr]
   multiple:
--- a/lib/cisco_node_utils/cmd_ref/interface_ospf.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface_ospf.yaml
@@ -16,7 +16,7 @@ bfd:
   # no config at all so need to grab the optional
   # match to get the whole config for checking
   # the mode
-  get_value: '/^\s*ip ospf bfd(?:\s+\S+)?$/'
+  get_value: '/^\s*ip ospf bfd *(?:\S+)?$/'
   set_value: '%s ip ospf bfd %s'
   default_value: ~
 
@@ -64,11 +64,15 @@ message_digest_password:
   default_value: ~
   get_value: '/^\s*ip ospf message-digest-key \d+ \S+ \d+ (\S+)/'
 
-network_type_p2p:
-  kind: boolean
-  get_value: '/^\s*ip ospf network point-to-point/'
+network_type:
+  get_value: '/^\s*ip ospf network (\S+)$/'
   set_value: "%s ip ospf network %s"
-  default_value: false
+
+network_type_default:
+  default_value: 'broadcast'
+
+network_type_loopback_default:
+  default_value: 'loopback'
 
 passive_interface:
   kind: boolean
--- a/lib/cisco_node_utils/cmd_ref/interface_ospf.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface_ospf.yaml
@@ -10,6 +10,16 @@ area:
   get_value: '/^\s*ip router ospf (\S+) area (\S+)/'
   set_value: "%s ip router ospf %s area %s"
 
+bfd:
+  # the config can be 'ip ospf bfd' or
+  # 'ip ospf bfd disable' or
+  # no config at all so need to grab the optional
+  # match to get the whole config for checking
+  # the mode
+  get_value: '/^\s*ip ospf bfd(?:\s+\S+)?$/'
+  set_value: '%s ip ospf bfd %s'
+  default_value: ~
+
 cost:
   kind: int
   get_value: '/^\s*ip ospf cost (\d+)/'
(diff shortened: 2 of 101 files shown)


lib/cisco_node_utils/cmd_ref/interface_ospf.yaml
```diff
@@ -10,6 +10,16 @@ area:
   get_value: '/^\s*ip router ospf (\S+) area (\S+)/'
   set_value: "%s ip router ospf %s area %s"
 
+bfd:
+  # the config can be 'ip ospf bfd' or
+  # 'ip ospf bfd disable' or
+  # no config at all so need to grab the optional
+  # match to get the whole config for checking
+  # the mode
+  get_value: '/^\s*ip ospf bfd(?:\s+\S+)?$/'
```
chrisvanheuveln: a little cleaner:

`get_value: '/^\s*ip ospf bfd *(?:\S+)?$/'`

lib/cisco_node_utils/interface.rb
```diff
@@ -173,6 +173,23 @@ def default_access_vlan
       config_get_default('interface', 'access_vlan')
     end
 
+    def bfd_echo
+      config_get('interface', 'bfd_echo', name: @name)
+    end
+
+    def bfd_echo=(val)
+      fail ArgumentError, 'Interface Must be physical or port-channel' unless
```
chrisvanheuveln: `/physical/ethernet/`

lib/cisco_node_utils/interface.rb
```diff
@@ -173,6 +173,23 @@ def default_access_vlan
...
+        @name[/eth|port/i]
+      state = (val ? '' : 'no')
+      Feature.bfd_enable
```
chrisvanheuveln, line 187: Seems like this should have a check here ( `return if state == 'no' and not Feature.bfd_enabled` ) to avoid enabling bfd when the manifest asks for 'default'.

edit: need to check all properties that enable bfd

lib/cisco_node_utils/cmd_ref/interface.yaml
```diff
@@ -21,6 +21,13 @@ all_interfaces:
   get_context: ~
   get_value: '/^interface (.*)/'
 
+bfd_echo:
+  _exclude: [ios_xr]
+  kind: boolean
+  get_value: '/^(?:no )?bfd echo$/'
+  set_value: '<state> bfd echo'
+  default_value: true
```
chrisvanheuveln, line 29: This one is tricky, I think. Is the default state always true, or just when feature bfd is enabled?

lib/cisco_node_utils/interface_ospf.rb
```diff
@@ -214,6 +216,46 @@ def dead_interval=(interval)
...
+    def bfd
+      val = config_get('interface_ospf', 'bfd', @interface.name)
+      return default_bfd if val == default_bfd
+      val.include?('disable') ? false : true
+    end
+
+    # interface %s
+    #   %s ip ospf bfd %s
+    def bfd=(val)
+      Feature.bfd_enable
```
chrisvanheuveln, line 232: `destroy` will call this setter regardless. This also needs a check for `'no' && not Feature.bfd_enabled`

lib/cisco_node_utils/interface_ospf.rb
```diff
@@ -214,6 +216,46 @@ def dead_interval=(interval)
...
+    def default_bfd
+      config_get_default('interface_ospf', 'bfd')
+    end
+
+    # CLI can be either of the following or none
+    # ip ospf bfd
+    # ip ospf bfd disable
+    def bfd
+      val = config_get('interface_ospf', 'bfd', @interface.name)
+      return default_bfd if val == default_bfd
```
chrisvanheuveln: This is really just a nil check. 

```
   def bfd
     val = config_get('interface_ospf', 'bfd', @interface.name)
     return if val.nil?
     val.include?('disable') ? false : true
end
```

lib/cisco_node_utils/interface_ospf.rb
```diff
@@ -214,6 +216,46 @@ def dead_interval=(interval)
                  @interface.name, '', interval.to_i)
     end
 
+    def default_bfd
```
chrisvanheuveln: Keep the properties in alpha order.

lib/cisco_node_utils/cmd_ref/interface_ospf.yaml
```diff
@@ -54,6 +64,12 @@ message_digest_password:
   default_value: ~
   get_value: '/^\s*ip ospf message-digest-key \d+ \S+ \d+ (\S+)/'
 
+network_type_p2p:
```
chrisvanheuveln: Why `network_type_p2p` instead of just `network`?  There are four different network types. I'd prefer to see this as a string, a la:

```
network_type:
  get_value: '/^ip ospf network (\S+)$/'
  set_value: "%s ip ospf network %s"
  default_value: 'broadcast'
```

lib/cisco_node_utils/cmd_ref/interface.yaml
mikewiebe, line 29: @chrisvanheuveln Not I understand the concern.  Wouldn't the same also apply for all BGP properties if `feature bgp` is enabled/disabled?

saichint, line 29: For now leaving it as it is. When developing the puppet provider code, will revisit this to see if there is a cleaner solution.

saichint:
All comments taken care except one which will be revisited during puppet dev

lib/cisco_node_utils/cmd_ref/interface.yaml
chrisvanheuveln, line 29: This looks okay. Code the getter/default_getter like interface switchport_vtp and it should work fine.

saichint, line 29: Done

PR #460, opened 2016-08-08
Ospf parameters for interfaces

saichint:
This PR is for adding four missing properties to the interface_ospf.
1. mtu_ignore
2. priority
3. shutdown
4. transmit_delay

All relevant tests are run on all nxos platforms including evergreen and they pass.

### DIFF
--- a/lib/cisco_node_utils/interface_ospf.rb
+++ b/lib/cisco_node_utils/interface_ospf.rb
@@ -92,6 +92,8 @@ module Cisco
       self.hello_interval = default_hello_interval
       config_set('interface_ospf', 'dead_interval',
                  @interface.name, 'no', '')
+      self.bfd = default_bfd
+      self.network_type = default_network_type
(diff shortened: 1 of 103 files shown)


lib/cisco_node_utils/interface_ospf.rb
```diff
@@ -250,6 +254,21 @@ def default_network_type
       config_get_default('interface_ospf', lookup)
     end
 
+    def mtu_ignore
+      config_get('interface_ospf', 'mtu_ignore', @interface.name)
+    end
+
+    # interface %s
```
mikewiebe, line 261: Why are these comments necessary?

saichint, line 261: They are not necessary. But the set methods in this ruby file had these comments like this before and I added for the new ones as well for consistency. I can get rid of them if you want but if we do that, let us do for the entire file (unless there is a special need).

PR #462, opened 2016-08-10
Minitest bug fixes

saichint:
This PR is for fixing the following minitests.
1. test_node_ext.rb --> test_get_boot --> the issue here is that the yaml file is getting isan_file_name for boot image on n/6/7k which is incorrect. On all platforms, kick_file_name returns boot image.
1. test_platform.rm --> test_system_image --> there are mutilple issues here. One is that the Platform.system_image itself is wrong. It is getting boot image instead of system image from the yaml. This works ok on n3/8/9k as there is only one image file, but on n5/6/7k, it is totally wrong as it is is showing boot image for system image. It is very surprising that this was never caught before (unless it is changed recently) as the return value is completely wrong. There is also another problem in the test file, where is is comparing the system image with "sh ver | i image" string. Again this works ok on n3/8/9k as there is only one image file but on others this returns both kickstart and system and it is comparing the first line and it is failing. Added an exclude as well to get rid of kickstart (exclude kickstart will have no effect on n3/8/9k) to take care of all platforms.
2. test_stp_global.rb --> bd related tests --> bd and vlan are mutually exclusive and this is taken care by removing any bds during setup. But the other way is not taken care. So if there are vlans left by previous tests, bd tests will fail. Added appropriate cleanups (Thanks Chris).
3. Removed test_last_reset from test_platform.rb as it has no added value as suggested by Mike

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ pkg/*
 vendor/*
 coverage
 ext/Rakefile
+tests/tacacs_server.yaml
 Gemfile.lock
 *.gem
 *.swp
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Changelog
   * bfd (@saichint)
 * OSPF
   * ospf_area (@saichint)
+  * ospf_area_vlink (@saichint)
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ Changelog
 
 * Extend bgp_neighbor with attributes:
    * `bfd`
+* Added client support for gRPC on IOS XR.
+* Smart dependency installation - installing this gem will install `grpc` on IOS XR and Linux environments, but not on NX-OS environments.
+
 
 ## [v1.3.1]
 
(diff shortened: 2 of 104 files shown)


chrisvanheuveln:
STP / Bridge-domain fixes look good.

@mikewiebe , @robert-w-gries Take a look at the boot_image / system_image changes. It seems like these have been changed back and forth a couple of times already. Rob, did you work on this once before?

mikewiebe:
@chrisvanheuveln I took a look but it appears you made some changes to this area most recently.

https://github.com/saichint/cisco-network-node-utils/commit/536b4c1f7b6a9687d26776a43f7e877d38baeaeb

chrisvanheuveln:
For system_image & boot_image I see very different results with n3k, n5k, n9k. For that matter what do these two actually mean? :-( 
These structured properties seem to change every new release so I'm of a mind to just replace these with a basic screen scrape instead. Let's discuss at next scrum.

robert-w-gries:
@chrisvanheuveln I haven't touched the `test_platform` code that Sai is changing AFAIK

PR #466, opened 2016-08-15
Fix half/full in test_duplex

chrisvanheuveln:
- For some 3k, need to munge 'half/full' into csv-style 'half,full' to play nice with interface_capabilities which needs to parse these as separate options
- Tested on n3048

### DIFF
--- a/lib/cisco_node_utils/interface.rb
+++ b/lib/cisco_node_utils/interface.rb
@@ -122,6 +122,7 @@ module Cisco
           next if k.nil? || v.nil?
           k.gsub!(/ \(.*\)/, '') # Remove any parenthetical text from key
           v.strip!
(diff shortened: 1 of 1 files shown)


lib/cisco_node_utils/interface.rb
```diff
@@ -122,6 +122,7 @@ def self.capabilities(intf, mode=:hash)
           next if k.nil? || v.nil?
           k.gsub!(/ \(.*\)/, '') # Remove any parenthetical text from key
           v.strip!
+          v.gsub!(%r{half/full}, 'half,full') if k == 'Duplex'
```
mikewiebe, line 125: Should we test this on multiple n3k flavors just to be safe?

PR #467, opened 2016-08-16
minitest fixes for vtp and resilient mode

saichint:
This PR is for fixing 2 problems in the test_vtp and resilient mode on n3k.
1. Setting the vtp domain twice. This fails on n3k, n9k but on n7k, this does not throw any error. Since we are setting the same value to the same property, this is not a major issue and this particular test has no added value. After some discussion, decided to remove the test.
2. If the vtp password is being set with special characters including backslashes, the password seen from nxapi on n7k is different from what is being set. This is fixed on n7k post helsinki release. Added comments to the test code that this test will fail on n7k running old images.
3. On Some n3k platforms, resilient mode is not supported for port-channel global provider. The fix is to check if throws an exception with proper strings for non support of the feature and return normally in these cases.

### DIFF
--- a/tests/ciscotest.rb
+++ b/tests/ciscotest.rb
@@ -288,9 +288,10 @@ class CiscoTestCase < TestCase
     #   '9  12  10/40 Gbps Ethernet Module  N77-F312FQ-25 ok'
     #   '2   6  Nexus 6xQSFP Ethernet Module  N5K-C5672UP-M6Q ok'
     #   '2   6  Nexus xxQSFP Ethernet Module  N6K-C6004-96Q/EF ok'
+    #   '2   4  Nexus 4xQSFP Ethernet Module  N6K-C6001-M4Q ok'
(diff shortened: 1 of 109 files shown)


tests/test_vtp.rb
```diff
@@ -176,6 +174,9 @@ def test_password_valid
     end
   end
 
+  # this fails on n7k running helsinki image but it
```
mikewiebe: Please use the version number `7.3(0)D1(1)` instead of the internal name `helsinki`

tests/ciscotest.rb
```diff
@@ -108,6 +108,14 @@ def config_and_warn_on_match(warn_match, *args)
     result
   end
 
+  # this method takes the CliError exception and either
```
mikewiebe: **Nit** Simplify comment block to read:

```
# Check exception and only fail if it does not contain message.
```

PR #468, opened 2016-08-17
fix for symmetry not supported on n3k

saichint:
On some n3k flavors, symmetry property is unsupported. This PR is to take care of the minitest, which returns success if this is unsupported. Added a generic test case to run on all n3k flavors for other cases.
minitest passes on all nxos platforms

### DIFF
--- a/tests/test_portchannel_global.rb
+++ b/tests/test_portchannel_global.rb
@@ -113,6 +113,8 @@ class TestPortchannelGlobal < CiscoTestCase
       global.resilient = global.default_resilient
       assert_equal(global.default_resilient, global.resilient)
     end
+  rescue Cisco::CliError => e
+    check_and_raise_error(e, 'This feature is not supported on this platform')
   end
 
   def test_load_balance_no_rotate
--- a/tests/test_portchannel_global.rb
+++ b/tests/test_portchannel_global.rb
@@ -117,6 +117,37 @@ class TestPortchannelGlobal < CiscoTestCase
     check_and_raise_error(e, 'This feature is not supported on this platform')
   end
 
+  def test_load_balance_no_rotate_no_symmetry
+    skip('Test not supported on this platform') unless n3k_in_n3k_mode?
+
+    global = create_portchannel_global
+    global.send(:port_channel_load_balance=,
+                'src-dst', 'ip-only', nil, nil, false, nil, nil)
+    assert_equal('src-dst',
+                 global.bundle_select)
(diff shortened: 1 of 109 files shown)


tests/test_portchannel_global.rb
```diff
@@ -146,6 +177,8 @@ def test_load_balance_no_rotate
     assert_equal(
       global.default_symmetry,
       global.symmetry)
+  rescue Cisco::CliError => e
+    check_and_raise_error(e, 'This command is only supported on T2 platforms.')
```
chrisvanheuveln: I don't think raising an error here is appropriate. You really only want this test to run when:

1) It's a T2 (i.e. 31xx)
2) It's running in 3k switch-mode

Which means you need a T2 helper method and then at line 152 do:

```
     skip('Test not supported on this platform') unless n3k_is_t2 && n3k_in_n3k_mode?
```

For the t2 helper you can either create a method in ciscotest to scape it from show ver, or add a new property in `show_version.yaml`:

```
chassis_id:
  get_value: "chassis_id"       # Nexus 3172 Chassis
```

I prefer the second option since I think we'll need this for other properties eventually, and it could be used by beaker as well. But right now it's just affecting some tests so a simple scraper method is fine too.

tests/test_portchannel_global.rb
```diff
@@ -37,6 +38,11 @@ def cleanup
     config_no_warn "no port-channel load-balance #{ethernet}"
   end
 
+  def n3k_is_t2?
+    return unless /N3/ =~ node.product_id
+    Platform.chassis['descr'].include?('Nexus 31') ? true : false
```
chrisvanheuveln: `'descr'` is a string. `include?` is for arrays. 

Also, the 31xx's use a string like `Nexus3000 C3132Q Chassis`

saichint: No. include? is in the strings as well. I used it before many times.
check this:
https://ruby-doc.org/core-2.2.0/String.html#method-i-include-3F.

But I did not know about 31xx being Nexus3000 C3132Q. 
Any suggestions on what to look for, to determinte if it is T2 chassis?

chrisvanheuveln: Ah, nevermind then.

Looks like 31xx displays a little different for 3132 vs 3172. You can probably match against something like this:

`[/Nexus3000 C31|Nexus 31/]

Validate your tests against n31x-101, n3k-161, and n3k-103.

PR #473, opened 2016-08-29
Rel140/refactor yum

mikewiebe:
**Summary of changes:**
- Our yum provider would only deactivate a patch when calling `Yum.remove`.  This update calls both `install deactivate <pkg_name>` and `install remove <pkg_name> forced` so that the package is really removed when `Yum.remove` is called.
  - This was causing an install failure when attempting to install the same pkg_name(built for different image versions) because the previous package was only getting deactivated, not actually removed.
- Refactor to install/query/and remove in the same test for efficiency.
- Enable additional nxos releases.

Tested on: n9k (I2, I3, I4, I5), n3k (I5).
Also Tested with beaker using puppet `develop` branch: n9k (I2, I3, I4, I5) (Native and Guestshell)

### DIFF
--- a/lib/cisco_node_utils/yum.rb
+++ b/lib/cisco_node_utils/yum.rb
@@ -71,6 +71,7 @@ module Cisco
     end
 
     def self.remove(pkg)
+      config_set('yum', 'deactivate', pkg)
       config_set('yum', 'remove', pkg)
     end
   end
--- a/lib/cisco_node_utils/yum.rb
+++ b/lib/cisco_node_utils/yum.rb
@@ -75,7 +75,7 @@ module Cisco
       # May not be able to remove the package immediately after
       # deactivation.
       try = 1
-      while try < 50
+      while try < 20
         o = config_set('yum', 'remove', pkg)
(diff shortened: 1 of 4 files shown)


lib/cisco_node_utils/yum.rb
```diff
@@ -71,7 +71,15 @@ def self.query(pkg)
...
-      config_set('yum', 'remove', pkg)
+      config_set('yum', 'deactivate', pkg)
+      # May not be able to remove the package immediately after
+      # deactivation.
+      try = 1
+      while try < 20
+        o = config_set('yum', 'remove', pkg)
+        break unless o[/operation is in progress, please try again later/]
+        sleep 1
+      end
```
chrisvanheuveln, line 82: infinite loop

Use this:

```
- try = 1
- while try < 20
+ while (try ||= 1) < 20
  ...
+ try += 1
end
```

mikewiebe: Good call.

chrisvanheuveln: Yes but would you say 'infinitely' good?

PR #474, opened 2016-08-31
Add storm_control and dhcp relay props for interface

saichint:
This PR is for adding storm control and dhcp relay properties to interface provider.
All these properties are excluded for ios_xr for now. If they are needed for ios_xr, separate PR needs to raised as it is tested only on nexus platforms.
There are 3 storm control properties (one each for broadcast, multicast and unicast). Each of them can be set separately for all nexus platforms except n7k. On n7k, if any of the properties is set, all the other 2 also are set to the same value. This is due to hardware limitation and cannot be fixed. So only unicast support is added for n7k and the other two are excluded.
All tests pass on n3k, n6k, n7k and n9k (using CI).


lib/cisco_node_utils/interface.rb
```diff
@@ -420,6 +420,108 @@ def default_ipv4_arp_timeout
...
+      Feature.dhcp_enable if state
+      config_set('interface', 'ipv4_dhcp_relay_src_addr_hsrp',
+                 name:  @name, state: state ? '' : 'no')
+    end
+
+    def default_ipv4_dhcp_relay_src_addr_hsrp
+      config_get_default('interface', 'ipv4_dhcp_relay_src_addr_hsrp')
+    end
+
+    # manifest will have 'ethernet 1/1', port-channel 100', vlan 2' etc.
```
mikewiebe: *nit* Simply comment to:
```
Normalize output by removing white space and downcasing interface names
```

lib/cisco_node_utils/interface.rb
```diff
@@ -420,6 +420,108 @@ def default_ipv4_arp_timeout
...
+    # The show cmd returns 'Ethernet1/1', 'port-channel100', Vlan2, etc.
+    # so make this idempotent, normalize the output to match the input.
+    def ipv4_dhcp_relay_src_intf
```
mikewiebe, line 475: If I understand correctly you are trying to downcase and insert whitespace?  I suggest you just normalize by downcasing and removing whitespace.

```
irb(main):012:0* intf = "Ethernet 1/1"
=> "Ethernet 1/1"
irb(main):013:0> intf.downcase.delete(' ')
=> "ethernet1/1"
irb(main):014:0> intf = "Ethernet1/1"
=> "Ethernet1/1"
irb(main):015:0> intf.downcase.delete(' ')
=> "ethernet1/1"
irb(main):016:0> intf = "ethernet 1/1"
=> "ethernet 1/1"
irb(main):017:0> intf.downcase.delete(' ')
=> "ethernet1/1"
irb(main):018:0> 
```

You can do this here and munge it in the type file so the manifest can specify the name in any of these formats.

lib/cisco_node_utils/interface.rb
```diff
@@ -526,6 +628,52 @@ def default_ipv6_acl_out
...
+    end
+
+    def default_ipv6_dhcp_relay_addr
+      config_get_default('interface', 'ipv6_dhcp_relay_addr')
+    end
+
+    # manifest will have 'ethernet 1/1', port-channel 100', vlan 2' etc.
+    # The show cmd returns 'Ethernet1/1', 'port-channel100', Vlan2, etc.
+    # so make this idempotent, normalize the output to match the input.
+    def ipv6_dhcp_relay_src_intf
```
mikewiebe, line 652: Same comment here and I would also just add a simplified inline comment:

```
...
  # Normalize by downcasing and removing white space
  inf.downcase.delete(' ')
```

tests/test_interface.rb
```diff
@@ -1485,4 +1485,237 @@ def test_ipv4_pim_sparse_mode
     i.ipv4_pim_sparse_mode = i.default_ipv4_pim_sparse_mode
     assert_equal(i.default_ipv4_pim_sparse_mode, i.ipv4_pim_sparse_mode)
   end
+
+  def test_storm_control_broadcast
+    interface = Interface.new(interfaces[0])
+    interface.switchport_enable(true)
+    if validate_property_excluded?('interface', 'storm_control_broadcast')
+      assert_nil(interface.storm_control_broadcast)
+      assert_nil(interface.default_storm_control_broadcast)
```
mikewiebe: IMO we only need one `asser_nil` in each of these tests.  I would just assert on the getter and remove the default getter.

tests/test_interface.rb
```diff
@@ -1485,4 +1485,237 @@ def test_ipv4_pim_sparse_mode
...
+  end
+
+  def test_ipv6_dhcp_relay_src_intf
+    interface = Interface.new(interfaces[0])
+    interface.switchport_enable(false)
+    if validate_property_excluded?('interface', 'ipv6_dhcp_relay_src_intf')
+      assert_nil(interface.ipv6_dhcp_relay_src_intf)
+      assert_nil(interface.default_ipv6_dhcp_relay_src_intf)
+      assert_raises(Cisco::UnsupportedError) do
+        interface.ipv6_dhcp_relay_src_intf = 'ethernet 2/1'
```
mikewiebe: Don't hardcode interface names.  Instead at the beginning of this method do this:

```
def test_ipv6_dhcp_relay_src_intf
    inf_name = interfaces[0]
    interface = Interface.new(inf_name)
```

Then use `intf_name`

tests/test_interface.rb
```diff
@@ -1485,4 +1485,227 @@ def test_ipv4_pim_sparse_mode
     i.ipv4_pim_sparse_mode = i.default_ipv4_pim_sparse_mode
     assert_equal(i.default_ipv4_pim_sparse_mode, i.ipv4_pim_sparse_mode)
   end
+
+  def test_storm_control_broadcast
```
mikewiebe, line 1497: So this is a bit beyond the scope of this PR but I think it's small enough and will add enough value that we should go ahead and do it.

The tests in this file enable quite a few features but never turn them off which can cause problems for other tests.  Chris added some logic a few days (`first_or_last_teardown`) ago that allows you to determine if the current test is the last test case being run. https://github.com/cisco/cisco-network-node-utils/commit/9f13ed247b75b2561b1b91daad5ab92f59edd566

I suggest you do the following.
- Turn of all features on the box (n3k, n5/6k, n7k, n9k) and run the tests in this file to see all of the features that are still enabled when the test complete.
- Create a method to disable all of the features and call it in the `teardown` method if it's the last test being run.

saichint: Added

PR #475, opened 2016-09-01
snmpuser minitest fix

saichint:
This PR is for removing the test which checks if we can create a user with less than 8 char password. This test no longer is valid because the platform code is now designed to accept smaller length passwords.


mikewiebe:
wow.. there are 30 commits in your history :).  Might be better to re-open the PR?

PR #477, opened 2016-09-02
Add defect/image check to minitest

chrisvanheuveln:
- Bails out of testcase if image matches pattern; no skip, no assert
- Prints 'NOOP' message and any add'l helpful text, e.g.:

```
Node under test:
  - name  - n5k-136
  - type  - N5K-C56128P
  - image - bootflash:///n6000-uk9.7.3.0.N1.1.bin

TestVrf#test_name_zero_length = 4.88 s = .
TestVrf#test_route_distinguisher = 11.79 s = .
TestVrf#test_remote_route_filtering = 2.01 s = .
TestVrf#test_name_too_long = 1.93 s = .
TestVrf#test_vpn_id = 2.07 s = .
TestVrf#test_vni = 2.85 s = S
TestVrf#test_mhost = 1.96 s = .

TestVrf#test_description -> NOOP <- [CSCuy36637: Cannot remove description]
TestVrf#test_description = 2.91 s = .
TestVrf#test_collection_default = 1.01 s = .
TestVrf#test_create_and_destroy = 2.89 s = .
TestVrf#test_name_type_invalid = 1.05 s = .
TestVrf#test_shutdown = 26.73 s = .

Finished in 62.072465s, 0.1933 runs/s, 0.4189 assertions/s.
12 runs, 26 assertions, 0 failures, 0 errors, 1 skips
```

### DIFF
--- a/tests/ciscotest.rb
+++ b/tests/ciscotest.rb
@@ -183,7 +183,7 @@ class CiscoTestCase < TestCase
 
   def defect?(pattern, msg)
     return false unless system_image.match(Regexp.new(pattern))
-    msg = "#{self.class}##{caller[0][/`.*'/][1..-2]} -> NOOP <- [#{msg}]"
+    msg = "#{self.class}##{caller[0][/`.*'/][1..-2]} -> NOT TESTED <- [#{msg}]"
     Cisco::Logger.warn(msg)
     true
   end
--- a/tests/ciscotest.rb
+++ b/tests/ciscotest.rb
@@ -16,10 +16,11 @@ require 'ipaddr'
 require 'resolv'
 require_relative 'basetest'
 require_relative 'platform_info'
+require_relative '../lib/cisco_node_utils/bridge_domain'
 require_relative '../lib/cisco_node_utils/interface'
 require_relative '../lib/cisco_node_utils/node'
+require_relative '../lib/cisco_node_utils/platform'
 require_relative '../lib/cisco_node_utils/vlan'
-require_relative '../lib/cisco_node_utils/bridge_domain'
 
 include Cisco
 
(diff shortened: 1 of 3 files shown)


tests/ciscotest.rb
```diff
@@ -176,6 +177,16 @@ def skip_nexus_i2_image?
       Utils.nexus_i2_image
   end
 
+  def system_image
+    @image ||= Platform.system_image
+  end
+
+  def defect?(pattern, msg)
+    return false unless system_image.match(Regexp.new(pattern))
+    puts "\n#{self.class}##{caller[0][/`.*'/][1..-2]} -> NOOP <- [#{msg}]"
```
robert-w-gries: Messages should be using the logger class instead of `puts`.

Nit: I'm not a fan of printing `NOOP` or that the test is marked as passing.  During our offline discussion, did we decide on asserting defects or skipping the testcase?

ping @mikewiebe

chrisvanheuveln: The logger message is not quite as pretty but okay. :-)
The NOOP/pass message was an ask from Mike. We agreed to pass on "wont-fix" bugs only. Defects that have a chance of being fixed should still fail.

Here's the new output using logger:

```
Node under test:
  - name  - n6k-77
  - type  - N6K-C6001-64P
  - image - bootflash:///n6000-uk9.7.3.0.N1.1.bin

W, [2016-09-06T08:23:07.298520 #7877]  WARN -- : TestVrf#test_description -> NOOP <- [CSCuy36637: Cannot remove description]
TestVrf#test_description = 4.77 s = .
```

chrisvanheuveln: Here's another idea. I removed the `W, [2016-09-06T08:23:07.298520 #7877]` from the log output and changed `NOOP` to `NOT TESTED`. Is this better?

```
WARN -- : TestVrf#test_description -> NOT TESTED <- [CSCuy36637: Cannot remove description]
TestVrf#test_description = 5.53 s = .
```

mikewiebe:
I believe we did decide to assert for defects in specific images where we know the test will fail because of the defect.  We also decided that we would make sure that any tests that are skipped are only skipped for valid reasons so that we can `trust` skips like we `trust` pass results.

mikewiebe:
👍

PR #480, opened 2016-09-09
(POC) product_id qualifier for 8k rebrand

chrisvanheuveln:
- This is just a POC diff for adding add'l checks to determine the product_id. 

**DO NOT MERGE THIS CODE.**
- The requirement here is to create a unique product_id for switches that share the same product_id but have a fundamental difference; e.g.
  - 8k may only be differentiated by the presense of 8k fabric modules
  - n3k-in-n9k-mode will have the same product_id as a "regular" n3k


chrisvanheuveln:
This is a POC-only commit that we'll keep in a feature branch for now. Do not merge into develop at this time.

PR #481, opened 2016-09-09
Dhcp Relay Global

saichint:
This PR is for adding new provider for dhcp_relay_global.
Tests passed on all nexus platforms (including CI).

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/dhcp_relay_global.yaml
+++ b/lib/cisco_node_utils/cmd_ref/dhcp_relay_global.yaml
@@ -36,6 +36,11 @@ ipv4_relay:
   get_value: '/^ip dhcp relay$/'
   set_value: "<state> ip dhcp relay"
   N3k: &relay_default_true
+    # auto_default is set to false because when
+    # this property is set to false, we do not
+    # find the regex pattern and since the default_value
+    # is true, it should not return true as its real
+    # value is false.
     auto_default: false
     default_value: true
   N5k: &relay_default_false
@@ -100,6 +105,11 @@ ipv6_relay:
   get_value: '/^ipv6 dhcp relay$/'
   set_value: "<state> ipv6 dhcp relay"
   N3k: &v6_relay_default_true
+    # auto_default is set to false because when
+    # this property is set to false, we do not
+    # find the regex pattern and since the default_value
+    # is true, it should not return true as its real
+    # value is false.
     auto_default: false
     default_value: true
   N5k: &v6_relay_default_false
--- a/lib/cisco_node_utils/cmd_ref/dhcp_relay_global.yaml
+++ b/lib/cisco_node_utils/cmd_ref/dhcp_relay_global.yaml
@@ -5,33 +5,33 @@ _exclude: [ios_xr]
 _template:
   get_command: "show running-config dhcp | no-more"
 
-information_option:
+ipv4_information_option:
   kind: boolean
   get_value: '/^ip dhcp relay information option$/'
   set_value: "<state> ip dhcp relay information option"
   default_value: false
 
-information_option_trust:
+ipv4_information_option_trust:
   _exclude: [N5k, N6k]
   kind: boolean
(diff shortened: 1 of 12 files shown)


lib/cisco_node_utils/dhcp_relay_global.rb
```diff
@@ -0,0 +1,304 @@
...
+
+module Cisco
+  # node_utils class for dhcp_relay_global
+  class DhcpRelayGlobal < NodeUtil
+    attr_reader :name
+
+    def initialize(name)
+      fail TypeError unless name.is_a?(String)
+      fail ArgumentError unless name == 'default'
+      @name = name.downcase
```
mikewiebe: Why do you need to downcase name here?  You generate an `ArgumentError` above if the name does not equal 'default'.

**_UPDATE**_ Thinking about this some more, we don't even really need a name at all.  I would treat this like vtp where initialize just enables the feature if `instantiate=true`.

https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/vtp.rb#L29

lib/cisco_node_utils/cmd_ref/dhcp_relay_global.yaml
```diff
@@ -0,0 +1,115 @@
...
+  get_value: '/^ip dhcp relay information trust-all$/'
+  set_value: "<state> ip dhcp relay information trust-all"
+  default_value: false
+
+ipv4_relay:
+  kind: boolean
+  get_value: '/^ip dhcp relay$/'
+  set_value: "<state> ip dhcp relay"
+  N3k: &relay_default_true
+    auto_default: false
```
mikewiebe, line 44: Why do you need to turn off auto_default here?

tests/test_dhcp_relay_global.rb
```diff
@@ -0,0 +1,279 @@
...
+# of the DhcpRelayGlobal class.
+class TestDhcpRelayGlobal < CiscoTestCase
+  @skip_unless_supported = 'dhcp_relay_global'
+  DEFAULT_NAME = 'default'
+
+  # Tests
+
+  def setup
+    super
+    config_no_warn 'no feature dhcp'
```
mikewiebe: `config_no_warn('no feature dhcp')`

tests/test_dhcp_relay_global.rb
```diff
@@ -0,0 +1,279 @@
...
+    end
+    assert_equal(drg.default_ipv4_src_intf, drg.ipv4_src_intf)
+    drg.ipv4_src_intf = 'port-channel200'
+    assert_equal('port-channel200', drg.ipv4_src_intf)
+    drg.ipv4_src_intf = drg.default_ipv4_src_intf
+    assert_equal(drg.default_ipv4_src_intf, drg.ipv4_src_intf)
+  end
+
+  def test_ipv4_sub_option_circuit_id_custom
+    skip_nexus_i2_image?
```
mikewiebe, line 181: Is this property broken or unsupported on i2 images?

tests/test_dhcp_relay_global.rb
```diff
@@ -26,16 +26,16 @@ class TestDhcpRelayGlobal < CiscoTestCase
...
 
   def teardown
     config_no_warn('no feature dhcp') if first_or_last_teardown
     super
   end
 
-  def create_dhcp_relay_global(name=DEFAULT_NAME)
-    DhcpRelayGlobal.new(name)
+  def create_dhcp_relay_global
+    DhcpRelayGlobal.new(true)
```
mikewiebe: Simplify `DhcpRelayGlobal.new`

tests/test_dhcp_relay_global.rb
saichint: yes, this is unsupported only on i2. Just tested it

saichint:
All minitests passed on all nxos platforms. Please merge if there are no more comments.

PR #484, opened 2016-09-21
bgp event_history

saichint:
This PR is for fixing router bgp event-history CLIs. In the latest images of N3k and N9k, all these CLIs accept size in bytes and also some of the defaults and other behavior is changed. The code changes are done in such a way that, for the older releases, it works as it was before and new code runs for the latest releases based on the image version and node types.

### DIFF
--- a/lib/cisco_node_utils/bgp.rb
+++ b/lib/cisco_node_utils/bgp.rb
@@ -416,24 +416,16 @@ module Cisco
       match = config_get('bgp', 'event_history_cli', @get_args)
       if match.is_a?(Array)
         return 'false' if match[0] == 'no '
-        if Utils.nexus_evergreen?
-          return default_event_history_cli unless match[1]
-          # check if the val is size in bytes
-          return match[1] if match[1] =~ /\A\d+\z/
+        if match[1]
+          return match[1] if match[1][/\A\d+\z/]
           return 'size_' + match[1]
-        else
-          return 'size_' + match[1] if match[1]
         end
       end
       default_event_history_cli
     end
 
     def event_history_cli=(val)
-      size = val[/small|medium|large|disable/]
-      if Utils.nexus_evergreen?
-        # check if the val is size in bytes
-        size = val if val =~ /\A\d+\z/
-      end
+      size = val[/small|medium|large|disable|\A\d+\z/]
       @set_args[:size] = size.nil? ? '' : "size #{size}"
       @set_args[:state] = val[/false/] ? 'no' : ''
       config_set('bgp', 'event_history_cli', @set_args)
@@ -451,24 +443,16 @@ module Cisco
       # This property requires auto_default=false
       if match.is_a?(Array)
         return 'false' if match[0] == 'no '
-        if Utils.nexus_evergreen?
-          return default_event_history_detail unless match[1]
-          # check if the val is size in bytes
-          return match[1] if match[1] =~ /\A\d+\z/
+        if match[1]
+          return match[1] if match[1][/\A\d+\z/]
           return 'size_' + match[1]
-        else
-          return 'size_' + match[1] if match[1]
         end
       end
       default_event_history_detail
     end
 
     def event_history_detail=(val)
-      size = val[/small|medium|large|disable/]
-      if Utils.nexus_evergreen?
-        # check if the val is size in bytes
-        size = val if val =~ /\A\d+\z/
-      end
+      size = val[/small|medium|large|disable|\A\d+\z/]
       @set_args[:size] = size.nil? ? '' : "size #{size}"
       @set_args[:state] = val[/false/] ? 'no' : ''
       config_set('bgp', 'event_history_detail', @set_args)
@@ -485,24 +469,16 @@ module Cisco
       match = config_get('bgp', 'event_history_events', @get_args)
       if match.is_a?(Array)
         return 'size_disable' if match[0] == 'no '
-        if Utils.nexus_evergreen?
-          return default_event_history_events unless match[1]
-          # check if the val is size in bytes
-          return match[1] if match[1] =~ /\A\d+\z/
+        if match[1]
+          return match[1] if match[1][/\A\d+\z/]
           return 'size_' + match[1]
-        else
-          return 'size_' + match[1] if match[1]
         end
       end
       default_event_history_events
     end
 
     def event_history_events=(val)
-      size = val[/small|medium|large|disable/]
-      if Utils.nexus_evergreen?
-        # check if the val is size in bytes
-        size = val if val =~ /\A\d+\z/
-      end
+      size = val[/small|medium|large|disable|\A\d+\z/]
       @set_args[:size] = size.nil? ? '' : "size #{size}"
       @set_args[:state] = val[/false/] ? 'no' : ''
       config_set('bgp', 'event_history_events', @set_args)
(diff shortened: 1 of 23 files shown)


lib/cisco_node_utils/bgp.rb
```diff
@@ -416,13 +416,24 @@ def event_history_cli
       match = config_get('bgp', 'event_history_cli', @get_args)
       if match.is_a?(Array)
         return 'false' if match[0] == 'no '
-        return 'size_' + match[1] if match[1]
```
mikewiebe, line 419: I believe we can simplify and remove the `Utils.nexus_evergreen?` check for `event_history_cli` like this:

``` ruby
def event_history_cli
  if match.is_a?(Array)
    return 'false' if match[0] == 'no '
    if match[1]
      return match[1] if match[1][/\A\d+\z/]
      return 'size_' + match[1]
    end
  end
  default_event_history_cli
end
```

lib/cisco_node_utils/bgp.rb
```diff
@@ -416,13 +416,24 @@ def event_history_cli
...
+        else
+          return 'size_' + match[1] if match[1]
+        end
       end
       default_event_history_cli
     end
 
     def event_history_cli=(val)
       size = val[/small|medium|large|disable/]
+      if Utils.nexus_evergreen?
```
mikewiebe: Don't think the `Utils.nexus_evergreen?` check is needed here either.  This can simply be

```
size = val[/small|medium|large|disable|\A\d+\z/]
```

lib/cisco_node_utils/bgp.rb
```diff
@@ -440,13 +451,24 @@ def event_history_detail
       # This property requires auto_default=false
       if match.is_a?(Array)
         return 'false' if match[0] == 'no '
-        return 'size_' + match[1] if match[1]
+        if Utils.nexus_evergreen?
```
mikewiebe: Similar refactoring can be done here.

lib/cisco_node_utils/bgp.rb
```diff
@@ -440,13 +451,24 @@ def event_history_detail
...
+        else
+          return 'size_' + match[1] if match[1]
+        end
       end
       default_event_history_detail
     end
 
     def event_history_detail=(val)
       size = val[/small|medium|large|disable/]
+      if Utils.nexus_evergreen?
```
mikewiebe: And here.

lib/cisco_node_utils/bgp.rb
mikewiebe: Similar refactoring can be done for other properties below that follow the same pattern.

lib/cisco_node_utils/bgp.rb
mikewiebe: Similar refactoring can be done for other properties below that follow the same pattern.

lib/cisco_node_utils/bgp.rb
```diff
@@ -486,21 +519,41 @@ def event_history_periodic
...
+        else
+          return 'size_' + match[1] if match[1]
+        end
       end
       default_event_history_periodic
     end
 
     def event_history_periodic=(val)
-      size = val[/small|medium|large|disable/]
+      if Utils.nexus_evergreen?
```
mikewiebe: Appears that this could also be further condensed without the need to call `Utils.nexus_evergree?`.

lib/cisco_node_utils/bgp.rb
saichint: Done

lib/cisco_node_utils/bgp.rb
saichint: Done

lib/cisco_node_utils/bgp.rb
saichint: Done here too. I got rid of Utils.nexus_evergreen? everywhere except the default for periodic. The get method for event_history_periodic is a bit of challenge but I got it though I am not sure if this is optimized enough. Let me know after you take another look

mikewiebe: Taking a look now to see if we can somehow get rid of `Utils.nexus_evergreen?` altogether.

mikewiebe:
I think it would be good to add support for event history errors while we are at it.  I5 and later of course.

lib/cisco_node_utils/bgp.rb
mikewiebe: Was not able to work around the `Utils.nexus_evergreen` check for periodic.

lib/cisco_node_utils/cisco_cmn_utils.rb
```diff
@@ -103,6 +103,14 @@ def self.nexus_i2_image
       true if Platform.image_version[/7.0.3.I2/]
     end
 
+    # Helper utility to check for evergreen
```
mikewiebe: I would prefer we create two generic api's as follows:

```
def self.image_version?(ver_regexp)
  require_relative 'platform'
  return true if Platform.image_version[ver_regexp]
end
```

```
def self.chassis_pid?(ver_regexp)
  require_relative 'platform'
  return true if Platform.chassis['pid'][ver_regexp]
end
```

This way we can handle default value differences for future properties without tying it to a specific name like evergreen.

lib/cisco_node_utils/bgp.rb
```diff
@@ -486,21 +495,31 @@ def event_history_periodic
...
+      size = val[/small|medium|large|disable|\A\d+\z/] unless val[/false|true/]
       @set_args[:size] = size.nil? ? '' : "size #{size}"
       @set_args[:state] = val[/false/] ? 'no' : ''
       config_set('bgp', 'event_history_periodic', @set_args)
       set_args_keys_default
     end
 
     def default_event_history_periodic
-      config_get_default('bgp', 'event_history_periodic')
+      if Utils.nexus_evergreen?
```
mikewiebe: We can rewrite this section to use new api's I suggest below as follows.

```
if image_version?(/7.0.3.I2|I3|I4/) or chassis_pid?(/N(5|6|7|8)/)
  config_get_default('bgp', 'event_history_periodic')
else
  config_get('bgp', 'event_history_periodic_bytes')
end
```

```
```

tests/test_router_bgp.rb
```diff
@@ -580,9 +580,99 @@ def test_default_enforce_first_as
...
+    bgp.event_history_periodic = 'size_disable'
+    assert_equal('size_disable', bgp.event_history_periodic)
+    bgp.event_history_periodic = '100000'
+    assert_equal('100000', bgp.event_history_periodic)
+    bgp.event_history_periodic = bgp.default_event_history_periodic
+    assert_equal(bgp.default_event_history_periodic,
+                 bgp.event_history_periodic)
+  end
 
+  def test_event_history
```
mikewiebe: I would just get rid of `test_event_history`

tests/test_router_bgp.rb
```diff
@@ -580,9 +580,99 @@ def test_default_enforce_first_as
     bgp.destroy
   end
 
-  def test_event_history
+  def test_event_history_cli_evergreen
```
mikewiebe: I would get ride of the `_evergreen` bit and just make sure this test tests for event_history_cli functionality regardless of version.  This of course means you do version checks in the test case.  Same goes for the other tests.

lib/cisco_node_utils/cisco_cmn_utils.rb
saichint: Done

lib/cisco_node_utils/bgp.rb
saichint: Done

tests/test_router_bgp.rb
saichint: Done

tests/test_router_bgp.rb
saichint: Done

saichint:
Did all the changes. We have to revisit this for events, errors and objstore sometime in the future.

PR #486, opened 2016-10-05
bgp event_history

saichint:
This PR is for adding event_history_errors and event_history_objstore to the bgp provider. It also fixes event_history_detail which has some issues. Finally it includes some minor fixes in the other event_history properties as the defaults and some enums are set differently from the older releases to evergreen.


mikewiebe:
👍 Looks good.

PR #490, opened 2016-10-18
Rel141/lacp suspend

mikewiebe:
PR https://github.com/cisco/cisco-network-node-utils/pull/489 modified tests validating `lacp_suspend` functionality to shutdown the interface before setting the property.

This PR further refactors the property to automatically shutdown the interface when needed so that this can be managed via puppet without needing to create more then one manifest.

Tested on: n3k, n9k, n5k, n6k

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/interface_portchannel.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface_portchannel.yaml
@@ -61,10 +61,10 @@ lacp_suspend_shut_needed:
   # This is used to determine if the port-channel interface
   # must be shutdown before setting lacp_suspend_individual
   kind: boolean
-  N3k:
-    default_only: false
-  else:
+  N9k:
     default_only: true
+  else:
+    default_only: false
 
 port_hash_distribution:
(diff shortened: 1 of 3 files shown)


lib/cisco_node_utils/cmd_ref/interface_portchannel.yaml
```diff
@@ -57,6 +57,15 @@ lacp_suspend_individual:
     auto_default: false
     default_value: true
 
+lacp_suspend_shut_needed:
+  # This is used to determine if the port-channel interface
+  # must be shutdown before setting lacp_suspend_individual
+  kind: boolean
+  N3k:
```
saichint: this problem is happening on n9k only as of now, so only shut down the interface for n9k. let us not change other platform behavior since they may or may not change.

mikewiebe: Good suggestion.  I made the change to only shutdown the interface when needed on the `n9k` platform.

Changes Tested on: n3k, n9k, n5k, n7k

saichint:
:+1:

PR #491, opened 2016-10-24
Fix/command config issue

mikewiebe:
**Problem Statement:**
Our agents code does the following for any commands that are pre-pended with ‘no’

Example: 'no bfd echo'
- Searches for the complete command ‘no bfd echo’ in the 'show running all' output.
- If it fails to find the command it strips the ‘no’ and searches for the command again in the proper context.
- If it still fails to find the command without the ‘no’ prefix then it assumes the attribute is already disabled and removes it from the list of commands that need to be applied so that we only apply the minimum set of commands to ensure idempotency.

Unfortunately this does not work for commands like ‘no bfd echo’ that do not nvgen when enabled.

This fix addresses the issue in the following ways.
- Removed the logic to strip the 'no' and search for the command again.  This allows commands to be applied in cases where we cannot determine if the feature is enabled or disabled.  This occurs when the command does not nvgen properly in either the enabled or disabled case.  
  - **NOTE:** This will break idempotency for some commands but this is our only option for the `cisco_command_config` provider.  It's more important for the command to be applied in this case.  If idempotency is important, then an existing provider that supports the feature should be used or a new one should be developed.
- Added the ability to specify a new test yaml key `nvgen: false` so that tests don't expect the specified test configuration to be present in the 'show running all' config.
- Added the ability to specify a new test yaml key `setup:` so specify any dependent configuration needed for the test.

Tested on: n3k(I2, I3, I5), n9k(I2, I3, I5), n6k, n7k

### DIFF
--- a/lib/cisco_node_utils/configparser_lib.rb
+++ b/lib/cisco_node_utils/configparser_lib.rb
@@ -125,12 +125,6 @@ module Cisco
             existing << submode.compare_with(config_submode)
             next
           end # if
-
-          prefix, base = base_commands(command)
-          if prefix != '' && !config.include_command?(base)
-            existing << config_line
-            next
-          end
         end
         existing
       end # compare_with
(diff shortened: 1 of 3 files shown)


saichint:
:+1:

PR #492, opened 2016-10-31
hsrp_global and interface_hsrp providers

saichint:
This PR is for adding hsrp_global and interface_hsrp providers.
All tests pass on all relevant platforms.

* interface_hsrp_group is incomplete and it is not included in any tests, will open another PR later once the support is complete, so please do review only for the above two providers.


tests/test_hsrp_global.rb
```diff
@@ -0,0 +1,83 @@
...
+    [:bfd_all_intf,
+     :extended_hold,
+    ].each do |prop|
+      assert_equal(hg.send("default_#{prop}"), hg.send("#{prop}")) if
+        hg.send("#{prop}")
+    end
+  end
+
+  def test_bfd_all_intf
+    hg = create_hsrp_global
```
mikewiebe: IMO, we don't need to obfuscate and wrap a single line in a helper method.  I would prefer we get rid of the create_hsr_global method and just call HsrpGlobal.new directly.

``` ruby
hg = HsrpGlobal.new
```

lib/cisco_node_utils/cmd_ref/hsrp_global.yaml
```diff
@@ -0,0 +1,22 @@
...
+  get_command: "show running-config hsrp all | no-more"
+
+bfd_all_intf:
+  _exclude: [N3k]
+  kind: boolean
+  get_value: '/^hsrp bfd all-interfaces$/'
+  set_value: '<state> hsrp bfd all-interfaces'
+  default_value: false
+
+extended_hold:
```
mikewiebe, line 15: Assuming this nvgens the same way on all nxos platforms that support this property we should really handle this property similar to `dampen_igp_metric` under bgp_af.rb.

**yaml:**

``` yaml
extend_hold:
  kind: int
  get_value: '/^(no )?hsrp timers extend-hold(?:\s+(\d+))?$/'
  set_value: '<state> hsrp timers extended-hold <time>'
  default_value: 10
```

**getter:**

``` ruby
def extend_hold
  match = config_get('hsrp_global', 'extended_hold', @get_args)
  if match.is_a?(Array)
    return nil if match[0][/no/]
    return match[1].to_i if match[1]
  end
  default_extend_hold
end
```

Alternatively, I would like to see if we can use the yaml `auto_default` property and write the yaml/getter as follows.

**yaml:**

``` yaml
extend_hold:
  auto_default: false
  get_value: '/^hsrp timers extend-hold(?:\s+(\d+))?$/'
  set_value: '<state> hsrp timers extended-hold <time>'
  default_value: 10
```

**getter:**

``` ruby
def extend_hold
  config_get('hsrp_global', 'extended_hold', @get_args)
end
```

My hope is that config_get will return the following:
1. `nil` if `no hsrp timers extended-hold` is configured.
2. `10` if `hsrp timers extended-hold` is configured.
3. `11` if `hsrp timers extended-hold 11` is configured.

mikewiebe: Ok, I actually worked up a quick POC for this and need to revise the suggestions above.  The following will work since this time I actually tested it :).

**yaml**

``` yaml
extended_hold:
  get_value: '/^(no )?hsrp timers extended-hold(?:\s+(\d+))?$/'
  set_value: '<state> hsrp timers extended-hold <time>'
  default_value: 10
```

**getter**

``` ruby
    def extended_hold
      match = config_get('hsrp_global', 'extended_hold', @get_args)
      return nil if match[0] == "no "
      return match[1].nil? ? default_extended_hold : match[1].to_i
    end
```

tests/test_hsrp_global.rb
```diff
@@ -0,0 +1,83 @@
...
+    end
+    assert_equal(hg.default_bfd_all_intf, hg.bfd_all_intf)
+    hg.bfd_all_intf = true
+    assert_equal(true, hg.bfd_all_intf)
+    hg.bfd_all_intf = hg.default_bfd_all_intf
+    assert_equal(hg.default_bfd_all_intf, hg.bfd_all_intf)
+  end
+
+  def test_extended_hold
+    hg = create_hsrp_global
```
mikewiebe: We should also test `hg.extended_hold = false`

lib/cisco_node_utils/interface_hsrp.rb
```diff
@@ -0,0 +1,189 @@
...
+    def to_s
+      "interface_hsrp #{name}"
+    end
+
+    def self.interfaces
+      hash = {}
+      intf_list = config_get('interface', 'all_interfaces')
+      return hash if intf_list.nil?
+
+      intf_list.each do |id|
```
mikewiebe: I am a bit confused here.  Why do we need to gather all interfaces on the device in this object.  We are not filtering on only hsrp enabled interfaces so not sure I see the value.  We can just use `Interface.interfaces` for this.  No need to duplicate the code in two different objects.

**Thinking more**Did you consider adding these hsrp properties to `interface.rb` instead of creating a new object?  I think that would be preferable since we can't really gather a list of hsrp_interface objects.  Contrast this with `interface_ospf`, we only gather interfaces if they are already part of an ospf area (https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/interface_ospf.rb#L52).  I don't think that applies to hsrp.

lib/cisco_node_utils/interface_hsrp.rb
```diff
@@ -0,0 +1,189 @@
...
+      set_args_keys(state: state)
+      Feature.bfd_enable if val
+      config_set('interface_hsrp', 'bfd', @set_args)
+    end
+
+    def default_bfd
+      config_get_default('interface_hsrp', 'bfd')
+    end
+
+    # hsrp delay minimum and reload are in the same CLI
```
mikewiebe: I had to do something similar for `router_ospf_vrf` property `timer_throttle_lsa`.  I think the following approach is a bit cleaner for the **getters** and avoids using `match[0]` and `match[1]` syntax and a single `config_get` call.

``` ruby
def delay
  match = config_get('interface_hsrp', 'delay', @get_args)
  return match.nil? ? default_delay : match.collect(&:to_i)
end

def default_delay
  [default_delay_minium , default_delay_reload]
end

def delay_minimum
  minimum, _reload = delay
  return minimum.nil? ? default_delay_minimum : minimum
end

def delay_reload
  _minimum, reload = delay
  return reload.nil? ? default_delay_reload : reload
end
```

lib/cisco_node_utils/interface_hsrp.rb
```diff
@@ -0,0 +1,189 @@
...
+      config_get_default('interface_hsrp', 'mac_refresh')
+    end
+
+    def use_bia
+      match = config_get('interface_hsrp', 'use_bia', @get_args)
+      return default_use_bia unless match
+      match.include?('scope') ? :use_bia_intf : :use_bia
+    end
+
+    def use_bia=(val)
```
mikewiebe: Please add a comment here to indicate the following:
`# Return early if device already set to the correct value`

lib/cisco_node_utils/interface_hsrp.rb
```diff
@@ -0,0 +1,189 @@
...
+      return if val == use_bia
+      # need to reset before set
+      if val
```
mikewiebe: Why are you checking `val`? When will it not be set?

mikewiebe: **Never mind**  Default case :)

mikewiebe:
@saichint Done with the review.

saichint:
All changes done. Removed interface_hsrp and merged that with interface. Had to change CHANLOG due to conflict and also some yaml files due to N9k-F refactoring

PR #496, opened 2016-11-10
interface hsrp group

saichint:
This PR is for adding a new provider interface_hsrp_group. The CLIs are not very cleanly designed so lot of work was done to take care of many scenarios.
The files to review are:
interface_hsrp_group.yaml
interface_hsrp_group.rb
test_interface_hsrp_group.rb

Please review the files entirely as some of these files were already merged without full review.


lib/cisco_node_utils/cmd_ref/interface_hsrp_group.yaml
```diff
@@ -32,33 +32,46 @@ authentication_timeout:
   kind: int
   default_value: 0
 
+group_name:
+  get_value: '/^name (\S+)/'
+  set_value: '<state> name <word>'
+  default_value: false
+
 hsrp_groups:
```
mikewiebe: _nit_ I think `groups` instead of `hsrp_groups` is fine.

lib/cisco_node_utils/interface_hsrp_group.rb
```diff
@@ -36,6 +36,7 @@ def initialize(interface, group_id, ip_type, instantiate=true)
```
mikewiebe, line 36: The following line of code needs to set instantiate to false.

https://github.com/cisco/cisco-network-node-utils/blob/3a8e8c5d4339e5faf5ab744cc413d23f3b59f3bc/lib/cisco_node_utils/interface_hsrp_group.rb#L48

`hash[intf][id][iptype] = InterfaceHsrpGroup.new(intf, id, iptype, false)`

mikewiebe, line 36: Invalid comment?  Should this be `# Create one hsrp group`?

https://github.com/cisco/cisco-network-node-utils/blob/3a8e8c5d4339e5faf5ab744cc413d23f3b59f3bc/lib/cisco_node_utils/interface_hsrp_group.rb#L67

mikewiebe, line 36: Do you need to specify `iptype:` here?  It's already one of the default keys in `set_args_keys_default` right?

https://github.com/cisco/cisco-network-node-utils/blob/3a8e8c5d4339e5faf5ab744cc413d23f3b59f3bc/lib/cisco_node_utils/interface_hsrp_group.rb#L76

lib/cisco_node_utils/cmd_ref/interface_hsrp_group.yaml
```diff
@@ -1,6 +1,6 @@
 # interface_hsrp_group
```
mikewiebe, line 1: When I configure an hsrp group on my n9k, the default value is `cisco`.  Not sure I understand why it's 'cleartext' here.

https://github.com/cisco/cisco-network-node-utils/blob/3a8e8c5d4339e5faf5ab744cc413d23f3b59f3bc/lib/cisco_node_utils/cmd_ref/interface_hsrp_group.yaml#L16

lib/cisco_node_utils/interface_hsrp_group.rb
mikewiebe: Not sure I understand the following line.
https://github.com/cisco/cisco-network-node-utils/blob/3a8e8c5d4339e5faf5ab744cc413d23f3b59f3bc/lib/cisco_node_utils/interface_hsrp_group.rb#L104

If str is nil, then it found no authentication configs.  How did you determine what the default values were for authentication in that case?  Is that even possible?

mikewiebe:
```
regexp = Regexp.new('(?<authtype>text|md5)'\
           ' *(?<keytype>key-chain|key-string|\S+)'\
           ' *(?<enctype>7|\S+)?'\
           ' *(?<keystring>\S+)?'\
           ' *(?<compatible>compatibility)?'\
           ' *(?<tval>timeout \d+)?')
```

tests/test_interface_hsrp_group.rb
```diff
@@ -14,7 +14,6 @@
```
mikewiebe, line 14: Make sure to test collection size zero.

https://github.com/cisco/cisco-network-node-utils/blob/bce1be11c2bfe582e02edeffb41d6bd7e532b475/tests/test_interface_hsrp_group.rb#L66

mikewiebe, line 14: You can collapse this conversion into a single line: 
https://github.com/cisco/cisco-network-node-utils/blob/bce1be11c2bfe582e02edeffb41d6bd7e532b475/lib/cisco_node_utils/interface_hsrp_group.rb#L325

This:

```
      mac.tr!('.', '')
      [2, 5, 8, 11, 14].each do |i|
        mac.insert i, ':'
      end
mac
```

Becomes this:

`mac.tr('.', '').scan(/.{1,2}/).join(':')`

mikewiebe, line 14: Conversely this:

```
        mac.tr!(':', '')
        mac.insert 4, '.'
        mac.insert 9, '.'
```

can become this:
`mac.tr(':', '').scan(/.{1,4}/).join('.')`

mikewiebe, line 14: This test case is really long.  Can we break this up into smaller meaningful test cases so it's easier to maintain?

https://github.com/cisco/cisco-network-node-utils/blob/bce1be11c2bfe582e02edeffb41d6bd7e532b475/tests/test_interface_hsrp_group.rb#L106

mikewiebe, line 14: Same the remaining tests.  Very long and my fear is that they will be difficult to maintain if things change going forward.  Let's make them more granular if possible.

https://github.com/cisco/cisco-network-node-utils/blob/bce1be11c2bfe582e02edeffb41d6bd7e532b475/tests/test_interface_hsrp_group.rb#L193

mikewiebe:
@saichint I am done with this round of reviews.  But forgot to submit them until just now! :)  Sorry.

mikewiebe (changes requested):

tests/test_interface_hsrp_group.rb
saichint: already there in the test

saichint: done

saichint:
made all changes. Also the travis-CI succeeded, the previous error must be transient as you said.

mikewiebe (approved):

lib/cisco_node_utils/interface_hsrp_group.rb
```diff
@@ -393,6 +474,10 @@ def priority_level_set(pri, lower, upper)
       config_set('interface_hsrp_group', 'priority_level', @set_args)
     end
 
+    # This CLI can take forms like:
```
mikewiebe, line 477: Overall changes looking better but I think this one could also use keyword regexp instead of positional array.

https://github.com/cisco/cisco-network-node-utils/blob/b1c5459c699ae88e2b2ddf9e2e8866f6d062a0b2/lib/cisco_node_utils/interface_hsrp_group.rb#L481

Here is a simple pattern that will work in the yaml file.

```
2.2.1 :066 > l1
 => "timers  1  3" 
2.2.1 :067 > l2
 => "timers msec 300  3" 
2.2.1 :068 > l3
 => "timers msec 750 msec 2500" 
2.2.1 :069 > /timers.*\s+(\d+).*\s+(\d+)/.match(l1)
 => #<MatchData "timers  1  3" 1:"1" 2:"3"> 
2.2.1 :070 > /timers.*\s+(\d+).*\s+(\d+)/.match(l2)
 => #<MatchData "timers msec 300  3" 1:"300" 2:"3"> 
2.2.1 :071 > /timers.*\s+(\d+).*\s+(\d+)/.match(l3)
 => #<MatchData "timers msec 750 msec 2500" 1:"750" 2:"2500"> 
2.2.1 :072 > 
```

saichint: The pattern you gave me has issues because I need to capture the "msec" at one or 2 places as well as these are configurable parameters (basically users can provide the timers in seconds or msec so they have the option to choose msec as true or false). Since this is only capturing the integers, it is not going to work fully. I did slightly different but used regex nevertheless in the ruby file instead of yaml. It is working now, please let me know if this is fine.

PR #497, opened 2016-11-17
Update platform_to_filter api for n9k

mikewiebe:
This update fixes a bug that caused the command reference hash to pull in yaml entries for both `n9k` and `n9k-f` platforms.

With this update we now correctly match fretta or non-fretta based n9k platforms but not both.

**Testing**
 - In progress on n9k and n9k-f platforms.  Will update PR with results when available.


saichint:
:+1:

PR #499, opened 2016-11-17
Add short tag test api

mikewiebe:
This adds a helper method called `product_tag` that returns a short but meaningful product tag that can be used in testing.


saichint:
:+1:

PR #500, opened 2016-11-23
Rel150/atherton fretta fixes

mikewiebe:
Fixes needed for fretta_camden and atherton enablement.

- Fabricpath was broken by a previous commit.  Restored the correct yaml entries.
- Added missing exclude for vlan.
- Added a rescue for the yum provider to catch two classes of errors but raise the same error in both cases.  There are times when install fails because of either of the two error classes.


saichint:
:+1:

PR #501, opened 2016-11-28
pim bfd

saichint:
This PR is for adding bfd property to pim and also to interface. All tests pass.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/pim.yaml
+++ b/lib/cisco_node_utils/cmd_ref/pim.yaml
@@ -21,6 +21,12 @@ all_ssm_ranges:
   multiple:
   get_value : '/^<afi> pim ssm range (.*)$/'
 
+bfd:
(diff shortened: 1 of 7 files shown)


chrisvanheuveln:
NB: Have you considered ipv6 pim bfd?

lib/cisco_node_utils/cmd_ref/pim.yaml
```diff
@@ -21,6 +21,12 @@ all_ssm_ranges:
   multiple:
   get_value : '/^<afi> pim ssm range (.*)$/'
 
+bfd:
+  kind: boolean
+  get_value: '/^ip pim bfd$/'
```
mikewiebe: To @chrisvanheuveln point about ipv6 support, this should be.

`get_value: '/^<afi> pim bfd$/'` and
`set_value: "<state> <afi> pim bfd"`

PR #502, opened 2016-11-30
Rel150/atherton fretta fixes

mikewiebe:
**Summary of changes:**
- In Atherton, commands that nvgen with double quotes are handled differently by nxapi.
  - Atherton: Double quotes are preserved in the output by escaping them.
  - Non Atherton(Helsinki, Camden --> Evergreen) Double quotes are removed from the output by nxapi.
  - I created a helper method `Utils.add_quotes` to add leading/trailing double quotes where needed for tests running on Atherton.
- Updates to `test_bgp.af` to include properties that are supported under l2vpn evpn address-family on Atherton only

**Tested on:** n7k (hesinki and atherton), n9k(evergreen), n5/6k(kokomo)

### DIFF
--- a/lib/cisco_node_utils/cisco_cmn_utils.rb
+++ b/lib/cisco_node_utils/cisco_cmn_utils.rb
@@ -360,5 +360,21 @@ module Cisco
       end
       merged
     end # merge_range
+
+    # Add double quotes to string.
+    #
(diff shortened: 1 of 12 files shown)


saichint:
The version check of '8.0' is too generic and this needs to be modified when we get the actual version numbers after FCS. For now it is ok.

PR #504, opened 2016-12-05
n7k atherton hsrp

saichint:
This PR is for enabling interface_hsrp and interface_hsrp_group on n7k atherton only. 
The ipv6 virtual ip MUST be in the same subnet as the interface (this restriction is added to the latest atherton code and most likely, the same condition will apply for all other platforms as well once the code gets synced), so some test code is changed to take care of this. The provider code has no changes. All tests pass on relevant platforms.


mikewiebe:
👍

PR #505, opened 2016-12-13
New nxapi_structured infra features

mikewiebe:
**Summary:**
This change introduces additional capability when using `nxapi_structured` as the `get_data_format`.  Often the data that gets returned is a table that contains multiple rows of data that have similar characteristics but only one of the rows is desired.  This update makes use of `get_context` and `get_value` yaml keys to specify enough context to easily drill down to the desired row.

Properties that use `nxapi_structured` where possible have been updated to use the new infrastructure features.  This change also allows us to fix issue: https://github.com/cisco/cisco-network-puppet-module/issues/408.

**Spot Check Testing On: n3k(I5), n5k, n6k, n7k(Helsinki and Atherton), n9k(I2), n9k fretta**

Full minitest regression in progress.

### DIFF
--- a/lib/cisco_node_utils/client/utils.rb
+++ b/lib/cisco_node_utils/client/utils.rb
@@ -144,7 +144,13 @@ class Cisco::Client
         fail "No match found for #{filter}" if data.length == 0
         data = data[0]
       else # data is array or hash
-        filter = filter.to_i if data.is_a? Array
+        if data.is_a? Array
+          final = []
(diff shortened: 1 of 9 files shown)


mikewiebe:
@saichint Done with the changes and testing after the changes.  Please merge once Travis completes if you don't have any more questions/comments.  Thanks!

saichint:
These are some pretty good changes to the infra.
:+1:

PR #507, opened 2017-01-06
route_map

saichint:
This PR is for adding a new provider for route-map configuration. This is very big provider with lot of properties. There are several issues on the platform side, which are being followed up separately. Since this is a large PR, is is being put up for code review due to time constraints. Basically the mini test code has skips for the platforms issues which need to be rewritten to add any bug IDs but the rest of the code should remain as it is. All tests pass on all platforms except for the intended skips.


tests/test_route_map.rb
```diff
@@ -0,0 +1,1462 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative 'ciscotest'
+require_relative '../lib/cisco_node_utils/route_map'
+
+def evergreen_higher?
```
mikewiebe: Evergreen or later

tests/test_route_map.rb
```diff
@@ -0,0 +1,1462 @@
...
+    rm = create_route_map
+    assert_equal(rm.default_match_community, rm.match_community)
+    assert_equal(rm.default_match_community_exact_match, rm.match_community_exact_match)
+    array = %w(public private)
+    rm.match_community_set(array, false)
+    assert_equal(array, rm.match_community)
+    assert_equal(false, rm.match_community_exact_match)
+    rm.match_community_set(array, true)
+    assert_equal(array, rm.match_community)
+    assert_equal(true, rm.match_community_exact_match)
```
mikewiebe: Make sure to use more specific assertion instead of assert_equal if possible.

https://github.com/cisco/cisco-network-node-utils/blob/develop/docs/README-develop-best-practices.md#mt2

PR #508, opened 2017-01-24
SNMP Community - Symbols Cause Churn

mikewiebe:
**Summary:**
This update fixes the issue with the `cisco_snmp_community` provider described in https://github.com/cisco/cisco-network-puppet-module/issues/414

This is a two part fix:
 - Refactor the underlying `snmp_community` object to use nxapi structured output instead of raw cli based output.
     - NOTE: Structured output for this provider is currently only supported on `n3k`, `n9k` and `n9k-f` platforms.  `n5k`, `n6k` and `n7k` will continue to use raw cli based nxapi and hence with this fix will still be unable to handle complex `cisco_snmp_community` names.  A future commit will address this for raw cli based output.
- The infrastructure handling nxapi structured output was refactored to better query data that may include special complex characters.

Tested on: `n3k`, `n9k`, `n9k-f`, `n5k`, `n6k`, `n7k`

### DIFF
--- a/lib/cisco_node_utils/client/utils.rb
+++ b/lib/cisco_node_utils/client/utils.rb
@@ -130,7 +130,7 @@ class Cisco::Client
   # @param keys [Array] lookup sequence
   def self.filter_data(data: nil,
                        keys: nil)
-    return nil if data.nil?
+    return nil if data.nil? || data.empty?
     keys ||= []
     keys.each do |filter|
       # if filter is a Hash and data is an array, check each
--- a/lib/cisco_node_utils/node.rb
+++ b/lib/cisco_node_utils/node.rb
@@ -87,7 +87,7 @@ module Cisco
         ref.hash['get_value'] = get_args[:value]
         ref.hash['drill_down'] = true
         get_args[:value] = nil
-        cach_flush
+        cache_flush
       end
       # cache_flush
       [get_args, ref]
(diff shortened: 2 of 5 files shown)


lib/cisco_node_utils/node.rb
```diff
@@ -87,8 +87,9 @@ def massage_structured(get_args, ref)
         ref.hash['get_value'] = get_args[:value]
         ref.hash['drill_down'] = true
         get_args[:value] = nil
+        cache_flush
       end
-      cache_flush
+      # cache_flush
```
saichint: Remove this line

saichint:
Only a minor comment removal.
Also take a look at this:
https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/node.rb#L83
I think we don't need to check the condition "if ref.hash['get_value'].is_a?(Array)" as the structured output is an array already. Add a comment to remove this condition in the future.

mikewiebe:
@saichint Addressed your PR comments.

lib/cisco_node_utils/client/utils.rb
```diff
@@ -130,7 +130,7 @@ def self.to_regexp(input)
   # @param keys [Array] lookup sequence
   def self.filter_data(data: nil,
                        keys: nil)
-    return nil if data.nil?
+    return nil if data.nil? || data.empty?
```
chrisvanheuveln, line 133: note for next time...
You can combine these into one check:

```
return nil if data.to_s.empty?
```

PR #509, opened 2017-01-24
itd ingress_intf issue fix

saichint:
This PR is for fixing ingress interface issue for itd service. The ingress interface has a next-hop property which is optional. The current code has an issue when the next-hop property is not specified. This PR fixes that issue.


mikewiebe:
@saichint Can you pull a fresh branch for this fix?  There are 45 commits listed that are likely not part of this fix :)

saichint:
opened another PR using a new branch

PR #513, opened 2017-01-27
route_map test skip for dplus images

saichint:
This PR is for skipping 2 tests (match_metric and extcommunity_rt) on dplus images of n9k because of platform issues. Tested on dplus n9k and they skip.

### DIFF
--- a/tests/test_route_map.rb
+++ b/tests/test_route_map.rb
@@ -26,8 +26,14 @@ def older_n9k?
   false
 end
 
-def evergreen_n9k?
-  return true if Utils.image_version?(/7.0.3.I5/) &&
+def dplus_n9k?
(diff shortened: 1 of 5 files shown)


tests/test_route_map.rb
```diff
@@ -340,8 +346,9 @@ def test_match_ipv6_multicast
   end
 
   def test_match_metric
-    # bug on fretta
-    skip('platform not supported for this test') if node.product_id[/N9.*-F/]
+    # bug CSCvc82934 on fretta and n9k running dplus
+    skip('platform not supported for this test') if node.product_id[/N9.*-F/] ||
```
mikewiebe, line 350: Would it be more appropriate to use `skip_legacy_defect` instead and list the ddts?

PR #515, opened 2017-02-10
Add logic to handle EOFError

mikewiebe:
**Summary:**
- There are cases where sending an HTTP request to nxapi may encounter an `EOFError`.  This update adds retry logic to resend the request and make the client handle this condition and self heal if possible.
  - We give up after 2 attempts.
- Additionally this update fixes a problem where some platforms require both `no feature-set fex` followed by `no install feature-set fex` and some don't.  The update to the test code handles both of these cases.

Tested on: n3k, n9k, n7k


mikewiebe:
@rahushen I assigned this to @saichint  but feel free to review this also if you have time! :)

saichint:
:+1:

PR #517, opened 2017-02-14
Add load-interval properties to interface

saichint:
This PR is for adding load-interval delay properties to interface provider. All tests pass on all platforms.

### DIFF
--- a/lib/cisco_node_utils/interface.rb
+++ b/lib/cisco_node_utils/interface.rb
@@ -806,6 +806,61 @@ module Cisco
       config_set('interface', 'feature_lacp', state: val ? '' : 'no')
     end
 
+    def load_interval_counter_1_delay
+      return nil if @name[/loop/i]
+      config_get('interface', 'load_interval_counter_1_delay', name: @name)
+    end
+
+    def load_interval_counter_1_delay=(val)
+      fail ArgumentError, 'Interface cannot be loopback' if @name[/loop/i]
+      config_set('interface', 'load_interval_counter_1_delay',
+                 name: @name, delay: val)
+    end
+
+    def default_load_interval_counter_1_delay
+      return nil if @name[/loop/i]
+      # for vlan and bdi the default is 60
+      if @name[/(vlan|bdi)/i]
+        config_get_default('interface',
+                           'load_interval_counter_1_delay_vlan_bdi')
+      else
+        config_get_default('interface', 'load_interval_counter_1_delay')
+      end
+    end
+
+    def load_interval_counter_2_delay
+      return nil if @name[/loop/i]
+      config_get('interface', 'load_interval_counter_2_delay', name: @name)
+    end
+
+    def load_interval_counter_2_delay=(val)
+      fail ArgumentError, 'Interface cannot be loopback' if @name[/loop/i]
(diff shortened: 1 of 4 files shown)


lib/cisco_node_utils/interface.rb
```diff
@@ -806,6 +806,61 @@ def feature_lacp_set(val)
       config_set('interface', 'feature_lacp', state: val ? '' : 'no')
     end
 
+    def load_interval_counter_1_delay
+      return nil if @name[/loop/i]
+      config_get('interface', 'load_interval_counter_1_delay', name: @name)
+    end
+
+    def load_interval_counter_1_delay=(val)
+      fail ArgumentError, 'Interface cannot be loopback' if @name[/loop/i]
```
mikewiebe: Are there any other logical interfaces we should consider or only loopbacks?

lib/cisco_node_utils/interface.rb
```diff
@@ -806,6 +806,61 @@ def feature_lacp_set(val)
       config_set('interface', 'feature_lacp', state: val ? '' : 'no')
     end
 
+    def load_interval_counter_1_delay
+      return nil if @name[/loop/i]
```
mikewiebe: Check to see if this `return nil` is even needed for the `puppet resource` command.

saichint:
Done with changes

lib/cisco_node_utils/interface.rb
```diff
@@ -806,6 +806,62 @@ def feature_lacp_set(val)
...
+      if @name[/(vlan|bdi)/i]
+        config_get_default('interface',
+                           'load_interval_counter_1_delay_vlan_bdi')
+      else
+        config_get_default('interface', 'load_interval_counter_1_delay')
+      end
+    end
+
+    def load_interval_counter_2_delay
+      return nil if @name[/loop/] || @name[/ethernet.*\S+\.\d+$/]
```
mikewiebe, line 832: Can you add a test for this nil condition for both loopbacks and subinterfaces?

lib/cisco_node_utils/interface.rb
```diff
@@ -806,6 +806,62 @@ def feature_lacp_set(val)
...
+      config_get('interface', 'load_interval_counter_2_delay', name: @name)
+    end
+
+    def load_interval_counter_2_delay=(val)
+      fail ArgumentError, 'Interface cannot be sub-intf or loopback' if
```
mikewiebe, line 837: Add a test for subinterfaces and loopbacks to verify this ArgumentError.  You can probably just loop through all 3 getters and setters.

mikewiebe:
@saichint Couple more comments to add negative test cases.

lib/cisco_node_utils/interface.rb
saichint: added

lib/cisco_node_utils/interface.rb
saichint: added

saichint: I did not add loop for get/set. I think the tests should be clean and readable. If you disagree, I will change it.

saichint:
let me know if you have more comments

PR #519, opened 2017-02-22
fix for double quotes errors for passwords

saichint:
This PR is for fixing errors while setting encryption passwords for tacacs and radius. The fix is basically to add double quotes to passwords if they are not already present when we do 'config get' or 'config_set'. The corresponding PR for puppet module is also being opened. The code works in tandem, so please refer to that PR as well.

### DIFF
--- a/lib/cisco_node_utils/radius_global.rb
+++ b/lib/cisco_node_utils/radius_global.rb
@@ -104,8 +104,7 @@ module Cisco
       str = config_get('radius_global', 'key')
       return if str.nil?
       str = str.strip
-      str = "\"#{str}\"" unless str.start_with?('"') && str.end_with?('"')
-      str
+      Utils.add_quotes(str)
     end
 
     def key_set(value, format)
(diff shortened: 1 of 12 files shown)


lib/cisco_node_utils/radius_global.rb
```diff
@@ -101,7 +101,11 @@ def key_format
     end
 
     def key
-      config_get('radius_global', 'key')
+      str = config_get('radius_global', 'key')
+      return if str.nil?
+      str = str.strip
+      str = "\"#{str}\"" unless str.start_with?('"') && str.end_with?('"')
```
mikewiebe: Add a helper method in https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/cisco_cmn_utils.rb for this and call instead of duplicating.

```ruby
add_quotes(value)
  value = "\"#{value}\"" unless value.start_with?('"') && value.end_with?('"')
  value
end
```

mikewiebe:
@saichint Make sure you run the tests across all releases.  Nxapi behaves differently based on release so want to make sure we don't have any unintended side-effects.  (fretta, camden -> ifd6, helsinki and atherton)

saichint:
Done with changes. Tested on all platforms and all releases. Since on some platforms the adding quotes is not working properly, those platforms are excluded.

PR #520, opened 2017-02-27
Upgrade services

rahushen:
- Updating the upgrade method to handle the delete_boot and force_all flags.

- Added a method 'image' that returns the booted image.


mikewiebe:
👍

PR #524, opened 2017-03-04
Upgrade services

rahushen:
Added Enhancements to existing APIs.
- Handle Backend Processing Error
- Upgrade? waits for config ready
- Rename Service to Upgrade


lib/cisco_node_utils/upgrade.rb
```diff
@@ -0,0 +1,116 @@
...
+        sleep 1
+        return true if box_online?
+      end
+      fail 'Configuration is still blocked'
+    end
+
+    # Attempts to upgrade the device to 'image'
+    def self.upgrade(version, image, media='bootflash:', del_boot=false,
+                     force_all=false)
+      # IMPORTANT - Check if version of image equals the version provided.
```
mikewiebe: Couple things here.
1. We should change media to be uri for consistency.
2. Given that we only support bootflash at the moment, we should probably fail if anything other then bootflash is used as a uri

tests/test_upgrade.rb
```diff
@@ -13,88 +13,94 @@
...
 
   def test_clear_status
-    Service.clear_status
-    refute(Service.upgraded?)
+    Upgrade.clear_status
+    refute(Upgrade.upgraded?)
   end
 
   def test_delete
     shell_command('touch /bootflash/foobar')
```
mikewiebe: Just realized that `shell_command` will create a file on bootflash if we are running this in a `native` or `guestshell` environment but will not work for oac.  The `shell_command` api is just a wrapper around  `run bash` or `run guestshell` but there is no equivalent `run oac` command today.  Not sure if this will work, but see if you can replace `shell_command('touch /bootflash/foobar')` with `config("show version > bootflash:foo.txt")` instead.  You may need to escape the `>`.

mikewiebe:
👍

PR #526, opened 2017-03-06
Add missing commit step to yum workflow

mikewiebe:
Our yum provider was not removing inactive packages from the device when the `remove` method was called.  The deactivated package requires a `commit` step before final removal.


saichint:
:+1:

PR #529, opened 2017-03-08
Add LACP mode to interface_channel_group

ghost:
I'm sure there's a bunch of edits needed but wanted to put in the PR for basic scaffolding of this change. I have not had a chance yet to run through the test suites against a live switch to validate if this will work as expected. Also, this would require updates on puppet/chef repos to expose the new property. All precommit/push tests passed however.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/interface_channel_group.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface_channel_group.yaml
@@ -16,13 +16,17 @@ channel_group:
   kind: int
   default_value: false
   nexus:
-    get_value: '/^channel-group (\d+)$/'
+    get_value: '/^channel-group (\d+)/'
   ios_xr:
     get_value: '/^bundle id (\d+).*$/'
     set_value: '<state> bundle id <group>'
 
 channel_group_set:
   set_value: '<state> channel-group <group> <force> mode <mode>'
+
+channel_mode:
+  kind: string
+  get_value: '/^channel-group \d+ mode (\w+)$/'
 
 description:
   kind: string
@@ -39,10 +43,6 @@ feature_lacp:
     set_value: "<state> feature lacp"
   ios_xr:
     default_only: true
-
-mode:
-  kind: string
-  get_value: '/^<state> channel-group (\d+) mode <mode>$/'
 
 shutdown:
   kind: boolean
--- a/lib/cisco_node_utils/cmd_ref/interface_channel_group.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface_channel_group.yaml
@@ -16,7 +16,7 @@ channel_group:
   kind: int
   default_value: false
   nexus:
-    get_value: '/^channel-group (\d+)\s?\S*\s?\S*$/'
+    get_value: '/^channel-group (\d+)/'
     set_value: '<state> channel-group <group> <force>'
   ios_xr:
     get_value: '/^bundle id (\d+).*$/'
--- a/lib/cisco_node_utils/cmd_ref/interface_channel_group.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface_channel_group.yaml
@@ -26,7 +26,7 @@ channel_group_mode:
   kind: string
   get_value: '/^channel-group \d+ mode (\w+)$/'
   set_value: '<state> channel-group <group> <force> mode <mode>'
-  default_value: 'on'
+  default_value: false
 
 description:
   kind: string
--- a/lib/cisco_node_utils/cmd_ref/interface_channel_group.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface_channel_group.yaml
@@ -17,16 +17,16 @@ channel_group:
   default_value: false
   nexus:
     get_value: '/^channel-group (\d+)/'
+    set_value: '<state> channel-group <group> <force>'
   ios_xr:
     get_value: '/^bundle id (\d+).*$/'
     set_value: '<state> bundle id <group>'
 
-channel_group_set:
-  set_value: '<state> channel-group <group> <force> mode <mode>'
-
-channel_mode:
+channel_group_mode:
   kind: string
   get_value: '/^channel-group \d+ mode (\w+)$/'
+  set_value: '<state> channel-group <group> <force> mode <mode>'
+  default_value: 'on'
 
 description:
   kind: string
--- a/lib/cisco_node_utils/cmd_ref/interface_channel_group.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface_channel_group.yaml
@@ -17,16 +17,32 @@ channel_group:
   default_value: false
   nexus:
     get_value: '/^channel-group (\d+)$/'
-    set_value: '<state> channel-group <group> <force>'
   ios_xr:
     get_value: '/^bundle id (\d+).*$/'
     set_value: '<state> bundle id <group>'
+
+channel_group_set:
+  set_value: '<state> channel-group <group> <force> mode <mode>'
 
 description:
   kind: string
   get_value: '/^description (.*)/'
   set_value: '<state> description <desc>'
   default_value: ''
+
+feature_lacp:
+  kind: boolean
+  nexus:
+    get_command: "show running | i ^feature"
+    context: ~
+    get_value: '/^feature lacp$/'
+    set_value: "<state> feature lacp"
+  ios_xr:
+    default_only: true
+
+mode:
+  kind: string
+  get_value: '/^<state> channel-group (\d+) mode <mode>$/'
 
 shutdown:
   kind: boolean
--- a/lib/cisco_node_utils/cmd_ref/interface_channel_group.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface_channel_group.yaml
@@ -16,7 +16,7 @@ channel_group:
   kind: int
   default_value: false
   nexus:
-    get_value: '/^channel-group (\d+)/'
+    get_value: '/^channel-group (\d+)\s?\S*\s?\S*$/'
     set_value: '<state> channel-group <group> <force>'
   ios_xr:
     get_value: '/^bundle id (\d+).*$/'
@@ -24,7 +24,7 @@ channel_group:
 
 channel_group_mode:
   kind: string
-  get_value: '/^channel-group \d+ mode (\w+)$/'
+  get_value: '/^channel-group \d+\s\S+\s(\w+)$/'
   set_value: '<state> channel-group <group> <force> mode <mode>'
   default_value: false
 
--- a/lib/cisco_node_utils/interface_channel_group.rb
+++ b/lib/cisco_node_utils/interface_channel_group.rb
@@ -80,23 +80,25 @@ module Cisco
 
     # common setter
     def channel_group_set(group, mode)
-      feature_lacp_set if mode == 'active' || mode == 'passive'
-      if group
-        state = ''
-        force = 'force'
-      else
-        state = 'no'
-        group = force = ''
-      end
-      config_set('interface_channel_group', 'channel_group_set',
-                 set_args_keys(state: state, group: group, force: force,
-                               mode: mode))
+      begin
+        feature_lacp_set if mode == 'active' || mode == 'passive'
+        if group
+          state = ''
+          force = 'force'
+        else
+          state = 'no'
+          group = force = ''
+        end
+        config_set('interface_channel_group', 'channel_group_set',
+                   set_args_keys(state: state, group: group, force: force,
+                                 mode: mode))
       rescue Cisco::CliError => e
         # Some XR platforms do not support channel-group configuration
         # on some OS versions. Since this is an OS version difference and not
         # a platform difference, we can't handle this in the YAML.
         raise unless e.message[/the entered commands do not exist/]
         raise Cisco::UnsupportedError.new('interface', 'channel_group')
+      end
     end
 
     # def channel_group=(group)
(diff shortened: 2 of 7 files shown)


lib/cisco_node_utils/cmd_ref/interface_channel_group.yaml
```diff
@@ -17,17 +17,33 @@ channel_group:
...
+channel_group_set:
+  set_value: '<state> channel-group <group> <force> mode <mode>'
+
 description:
   kind: string
   get_value: '/^description (.*)/'
   set_value: '<state> description <desc>'
   default_value: ''
 
+feature_lacp:
```
mikewiebe, line 37: Better to put `feature_lacp:` into the feature specific yaml/api file
* https://github.com/tomcooperca/cisco-network-node-utils/blob/c4609185d3a5d9afc4a7eec20ef8e00d6e91f7bc/lib/cisco_node_utils/cmd_ref/feature.yaml
* https://github.com/tomcooperca/cisco-network-node-utils/blob/c4609185d3a5d9afc4a7eec20ef8e00d6e91f7bc/lib/cisco_node_utils/feature.rb

mikewiebe: We also have a test file for features.  Simple feature tests can be enabled by adding an entry like the following:

https://github.com/tomcooperca/cisco-network-node-utils/blob/c4609185d3a5d9afc4a7eec20ef8e00d6e91f7bc/tests/test_feature.rb#L89

lib/cisco_node_utils/interface_channel_group.rb
```diff
@@ -64,31 +64,63 @@ def channel_group
...
-      # the properties of the port-channel interface will be different from
-      # the ethernet interface. 'force' is not needed if the port-channel is
-      # created as a result of the channel-group cli but since it does no
-      # harm we will use it every time.
-      # IOS XR simply ignores 'force'
+    def mode
+      config_get('interface_channel_group', 'mode', @get_args)
+    end
+
+    def feature_lacp?
```
mikewiebe: Same comment here.  Move these into `feature.rb`.

lib/cisco_node_utils/interface_channel_group.rb
```diff
@@ -64,31 +64,63 @@ def channel_group
...
+                               mode: mode))
     rescue Cisco::CliError => e
-      # Some XR platforms do not support channel-group configuration
-      # on some OS versions. Since this is an OS version difference and not
-      # a platform difference, we can't handle this in the YAML.
       raise unless e.message[/the entered commands do not exist/]
       raise Cisco::UnsupportedError.new('interface', 'channel_group')
     end
 
+    # def channel_group=(group)
```
mikewiebe: You should be able to re-use this setter instead of creating a new method called `channel_group_set` and just call the `mode` method you created above inside the setter instead of passing the value in as an argument.

```ruby
def channel_group=(group)
  feature_lacp_set if mode...
```

tests/test_snmpserver.rb
```diff
@@ -111,7 +111,7 @@ def test_contact_special_chars
     end
     assert_equal(contact, newcontact)
     # set to default
-    snmpserver.contact = snmpserver.default_contact
+    # snmpserver.contact = snmpserver.default_contact
```
mikewiebe: I assume these changes are unrelated...

lib/cisco_node_utils/version.rb
```diff
@@ -14,7 +14,7 @@
 
 # Container module for version number only.
 module CiscoNodeUtils
-  VERSION = '1.5.0'
+  VERSION = '1.5.1'
```
mikewiebe: We are getting ready to release the `1.6.0` version of the gem but you can just leave it as `1.5.0` for now.

lib/cisco_node_utils/interface_channel_group.rb
mikewiebe: @tomcooperca Ignore this comment.  It would be best to create a new api `channel_group_set(group,mode)` like you do here, but we may need to keep the old method around for backward compatibility.  Still working through those details.

ghost: Yeah, this one is a little more complex. For example, if you enter `channel-mode 123 mode on`, the resultant config present in the device is simply `channel-mode 123`.  I modeled using the vPC keepalive setter since it was multiple parameters being based into a single line similar to this type. I'll leave it like the latest commit for now as I was able to get the getters & setter working in IRB on a switch. See below:

```
SB2SF02LF01(config)# int eth1/53
SB2SF02LF01(config-if)# channel-group 100 mode active 
SB2SF02LF01(config-if)# end
SB2SF02LF01# show run int eth1/53

!Command: show running-config interface Ethernet1/53
!Time: Thu Mar  9 19:34:12 2017

version 7.0(3)I4(2)

interface Ethernet1/53
  description I,vPC Peerlink Po100
  switchport
  switchport mode trunk
  spanning-tree port type network
  storm-control broadcast level 90.00
  storm-control multicast level 90.00
  channel-group 100 mode active
  no shutdown

irb(main):008:0> test = Cisco::Node.new()
^[[A=> Node: client:'<Cisco::Client::NXAPI of localhost>' cmd_ref:'CommandReference for 'N9K-C93180YC-EX' (platform:'nexus', data formats:[:nxapi_structured, :cli]) based on 71 files'
irb(main):009:0> test.config_get('interface_channel_group','description',name: 'ethernet1/53')
=> "I,vPC Peerlink Po100"
irb(main):010:0> test.config_get('interface_channel_group','channel_group',name: 'ethernet1/53')
=> 100
irb(main):011:0> test.config_get('interface_channel_group','channel_mode',name: 'ethernet1/53') 
=> "active"
irb(main):006:0> test.config_set('interface_channel_group','channel_group_set',{name: 'ethernet1/53', state: '', force: 'force', group: '200', mode: 'active'})  
=> [{"code"=>"200", "msg"=>"Success", "body"=>{}}, {"code"=>"200", "msg"=>"Success", "body"=>{}}]
---
SB2SF02LF01# show run int eth1/53

!Command: show running-config interface Ethernet1/53
!Time: Thu Mar  9 19:38:40 2017

version 7.0(3)I4(2)

interface Ethernet1/53
  description I,vPC Peerlink Po100
  channel-group 200 mode active
  no shutdown

**Comment by mikewiebe** [comment]
@tomcooperca 

I had a chance to spend some time today thinking about the over all design and here are my thoughts.

Starting with the manifest, I suggest that mode be a separate property instead of trying to combine both into an array or hash value as part of `channel_group`.

```puppet
    cisco_interface_channel_group { 'Ethernet1/2':
      channel_group            => 100,
      channel_group_mode       => 'active',
    }
```
Design Goals:
* We need to make sure we maintain backwards compatibility with manifests that don't specify the `channel_group_mode` property, so let's keep the old `channel_group` getter and setter for now.
* I checked across different Nexus platforms (n3k, n9k and n7k) and all of them have consistent cli for this property including options so the yaml entries should be simple.
* The cleanest option will be to remove the existing channel-group setting if either the group number or mode differs from what's specified in the manifest instead of trying to do an inline replace.

Let me know if you want to setup a working webex session to hammer out the details.

Thanks!

ghost:
Thanks Mike. I agree, we should maintain backward compatibility while trying to expose this property. I used the `vpc.rb` as a model and it looks like the team here has largely solved this with other items. It just needs to be tidied up, tested and validated.

Do you have time tomorrow morning? I'll send you an email :+1:

mikewiebe:
@tomcooperca Yes, tomorrow morning works for me.

mikewiebe:
@tomcooperca In advance of our discussion later today.  I added tests to your diff and refactored stuff a bit.  We can dive into the details when we meet.

```ruby
diff --git a/lib/cisco_node_utils/cmd_ref/interface_channel_group.yaml b/lib/cisco_node_utils/cmd_ref/interface_channel_group.yaml
index d276625..59a5079 100644
--- a/lib/cisco_node_utils/cmd_ref/interface_channel_group.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface_channel_group.yaml
@@ -26,7 +26,7 @@ channel_group_mode:
   kind: string
   get_value: '/^channel-group \d+ mode (\w+)$/'
   set_value: '<state> channel-group <group> <force> mode <mode>'
-  default_value: 'on'
+  default_value: false
 
 description:
   kind: string
diff --git a/lib/cisco_node_utils/interface_channel_group.rb b/lib/cisco_node_utils/interface_channel_group.rb
index 30ac711..983db36 100644
--- a/lib/cisco_node_utils/interface_channel_group.rb
+++ b/lib/cisco_node_utils/interface_channel_group.rb
@@ -63,14 +63,22 @@ module Cisco
     def channel_group_mode
       mode = config_get('interface_channel_group', 'channel_group_mode',
                         @get_args)
-      'on' if mode == 'false'
+      mode = mode ? mode : 'on' if channel_group
       mode
     end
 
     def channel_group_mode=(mode)
-      mode = 'on' unless mode
-      set_args_keys(mode: mode)
-      config_set('interface_channel_group', 'channel_group_mode', @set_args)
+      group = channel_group
+
+      set_args_keys(state: 'no', group: group, force: '')
+      config_set('interface_channel_group', 'channel_group', @set_args)
+      if mode
+        set_args_keys(state: '', group: group, force: 'force', mode: mode)
+        config_set('interface_channel_group', 'channel_group_mode', @set_args)
+      else
+        set_args_keys(state: '', group: group, force: 'force')
+        config_set('interface_channel_group', 'channel_group', @set_args)
+      end
     end
 
     def channel_group
diff --git a/tests/test_interface_channel_group.rb b/tests/test_interface_channel_group.rb
index 15b34c3..51e39b3 100644
--- a/tests/test_interface_channel_group.rb
+++ b/tests/test_interface_channel_group.rb
@@ -49,6 +49,52 @@ class TestInterfaceChanGrp < CiscoTestCase
     skip(e.to_s)
   end
 
+  def test_channel_group_mode
+    i = @intf
+    group = 55
+
+    # Default Case: group = mode = false
+    refute(i.channel_group)
+    refute(i.channel_group_mode)
+
+    # group = 55, mode = on
+    i.channel_group = group
+    assert_equal(group, i.channel_group)
+    assert_equal('on', i.channel_group_mode)
+
+    # group = 55, mode = active
+    i.channel_group_mode = 'active'
+    assert_equal(group, i.channel_group)
+    assert_equal('active', i.channel_group_mode)
+
+    # group = 55, mode = passive
+    i.channel_group_mode = 'passive'
+    assert_equal(group, i.channel_group)
+    assert_equal('passive', i.channel_group_mode)
+
+    # group = 55, mode = on
+    i.channel_group_mode = 'on'
+    assert_equal(group, i.channel_group)
+    assert_equal('on', i.channel_group_mode)
+
+    # group = 66, mode = active
+    group = 66
+    i.channel_group = group
+    i.channel_group_mode = 'active'
+    assert_equal(group, i.channel_group)
+    assert_equal('active', i.channel_group_mode)
+
+    # group = 66, mode = on
+    i.channel_group_mode = false
+    assert_equal(group, i.channel_group)
+    assert_equal('on', i.channel_group_mode)
+
+    # Default Case: group = mode = false
+    i.channel_group = i.default_channel_group
+    refute(i.channel_group)
+    refute(i.channel_group_mode)
+  end
+
   def test_description
     i = @intf
     desc = 'test desc'

```

ghost:
Still need to separate out `feature_lacp`

lib/cisco_node_utils/cmd_ref/interface_channel_group.yaml
```diff
@@ -16,18 +16,34 @@ channel_group:
   kind: int
   default_value: false
   nexus:
-    get_value: '/^channel-group (\d+)$/'
+    get_value: '/^channel-group (\d+)\s?\S*\s?\S*$/'
```
mikewiebe: We don't need to include this change with the current design.  I was only suggesting an alternative way of capturing both the `id` and `mode` but since you have the `channel_group_mode` getter below this can remain unchanged.

`get_value: '/^channel-group (\d+)$/'`

tests/test_snmpserver.rb
mikewiebe: Probably safe to go ahead and uncomment these now.

lib/cisco_node_utils/cmd_ref/interface_channel_group.yaml
ghost: The `$/` might be problematic for portchannels with a configured mode. It won't match with the end-of-line token. if it's excluded, it should not affect the existing `channel_group` prop.

mikewiebe: Fair enough.  You should be able to match either case by simply removing the trailing anchor `$`.

```ruby
get_value: '/^channel-group (\d+)/'
```

```ruby
2.2.1 :010 >   line = 'channel-group 5'
 => "channel-group 5" 
2.2.1 :011 > /^channel-group (\d+)/.match(line)
 => #<MatchData "channel-group 5" 1:"5"> 
2.2.1 :012 > 
2.2.1 :013 >   line = 'channel-group 5 mode active'
 => "channel-group 5 mode active" 
2.2.1 :014 > 
2.2.1 :015 >   /^channel-group (\d+)/.match(line)
 => #<MatchData "channel-group 5" 1:"5"> 
2.2.1 :016 > 
```

mikewiebe:
@tomcooperca I left one more comment about the `get_value` channel-group.  Once this is addressed we should be able to merge.

mikewiebe:
@tomcooperca Looks good.  I will go ahead and merge.  Not sure if you have started working on the puppet side yet, but often additional tweaking may be required on the node_utils side once you get into the details so let me know if you run into any issues.  In particular you *might* find creating a single setter for both id and mode may be easier integrate in puppet.

PR #532, opened 2017-03-17
Handle version difference for development builds

rahushen:
Changing parsing logic to handle development builds

### DIFF
--- a/lib/cisco_node_utils/upgrade.rb
+++ b/lib/cisco_node_utils/upgrade.rb
@@ -58,7 +58,15 @@ module Cisco
       if image && uri
         config_get('upgrade', 'image_version', image: image, uri: uri)
       else
-        config_get('show_version', 'version').split(' ')[0]
+        version = config_get('show_version', 'version')
+        # show version displays version differently for release and
+        # development builds.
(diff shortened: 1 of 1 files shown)


lib/cisco_node_utils/upgrade.rb
```diff
@@ -58,7 +58,15 @@ def self.image_version(image=nil, uri=nil)
...
       else
-        config_get('show_version', 'version').split(' ')[0]
+        version = config_get('show_version', 'version')
+        # show version displays version differently for release and
+        # development builds.
+        # Eg: release build
+        #       NXOS: version 7.0(3)I4(2)
+        # Eg: development build
+        #       NXOS: version 7.0(3)IFD6(1) [build 7.0(3)IGD7(0.65)]
+        return version.split(' ')[0] unless version[/build\s+\S+]/]
```
mikewiebe: Why do we need to split for the release build case?  Can't we just return version unchanged?

rahushen: True. I'm taking that out.

mikewiebe:
👍

PR #533, opened 2017-03-17
fixes for freeport issues

saichint:
This PR is for fixing minitest errors when the tests are run against freeport code. All relevant tests are run on n3k/n9k (beaker too) and they pass.

### DIFF
--- a/lib/cisco_node_utils/vlan.rb
+++ b/lib/cisco_node_utils/vlan.rb
@@ -76,7 +76,7 @@ module Cisco
         (result[2]['body'].split(errors) - ['']).each_slice(2).map(&:join)
       error_list.each do |msg|
         next if ignore_message && msg.to_s.include?(ignore_message)
-        fail Cisco::CliError
+        fail Cisco::CliError, result[2]['body']
       end
     end
 
--- a/lib/cisco_node_utils/vlan.rb
+++ b/lib/cisco_node_utils/vlan.rb
@@ -76,7 +76,7 @@ module Cisco
         (result[2]['body'].split(errors) - ['']).each_slice(2).map(&:join)
       error_list.each do |msg|
         next if ignore_message && msg.to_s.include?(ignore_message)
-        fail result[2]['body']
+        fail Cisco::CliError
       end
     end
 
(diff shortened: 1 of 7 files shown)


tests/test_vlan.rb
```diff
@@ -207,7 +207,7 @@ def test_state_valid
 
   def test_shutdown_extended
     v = Vlan.new(2000)
-    assert_raises(RuntimeError, 'vlan misconfig did not raise RuntimeError') do
+    assert_raises(CliError, 'vlan misconfig did not raise RuntimeError') do
```
mikewiebe: The message should reference `CliError` now instead of `RuntimeError`.

mikewiebe:
@saichint One minor comment.

tests/test_vlan_private.rb
```diff
@@ -381,7 +381,7 @@ def test_association_failure
       assert_equal(result, v1.pvlan_association)
 
       pv_type = 'isolated'
-      assert_raises(RuntimeError, 'vlan misconf did not raise RuntimeError') do
+      assert_raises(CliError, 'vlan misconf did not raise RuntimeError') do
```
mikewiebe: This message needs to be updated.

lib/cisco_node_utils/vlan.rb
```diff
@@ -76,7 +76,7 @@ def cli_error_check(result, ignore_message=nil)
         (result[2]['body'].split(errors) - ['']).each_slice(2).map(&:join)
       error_list.each do |msg|
         next if ignore_message && msg.to_s.include?(ignore_message)
-        fail result[2]['body']
+        fail Cisco::CliError
```
mikewiebe: Did not realize this until now, but we loose granularity.  We still need to bubble the actual error message `result[2]['body']` up so that puppet users see the specific error message.

tests/test_ace.rb
```diff
@@ -112,15 +112,20 @@ def test_addrs
     assert_equal(val, a.src_addr)
     assert_equal(val, a.dst_addr)
 
+    config_no_warn('object-group ip address my_addrgroup_ipv4')
```
mikewiebe, line 115: Did we get confirmation that this object being present is intended/expected behavior in `I6`?

saichint: will send email shortly

PR #538, opened 2017-04-07
eth int default

saichint:
This PR for adding purge_config property to interface provider so that the physical interface can be put back into default state. All tests pass

### DIFF
--- a/lib/cisco_node_utils/interface.rb
+++ b/lib/cisco_node_utils/interface.rb
@@ -1988,5 +1988,17 @@ module Cisco
         end
       end
     end
+
+    def purge_config=(val)
+      fail ArgumentError, 'Interface MUST be ethernet' unless
+        @name[/ethernet/]
+      fail ArgumentError, 'Only true is allowed for this config' unless val
+      config_set('interface', 'purge_config', name: @name) if val
+    end
+
+    def purge_config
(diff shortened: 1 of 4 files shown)


mikewiebe:
Please reference the GitHub issue number so it gets linked.

saichint:
This update fixes the issue with the cisco_interface provider described in https://github.com/cisco/cisco-network-puppet-module/issues/424

lib/cisco_node_utils/interface.rb
```diff
@@ -1988,5 +1988,17 @@ def set_range_based_params(list, param_name)
         end
       end
     end
+
+    def purge_config=(val)
+      fail ArgumentError, 'Interface MUST be ethernet' unless
```
mikewiebe: I would reword this as `'purge_config is only supported on Ethernet interfaces`

lib/cisco_node_utils/interface.rb
```diff
@@ -1988,5 +1988,17 @@ def set_range_based_params(list, param_name)
...
+        @name[/ethernet/]
+      fail ArgumentError, 'Only true is allowed for this config' unless val
```
mikewiebe: I would remove the second `fail` and support both `true` and `false`.  You can simply add an early `return unless val` at the beginning of the method.

PR #539, opened 2017-04-11
SPAN session provider

ghost:
New SPAN session provider for NU. Only have included the more basic settings so far with a standard test.

*Note, due to work restrictions, I don't have a proper dev environment. As such, I haven't ran the minitest for this new provider yet. However, when I do, I will post the test results here.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/span_session.yaml
+++ b/lib/cisco_node_utils/cmd_ref/span_session.yaml
@@ -7,10 +7,10 @@ _template:
 all_sessions:
   multiple:
   get_context: ~
-  get_value: '/^monitor session (\d+)$/'
+  get_value: '/^monitor session (\d+)/'
 
 create:
-  set_value: 'monitor session %s'
+  set_value: 'monitor session <id> <type> <val>'
 
 description:
   kind: string
@@ -28,20 +28,21 @@ destination:
   set_value: '<state> destination interface <intf_name>'
 
 destroy:
-  set_value: 'no monitor session %s'
+  set_value: 'no monitor session <id>'
 
 session_id:
   kind: int
   default_value: false
-  get_value: '/^monitor session (\d+)$/'
-  set_value: '<state> monitor session <id>'
+  get_value: '/^monitor session (\d+)/'
+  set_value: '<state> monitor session <id> <type> <val>'
 
 shutdown:
   kind: boolean
+  default_value: true
   get_context: ["monitor session <id>"]
-  get_value: '/^(\S+) shutdown$/'
+  get_value: '/^no shut$/'
   set_context: ["monitor session <id>"]
-  set_value: '<state> shutdown'
+  set_value: '<state> shut'
 
 source_interface:
   kind: string
@@ -61,4 +62,4 @@ type:
   kind: string
   default_value: 'local'
   get_value: '/^monitor session \d+ type (\w+)$/'
-  set_value: 'monitor session <id> type <type>'
+  set_value: 'monitor session <id> <type> <val>'
--- a/lib/cisco_node_utils/cmd_ref/span_session.yaml
+++ b/lib/cisco_node_utils/cmd_ref/span_session.yaml
@@ -3,6 +3,8 @@
 _template:
   nexus:
     get_command: "show running monitor all"
+    get_context: ["monitor session <id>(?: .*)?"]
+    set_context: ["monitor session <id>"]
 
 all_sessions:
   multiple:
@@ -10,56 +12,52 @@ all_sessions:
   get_value: '/^monitor session (\d+)/'
 
 create:
+  set_context: ~
   set_value: 'monitor session <id>'
 
 description:
   kind: string
-  get_context: ["monitor session <id>"]
   get_value: '/^description (.*)/'
-  set_context: ["monitor session <id>"]
   set_value: '<state> description <description>'
   default_value: ''
 
 destination:
   kind: string
-  get_context: ["monitor session <id>"]
-  get_value: '/^destination interface <intf_name>$/'
-  set_context: ["monitor session <id>"]
+  get_value: '/^destination interface (\S+)$/'
   set_value: '<state> destination interface <intf_name>'
 
 destroy:
+  set_context: ~
   set_value: 'no monitor session <id>'
 
 session_id:
   kind: int
   default_value: false
+  get_context: ~
   get_value: '/^monitor session (\d+)/'
+  set_context: ~
   set_value: '<state> monitor session <id> <type> <val>'
 
 shutdown:
   kind: boolean
   default_value: true
-  get_context: ["monitor session <id>"]
   get_value: '/^no shut$/'
-  set_context: ["monitor session <id>"]
   set_value: '<state> shut'
 
 source_interface:
   kind: string
-  get_context: ["monitor session <id>"]
   get_value: '/^source interface (\S+) (\w+)$/'
-  set_context: ["monitor session <id>"]
   set_value: '<state> source interface <int_name> <direction>'
 
 source_vlan:
   kind: int
-  get_context: ["monitor session <id>"]
   get_value: '/^source vlan (\S+) (\w+)$/'
-  set_context: ["monitor session <id>"]
   set_value: '<state> source vlan <vlans> <direction>'
 
 type:
   kind: string
   default_value: 'local'
+  get_context: ~
   get_value: '/^monitor session \d+ type (\w+)$/'
+  set_context: ~
   set_value: 'monitor session <id> type <type>'
--- a/lib/cisco_node_utils/cmd_ref/span_session.yaml
+++ b/lib/cisco_node_utils/cmd_ref/span_session.yaml
@@ -36,6 +36,13 @@ session_id:
   get_value: '/^monitor session (\d+)$/'
   set_value: '<state> monitor session <id>'
 
+shutdown:
+  kind: boolean
+  get_context: ["monitor session <id>"]
+  get_value: '/^(\S+) shutdown$/'
+  set_context: ["monitor session <id>"]
+  set_value: '<state> shutdown'
+
 source_interface:
   kind: string
   get_context: ["monitor session <id>"]
--- a/lib/cisco_node_utils/cmd_ref/span_session.yaml
+++ b/lib/cisco_node_utils/cmd_ref/span_session.yaml
@@ -17,7 +17,7 @@ description:
   get_context: ["monitor session <id>"]
   get_value: '/^description (.*)/'
   set_context: ["monitor session <id>"]
-  set_value: '<state> description <desc>'
+  set_value: '<state> description <description>'
   default_value: ''
 
 destination:
@@ -62,4 +62,4 @@ type:
   kind: string
   default_value: 'local'
   get_value: '/^monitor session \d+ type (\w+)$/'
-  set_value: 'monitor session <id> <type> <val>'
+  set_value: 'monitor session <id> type <type>'
--- a/lib/cisco_node_utils/cmd_ref/span_session.yaml
+++ b/lib/cisco_node_utils/cmd_ref/span_session.yaml
@@ -10,7 +10,7 @@ all_sessions:
   get_value: '/^monitor session (\d+)/'
 
 create:
-  set_value: 'monitor session <id> <type> <val>'
+  set_value: 'monitor session <id>'
 
 description:
   kind: string
--- a/lib/cisco_node_utils/cmd_ref/span_session.yaml
+++ b/lib/cisco_node_utils/cmd_ref/span_session.yaml
@@ -1,0 +1,57 @@
+# SPAN session
+---
+_template:
+  nexus:
+    get_command: "show running monitor all"
+
+all_sessions:
+  multiple:
+  get_context: ~
+  get_value: '/^monitor session (\d+)$/'
+
+create:
+  set_value: 'monitor session %s'
+
+description:
+  kind: string
+  get_context: ["monitor session <id>"]
+  get_value: '/^description (.*)/'
+  set_context: ["monitor session <id>"]
+  set_value: '<state> description <desc>'
+  default_value: ''
+
+destination:
+  kind: string
+  get_context: ["monitor session <id>"]
+  get_value: '/^destination interface <intf_name>$/'
+  set_context: ["monitor session <id>"]
+  set_value: '<state> destination interface <intf_name>'
+
+destroy:
+  set_value: 'no monitor session %s'
+
+session_id:
+  kind: int
+  default_value: false
+  get_value: '/^monitor session (\d+)$/'
+  set_value: '<state> monitor session <id>'
+
+source_interface:
+  kind: string
+  get_context: ["monitor session <id>"]
+  get_value: '/^source interface (\S+) (\w+)$/'
+  set_context: ["monitor session <id>"]
+  set_value: '<state> source interface <int_name> <direction>'
+
+source_vlan:
+  kind: int
+  get_context: ["monitor session <id>"]
+  get_value: '/^source vlan (\S+) (\w+)$/'
+  set_context: ["monitor session <id>"]
+  set_value: '<state> source vlan <vlans> <direction>'
+
+type:
+  kind: string
+  default_value: 'local'
+  get_value: '/^monitor session \d+ type (\w+)$/'
+  set_value: 'monitor session <id> type <type>'
--- a/lib/cisco_node_utils/span_session.rb
+++ b/lib/cisco_node_utils/span_session.rb
@@ -1,0 +1,40 @@
+# Cisco folks can insert the appropriate license info here... :)
+
+require_relative 'node_util'
+
+module Cisco
+  class SpanSession < NodeUtil
+    attr_reader :session_id
+
+    def initialize(session_id)
+      validate_args(session_id)
+    end
+
+    def self.sessions
+      hash = {}
+      all = config_get('span_session', 'all_sessions')
+      return hash if all.nil?
+
+      all.each do |id|
+        id = id.downcase
+        hash[id] = SpanSession.new(id)
+      end
+      hash
+    end
+
+    def validate_args(session_id)
+      fail TypeError unless session_id.is_a?(Integer)
+      fail ArgumentError unless session_id >= 1 || session_id <= 32
+      @session_id = session_id.downcase
+      set_args_keys
+    end
+
+    def create
+      config_set('span_session', 'create', @session_id)
+    end
+
+    def destroy
+      config_set('span_session', 'destroy', @session_id)
+    end
+
+    
--- a/lib/cisco_node_utils/span_session.rb
+++ b/lib/cisco_node_utils/span_session.rb
@@ -1,9 +1,10 @@
-# Cisco folks can insert the appropriate license info here... :)
-
+# Insert appropriate license here
 require_relative 'node_util'
 require_relative 'interface'
 
+
 module Cisco
+  # node utils class for SPAN (switchport analyzer) sessions
   class SpanSession < NodeUtil
     attr_reader :session_id
 
@@ -59,7 +60,7 @@ module Cisco
 
     def destination=(int)
       # fail if int is not a valid interface
-      fail TypeError unless Interface.interfaces.has_key?(int)
+      fail TypeError unless Interface.interfaces.key?(int)
       config_set('span_session', 'destination', id: @session_id, intf_name: int)
     end
 
@@ -88,7 +89,7 @@ module Cisco
     def source_interface=(sources)
       fail TypeError unless sources.is_a?(Hash)
       sources.each do |name,dir|
-        fail TypeError unless Interface.interfaces.has_key?(name)
+        fail TypeError unless Interface.interfaces.key?(name)
         config_set('span_session', 'source_interface', id: @session_id,
                     state: '', int_name: name, direction: dir)
       end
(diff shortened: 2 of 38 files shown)


mikewiebe:
@tomcooperca I was out of the office Monday - Wed this week so just now catching up on these PR's  I will review them today/tomorrow and post any feedback.

tests/test_span_session.rb
```diff
@@ -0,0 +1,99 @@
...
+require_relative '../lib/cisco_node_utils/span_session'
+require_relative '../lib/cisco_node_utils/interface'
+require_relative '../lib/cisco_node_utils/vlan'
+require_relative '../lib/cisco_node_utils/cisco_cmn_utils'
+
+include Cisco
+
+# TestSpanSession - Minitest for SPAN session node utility
+class TestSpanSession < CiscoTestCase
+  @skip_unless_supported = 'session'
```
mikewiebe: This line needs to match the name of the yaml file.  Otherwise it will always skip the tests.

 `@skip_unless_supported = 'span_session'`

lib/cisco_node_utils/cmd_ref/span_session.yaml
```diff
@@ -0,0 +1,64 @@
+# SPAN session
```
mikewiebe, line 1: Curious about why you named this object `span_session` rather then `monitor_session`.

lib/cisco_node_utils/cmd_ref/span_session.yaml
```diff
@@ -0,0 +1,64 @@
...
+  nexus:
+    get_command: "show running monitor all"
+
+all_sessions:
+  multiple:
+  get_context: ~
+  get_value: '/^monitor session (\d+)$/'
+
+create:
+  set_value: 'monitor session %s'
```
mikewiebe: I would prefer that we use `Key-value style` wildcards instead of `Printf style` wildcards.  Also keeps it consistent with the other properties in this file.

`set_value: 'monitor session <id>`

lib/cisco_node_utils/cmd_ref/span_session.yaml
```diff
@@ -0,0 +1,64 @@
...
+  default_value: ''
+
+destination:
+  kind: string
+  get_context: ["monitor session <id>"]
+  get_value: '/^destination interface <intf_name>$/'
+  set_context: ["monitor session <id>"]
+  set_value: '<state> destination interface <intf_name>'
+
+destroy:
```
mikewiebe: `Key-value style` wildcards instead of `Printf style` wildcards

lib/cisco_node_utils/span_session.rb
```diff
@@ -0,0 +1,125 @@
...
+        id = id.downcase
+        hash[id] = SpanSession.new(id)
+      end
+      hash
+    end
+
+    def validate_args(session_id)
+      fail TypeError unless session_id.is_a?(Integer)
+      fail ArgumentError unless session_id >= 1 || session_id <= 32
+      @session_id = session_id.downcase
```
mikewiebe: Using `downcase` here will generate an error since `session_id` is a `Fixnum`.  `@session_id = session_id` should be sufficient.

lib/cisco_node_utils/span_session.rb
```diff
@@ -0,0 +1,125 @@
...
+      all.each do |id|
+        id = id.downcase
+        hash[id] = SpanSession.new(id)
+      end
+      hash
+    end
+
+    def validate_args(session_id)
+      fail TypeError unless session_id.is_a?(Integer)
+      fail ArgumentError unless session_id >= 1 || session_id <= 32
```
mikewiebe: I would not check the `session_id` range to make sure it's between 1 and 32 inclusive since it could be different across various platforms.

lib/cisco_node_utils/span_session.rb
```diff
@@ -0,0 +1,125 @@
...
+      validate_args(session_id)
+    end
+
+    def self.sessions
+      hash = {}
+      all = config_get('span_session', 'all_sessions')
+      return hash if all.nil?
+
+      all.each do |id|
+        id = id.downcase
```
mikewiebe: Given that id is numeric, you should not need to `downcase`.

lib/cisco_node_utils/span_session.rb
```diff
@@ -0,0 +1,125 @@
...
+        hash[id] = SpanSession.new(id)
+      end
+      hash
+    end
+
+    def validate_args(session_id)
+      fail TypeError unless session_id.is_a?(Integer)
+      fail ArgumentError unless session_id >= 1 || session_id <= 32
+      @session_id = session_id.downcase
+      set_args_keys
```
mikewiebe: I don't see where you are defining `set_args_keys`.  If you want to use it here are some examples from other objects where we define it for use within the object.

https://github.com/cisco/cisco-network-node-utils/blob/45aaac5fb9098b7a9d8322268e390b25ed324c05/lib/cisco_node_utils/vxlan_vtep_vni.rb#L65

The main purpose of this helper method is to create a list of base keys that are used for all properties or most properties in an object and then add property specific keys to the list of base keys as needed and then reset back to the base list.

This object IMO probably does not need it but I will leave that up to you.

lib/cisco_node_utils/span_session.rb
```diff
@@ -0,0 +1,125 @@
...
+require_relative 'node_util'
+require_relative 'interface'
+
+
+module Cisco
+  # node utils class for SPAN (switchport analyzer) sessions
+  class SpanSession < NodeUtil
+    attr_reader :session_id
+
+    def initialize(session_id)
```
mikewiebe: Couple of things for the `initialize` method.

* Add an argument called `instantiate` and set it to `true` by default.
* Call your `create` method when `instantiate` is true so that the monitor session gets created.
* When you add the puppet layer, you will call this method with `instantiate` to false when you query for an existing list of `span_session` resources that are already configured.

```ruby
def initialize(session_id, instantiate=true)
  validate_args(session_id.to_i)
  create if instantiate
end
```

lib/cisco_node_utils/cmd_ref/span_session.yaml
```diff
@@ -0,0 +1,64 @@
...
+  set_value: 'no monitor session %s'
+
+session_id:
+  kind: int
+  default_value: false
+  get_value: '/^monitor session (\d+)$/'
```
mikewiebe: You need to remove the `$` in this case as well so it matches lines with and without the type.

lib/cisco_node_utils/span_session.rb
```diff
@@ -0,0 +1,125 @@
...
+      validate_args(session_id)
```
mikewiebe: Better to cast `session_id` as an integer when you pass it into `validate`.

`validate_args(session_id.to_i)`

lib/cisco_node_utils/span_session.rb
```diff
@@ -0,0 +1,125 @@
...
+    end
+
+    def self.sessions
+      hash = {}
+      all = config_get('span_session', 'all_sessions')
+      return hash if all.nil?
+
+      all.each do |id|
+        id = id.downcase
+        hash[id] = SpanSession.new(id)
```
mikewiebe: Once you add the `instantiate` argument to initialize above, call this with `instantiate` set to `false`.  This way you create a list of objects to manage each `span_session` object but you don't try and configure the device since they already exist.

`hash[id] = SpanSession.new(id, false)`

lib/cisco_node_utils/cmd_ref/span_session.yaml
```diff
@@ -0,0 +1,64 @@
...
+  set_value: '<state> monitor session <id>'
+
+shutdown:
+  kind: boolean
+  get_context: ["monitor session <id>"]
+  get_value: '/^(\S+) shutdown$/'
```
mikewiebe: How does this nvgen on your test device?  In my case I see the following.

**Default State**
```
# show running monitor all
monitor session 1 type local
  no description
  shut
# 
```
I would refactor the yaml entry as follows:

```yaml
  kind: boolean
  default_value: true
  get_context: ["monitor session <id>"]
  get_value: '/^no shut$/'
  set_context: ["monitor session <id>"]
  set_value: '<state> shutdown'
```
In this way, when you call `config_get('span_session', 'shutdown', id: @session_id)` it returns `true` if it's in the shutdown state, else `false`.

lib/cisco_node_utils/cmd_ref/span_session.yaml
```diff
@@ -0,0 +1,64 @@
+# SPAN session
+---
+_template:
+  nexus:
+    get_command: "show running monitor all"
+
+all_sessions:
+  multiple:
+  get_context: ~
+  get_value: '/^monitor session (\d+)$/'
```
mikewiebe: Don't add the `$` anchor to the end of this line, otherwise it will not match cases where additional information exists besides just the session_id.

```
monitor session 1 type local
monitor session 7 type erspan-destination
```

tests/test_span_session.rb
```diff
@@ -0,0 +1,99 @@
...
+    assert_equal(span.session_id, 1)
+    assert_equal(span.type, 'local') # default session type
+    assert(span.shutdown)
+    span.destroy
+  end
+
+  def test_remove_session
+    span = SpanSession.new(1)
+    span.destroy
+    refute(span, 'Session was not cleaned up correctly...')
```
mikewiebe: This should be `refute(span.session_id, 'session_id 1 was not cleaned up correctly...')`.  This will make sure the `session_id` getter returns `false` when the object has been destroyed.

tests/test_span_session.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  def test_create_session_invalid_id
+    e = assert_raises(CliError) { SpanSession.new(33) }
+    assert_match(/Invalid value.range/, e.message)
+  end
+
+  def test_session_type
+    span = SpanSession.new(1)
+    erspan_type = 'erspan-source'
+    span.type = erspan_type
```
mikewiebe: I tried running this test but I get the following error:
```
TestSpanSession#test_session_type:
Cisco::CliError: [#<Cisco::SpanSession:0x007f49f6d16450>] The command 'monitor session 1 type erspan-source' was rejected with error:
ERROR: Invalid ssn type specified
```
You will likely need to remove the session id it it exists and then reconfigure it when changing the type.

lib/cisco_node_utils/cmd_ref/span_session.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+create:
+  set_value: 'monitor session %s'
+
+description:
+  kind: string
+  get_context: ["monitor session <id>"]
+  get_value: '/^description (.*)/'
+  set_context: ["monitor session <id>"]
+  set_value: '<state> description <desc>'
```
mikewiebe: The key here `<desc>` needs to change to `<description>` so that it matches your setter below.

tests/test_span_session.rb
```diff
@@ -0,0 +1,99 @@
...
+    end
+  end
+
+  def test_create_new_session
+    span = SpanSession.new(1)
+
+    assert_equal(span.session_id, 1)
+    assert_equal(span.type, 'local') # default session type
+    assert(span.shutdown)
+    span.destroy
```
mikewiebe: You can actually remove the `span.destroy` lines that you have at the end of each test block because the `teardown` method is automatically called after each test is run.

tests/test_span_session.rb
```diff
@@ -0,0 +1,99 @@
...
+    desc = 'SPAN session 1'
+    span.description = desc
+    assert_equal(span.description, desc)
+    span.destroy
+  end
+
+  def test_session_source_interface
+    span = SpanSession.new(1)
+    po_int = Interface.new('port-channel1')
+    ints = { 'Ethernet1/1' => 'rx',
```
mikewiebe: The rubocop errors are caused by this section of code:

```
Offenses:

lib/cisco_node_utils/span_session.rb:5:1: C: Extra blank line detected.

tests/test_span_session.rb:69:14: C: Align the elements of a hash literal if they span more than one line.

    ints = { 'Ethernet1/1' => 'rx',

             ^^^^^^^^^^^^^^^^^^^^^

tests/test_span_session.rb:70:14: C: Align the elements of a hash literal if they span more than one line.

             'Ethernet1/2' => 'tx',

             ^^^^^^^^^^^^^^^^^^^^^

tests/test_span_session.rb:71:12: C: Align the elements of a hash literal if they span more than one line.

           'port-channel1' => 'both',

           ^^^^^^^^^^^^^^^^^^^^^^^^^

tests/test_span_session.rb:72:17: C: Align the elements of a hash literal if they span more than one line.

                'sup-eth0' => 'rx' }

                ^^^^^^^^^^^^^^^^^^

208 files inspected, 5 offenses detected
```
I believe if you just line them up as follows it will resolve the errors.

```ruby
    ints = { 'Ethernet1/1'   => 'rx',
             'Ethernet1/2'   => 'tx',
             'port-channel1' => 'both',
             'sup-eth0'      => 'rx' }

```

mikewiebe:
@tomcooperca I am finished with the first round of review.  Thanks again for contributing to the module.  Great to see new features being added!

Make sure to add an entry to the CHANGELOG to indicate that a new feature was added.

https://github.com/cisco/cisco-network-node-utils/blob/develop/CHANGELOG.md#added

lib/cisco_node_utils/cmd_ref/span_session.yaml
ghost: Switch Port Analyzer (SPAN) is the feature. `monitor session` is just the CLI stanza so I didn't want to conflate the two. Just seems cleaner and prevents confusion if `monitor session` becomes a different feature altogether in NXOS. Open to feedback on that though and can be easily changed back.

mikewiebe: 👍 Makes sense to me.  I agree with your design choice.

lib/cisco_node_utils/span_session.rb
```diff
@@ -0,0 +1,123 @@
...
+    def shutdown=(bool)
+      fail TypeError unless bool.is_a?(Boolean)
+      config_set('span_session', 'shutdown', id: @session_id, shutdown: bool)
+    end
+
+    def source_interface
+      config_get('span_session', 'source_interface', id: @session_id)
+    end
+
+    def source_interface=(sources)
```
mikewiebe, line 87: @tomcooperca How would you like to represent the `source_interface` and `source_vlan` properties in the manifest?  Typically we have represented these as array's since there are more then one and they are additive.  We will need to re-work the getters/setters for  these but want to make sure you are ok with representing them as array's first.

We also have helper methods to deal with arrays and the vlan ranges that we can take advantage of.

lib/cisco_node_utils/cmd_ref/span_session.yaml
```diff
@@ -0,0 +1,65 @@
+# SPAN session
+---
+_template:
```
mikewiebe, line 3: I made some edits to this file to address the following (diffs below)
* Consolidate `get_context` and `set_context` under `_template` for all applicable properties
* Fixed `destination` `get_value` regexp.
* Fixed `type` `get_value` regexp.

```ruby
diff --git a/lib/cisco_node_utils/cmd_ref/span_session.yaml b/lib/cisco_node_utils/cmd_ref/span_session.yaml
index fd4d2b2..9e78b84 100644
--- a/lib/cisco_node_utils/cmd_ref/span_session.yaml
+++ b/lib/cisco_node_utils/cmd_ref/span_session.yaml
@@ -3,6 +3,8 @@
 _template:
   nexus:
     get_command: "show running monitor all"
+    get_context: ["monitor session <id>(?: .*)?"]
+    set_context: ["monitor session <id>"]
 
 all_sessions:
   multiple:
@@ -10,56 +12,52 @@ all_sessions:
   get_value: '/^monitor session (\d+)/'
 
 create:
+  set_context: ~
   set_value: 'monitor session <id>'
 
 description:
   kind: string
-  get_context: ["monitor session <id>"]
   get_value: '/^description (.*)/'
-  set_context: ["monitor session <id>"]
   set_value: '<state> description <description>'
   default_value: ''
 
 destination:
   kind: string
-  get_context: ["monitor session <id>"]
-  get_value: '/^destination interface <intf_name>$/'
-  set_context: ["monitor session <id>"]
+  get_value: '/^destination interface (\S+)$/'
   set_value: '<state> destination interface <intf_name>'
 
 destroy:
+  set_context: ~
   set_value: 'no monitor session <id>'
 
 session_id:
   kind: int
   default_value: false
+  get_context: ~
   get_value: '/^monitor session (\d+)/'
+  set_context: ~
   set_value: '<state> monitor session <id> <type> <val>'
 
 shutdown:
   kind: boolean
   default_value: true
-  get_context: ["monitor session <id>"]
   get_value: '/^no shut$/'
-  set_context: ["monitor session <id>"]
   set_value: '<state> shut'
 
 source_interface:
   kind: string
-  get_context: ["monitor session <id>"]
   get_value: '/^source interface (\S+) (\w+)$/'
-  set_context: ["monitor session <id>"]
   set_value: '<state> source interface <int_name> <direction>'
 
 source_vlan:
   kind: int
-  get_context: ["monitor session <id>"]
   get_value: '/^source vlan (\S+) (\w+)$/'
-  set_context: ["monitor session <id>"]
   set_value: '<state> source vlan <vlans> <direction>'
 
 type:
   kind: string
   default_value: 'local'
-  get_value: '/^monitor session \d+ type (\w+)$/'
+  get_context: ~
+  get_value: '/^monitor session \d+ type (\S+)$/'
+  set_context: ~
   set_value: 'monitor session <id> type <type>'
```

lib/cisco_node_utils/span_session.rb
```diff
@@ -0,0 +1,123 @@
...
+      end
+    end
+
+    def destination
+      config_get('span_session', 'destination', id: @session_id)
+    end
+
+    def destination=(int)
+      # fail if int is not a valid interface
+      fail TypeError unless Interface.interfaces.key?(int)
```
mikewiebe: You need to downcase `int` so that you find it in the list of available interfaces `fail TypeError unless Interface.interfaces.key?(int.downcase)`

lib/cisco_node_utils/span_session.rb
```diff
@@ -0,0 +1,123 @@
...
+      config_set('span_session', 'destination', id: @session_id, intf_name: int)
```
mikewiebe: Make sure to includde the `state:` key for this `config_set` or it will error out.

lib/cisco_node_utils/span_session.rb
ghost: So for the source interfaces/vlans, I was hoping to use a Hash only because the data is both the interface names/vlan ID's in addition to the direction. I know the `cisco_bgp_af` uses an array-of-arrays for the `networks` property but this was awkward to use with Puppet DSL w/ Hiera data lookups and required a custom function in Ruby to transform a basic Puppet-native data type into the one the property was expecting. There's no clean and easy way to take in (for example) a hash via Hiera then transform it into an array of array pairs. 

Thoughts?
Ideally, my end result as a user of ciscopuppet would be something like this:
```
$span_sources = { 'Ethernet1/1' => 'both', 'Ethernet1/2', => 'in', 'portchannel-10' => 'out' }

cisco_span_session { '5':
  ensure                => present,
  description           => 'local_span_session',
  shutdown              => false,
  type                  => 'local',
  source_interfaces     => $span_sources,
  destination_interface => 'Ethernet1/48'
}
```

mikewiebe: @tomcooperca Makes perfect sense.  I agree that representing this kind of data in a hash is more user-friendly and if it integrates cleanly with Hiera that is even more reason to use a hash.

Have you been able to verify that the code you have now for `source_interface` and `source_vlan` works in your testing?

mikewiebe: @tomcooperca I ran the tests with your latest diffs but still see a couple of failures.

```
TestSpanSession#test_session_source_vlans = 1.99 s = E
TestSpanSession#test_remove_session = 1.54 s = .
TestSpanSession#test_session_type = 1.56 s = F
TestSpanSession#test_create_session_invalid_id = 1.48 s = .
TestSpanSession#test_session_description = 1.61 s = .
TestSpanSession#test_session_destination_int = 2.76 s = .
TestSpanSession#test_create_new_session = 1.57 s = .
TestSpanSession#test_session_source_interface = 2.89 s = E

Finished in 20.339367s, 0.3933 runs/s, 0.4917 assertions/s.

  1) Error:
TestSpanSession#test_session_source_vlans:
TypeError: TypeError
    /nobackup/mwiebe/NU/tom/cisco-network-node-utils/lib/cisco_node_utils/span_session.rb:104:in `block in source_vlan='
    /nobackup/mwiebe/NU/tom/cisco-network-node-utils/lib/cisco_node_utils/span_session.rb:103:in `each_pair'
    /nobackup/mwiebe/NU/tom/cisco-network-node-utils/lib/cisco_node_utils/span_session.rb:103:in `source_vlan='
    tests/test_span_session.rb:83:in `test_session_source_vlans'


  2) Failure:
TestSpanSession#test_session_type [tests/test_span_session.rb:53]:
Expected: "local"
  Actual: "erspan-source"


  3) Error:
TestSpanSession#test_session_source_interface:
TypeError: TypeError
    /nobackup/mwiebe/NU/tom/cisco-network-node-utils/lib/cisco_node_utils/span_session.rb:91:in `block in source_interface='
    /nobackup/mwiebe/NU/tom/cisco-network-node-utils/lib/cisco_node_utils/span_session.rb:89:in `each_pair'
    /nobackup/mwiebe/NU/tom/cisco-network-node-utils/lib/cisco_node_utils/span_session.rb:89:in `source_interface='
    tests/test_span_session.rb:70:in `test_session_source_interface'

8 runs, 10 assertions, 1 failures, 2 errors, 0 skips
```

If you are working on additional changes, I will wait for new diffs.  If you want me to investigate and make modifications let me know.  Thanks!

mikewiebe: @tomcooperca Have not heard back so I am going to go ahead and merge this as is and do the fixup on my side.  I will let you know once it's complete so you can make sure the changes work on your end.

PR #540, opened 2017-04-11
Add missing vpc_domain boolean features

ghost:
Added `peer-switch`, `ip arp synchronize` and `ipv6 nd synchronize` boolean properties to vpc NU provider.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Changelog
 ### New feature support
 
 ### Added
+* Extend vpc with attributes:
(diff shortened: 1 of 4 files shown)


mikewiebe:
@tomcooperca Looks good 👍.

I ran the tests on the n3k, n9k and n7k and all tests passed.

The only thing the remains is to add a note to the CHANGELOG https://github.com/cisco/cisco-network-node-utils/blob/develop/CHANGELOG.md#added for the new properties and then I will merge it.

PR #541, opened 2017-04-13
Interface channel group

saichint:
This PR is for fixing interface_channel_group provider for the channel_group_mode. Some methods were merged to make it more clean and compact. All tests pass on all platforms.

### DIFF
--- a/lib/cisco_node_utils/interface_channel_group.rb
+++ b/lib/cisco_node_utils/interface_channel_group.rb
@@ -64,24 +64,25 @@ module Cisco
     def channel_group_mode
       match = config_get('interface_channel_group', 'channel_group', @get_args)
       return match unless match
-      mode = match[1].nil? ? 'on' : match[1]
+      mode = match[1].nil? ? default_channel_group_mode : match[1]
       mode
     end
 
-    def channel_group_mode_set(group, mode='on')
+    def channel_group_mode_set(group, mode=false)
       cgroup = channel_group
       set_args_keys(state: 'no', group: cgroup, force: '', mode: '', val: '')
       config_set('interface_channel_group', 'channel_group', @set_args) if
         cgroup
       return unless group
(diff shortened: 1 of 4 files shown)


lib/cisco_node_utils/interface_channel_group.rb
```diff
@@ -62,60 +62,45 @@ def set_args_keys(hash={}) # rubocop:disable Style/AccessorMethodName
...
+                      mode: '', val: '')
         config_set('interface_channel_group', 'channel_group', @set_args)
       end
     end
 
     def channel_group
-      config_get('interface_channel_group', 'channel_group', @get_args)
-    end
-
-    def channel_group=(group)
```
mikewiebe, line 90: Should we keep `channel_group=` for backward compatibility?  Will this impact existing manifests if we don't?

mikewiebe:
@saichint Just one comment.  I want to be certain we are not changing the manifest.  We will essentially have two properties now.  `channel_group` and `channel_group_mode` correct?  In other words old manifests that only have `channel_group` specified will still be ok.

lib/cisco_node_utils/interface_channel_group.rb
saichint: not required. The whole point of this is that we just need one method which takes both channel group and mode. If mode is not specified, it will be false by default which the method already defines. 
see this:
https://github.com/cisco/cisco-network-node-utils/blob/channel_group/tests/test_interface_channel_group.rb#L89

saichint: The old manifests will work as before. No change is needed. If we put the mode in the manifest, it will take it, if not, it will be false which is default.

mikewiebe: 👍

PR #542, opened 2017-04-13
(NETDEV-29) Enhance ntp_config and ntp_server

shermdog:
This PR enhances the existing ntp_config and ntp_server types as specified in  puppetlabs/netdev_stdlib#21

ntp_config
 - authenticate
 - trusted_key

ntp_server
 - key
 - maxpoll
 - minpoll
 - source_interface
 - vrf

ntp_auth_key
  - key
  - algorithm
  - mode
  - password

ntp_server initialize now uses options hash
  * Prior to this release ntp_server accepted positional arguments for id and
  prefer.  New behavior is to pass attributes as a hash.

  Example:
  ```
  options = { 'name' => id, 'key' => '999', 'prefer' => 'true',
              'minpoll' => '5', 'maxpoll' => '8', 'vrf' => 'red' }
  Cisco::NtpServer.new(options, true)
  ```

### DIFF
--- a/lib/cisco_node_utils/ntp_server.rb
+++ b/lib/cisco_node_utils/ntp_server.rb
@@ -2,7 +2,7 @@
 #
 # Jonathan Tripathy et al., September 2015
 #
-# Copyright (c) 2014-2016 Cisco and/or its affiliates.
+# Copyright (c) 2014-2017 Cisco and/or its affiliates.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


shermdog:
Opening PR to review current methods, code, etc.  Will be adding tests shortly.

lib/cisco_node_utils/ntp_server.rb
```diff
@@ -66,20 +64,41 @@ def ==(other)
...
       config_set('ntp_server', 'server', state: '', ip: @ntpserver_id,
-                  prefer: @ntpserver_prefer ? 'prefer' : '')
+                  prefer: @prefer == 'true' ? 'prefer' : '',
+                  vrf: @vrf ? "use-vrf #{@vrf}" : '',
+                  key: @key ? "key #{@key}" : '',
+                  minpoll: @minpoll ? "minpoll #{@minpoll}" : '',
+                  maxpoll: @maxpoll ? "maxpoll #{@maxpoll}" : '')
     end
 
     def destroy
```
mikewiebe, line 76: I believe you may be able to simplify the `destroy` method and only specify the `@ntpserver_id`. I did some quick testing and I was able to remove the config without specifying any of the other parameters.

`config_set('ntp_server', state: 'no', ip: @ntpserver_id)`

lib/cisco_node_utils/ntp_server.rb
```diff
@@ -66,20 +64,41 @@ def ==(other)
...
-                 state: 'no', ip: @ntpserver_id, prefer: '')
+                 state: 'no', ip: @ntpserver_id, prefer: '', vrf: '',
+                 key: '', minpoll: '', maxpoll: '')
     end
 
     def name
       @ntpserver_id
     end
 
+    def key
```
mikewiebe: The following is simpler and will get rid of the rubocop errors for the object properties.

`attr_reader :key, :maxpoll, :minpoll, :prefer, :vrf`

mikewiebe:
@shermdog Looks good to me.  I gave you a few inline comments and here are a few general housekeeping things:

* Update Copyright date ranges for any files you touch
* Update the https://github.com/cisco/cisco-network-node-utils/blob/develop/CHANGELOG.md#changelog
* Update/add tests in:
  * https://github.com/cisco/cisco-network-node-utils/blob/develop/tests/test_ntp_config.rb
  * https://github.com/cisco/cisco-network-node-utils/blob/develop/tests/test_ntp_server.rb
* Resolve the rubocop errors.

lib/cisco_node_utils/ntp_server.rb
shermdog: I get an exception without it

```
No value specified for 'prefer' in 'no ntp server 2.2.2.2 <prefer> <vrf> <key> <minpoll> <maxpoll>'
```

mikewiebe: Ah, that's because the other properties need to get passed in as empty string.  I suppose this is fine.

tests/test_ntp_config.rb
```diff
@@ -26,28 +26,75 @@ def setup
...
+    # teardown runs at the end of each test
+    no_ntpconfig
+    super
+  end
+
   def no_ntpconfig
     # Turn the feature off for a clean test.
     if platform == :ios_xr
-      config("no ntp source #{interfaces[0]}")
+      config("no ntp source #{interfaces[1]}")
```
mikewiebe, line 38: Why the change from `interface[0]` to `interface[1]`?

shermdog: The test was using `interface[1]` but the teardown was not.  This was leaving cruft behind after the tests were complete.

tests/test_ntp_server.rb
```diff
@@ -83,4 +92,54 @@ def test_multiple
     refute_includes(Cisco::NtpServer.ntpservers, id1)
     refute_includes(Cisco::NtpServer.ntpservers, id2)
   end
+
+  def test_domain_name
+    id = '0.us.pool.ntp.org'
```
mikewiebe: When I run this on my testbed it fails because we are behind a proxy.  This is fine for now, but I will probably move this into a yaml configuration file that automatically skips the test if the config file is unavailable and then anyone that wants to run this test can add any internal or external servers

tests/test_ntp_server.rb
```diff
@@ -83,4 +92,54 @@ def test_multiple
     refute_includes(Cisco::NtpServer.ntpservers, id1)
     refute_includes(Cisco::NtpServer.ntpservers, id2)
   end
+
+  def test_domain_name
```
mikewiebe: Does `test_domain_name` pass for you?  When I run it on my testbed the configuration is accepted but the ntp server does not actually get configured...  investigating to see if it's just the version of software I am running.

tests/test_ntp_config.rb
mikewiebe: 👍

tests/test_ntp_server.rb
shermdog: It does, but requires DNS to be functional in the default VRF.

On some platforms / os the device will silently fail if it cannot resolve the hostname.

I'm actually okay with taking out the hostname test as it's problematic.

mikewiebe:
@shermdog I want to wait on merging this until we are ready to merge https://github.com/cisco/cisco-network-puppet-module/pull/447.  If we merge now, our beaker tests will start to fail because of the changes you made to pass opts as a hash instead of positional args.

tests/test_ntp_server.rb
mikewiebe: Yeah.. let's just remove it for now.

mikewiebe:
👍

PR #545, opened 2017-04-27
Exclude

saichint:
This PR is for adding some properties which were excluded as there was no support before. With new software versions on the nodes, some of the excluded properties are now supported on the platform and now on puppet. They still skip for the older versions of the software. Tested and found ok.

### DIFF
--- a/docs/README-develop-node-utils-APIs.md
+++ b/docs/README-develop-node-utils-APIs.md
@@ -178,7 +178,7 @@ The properties in this example require additional context for their `get_value` 
 feature:
   # feature eigrp must be enabled before configuring router eigrp
   kind: boolean
-  os_version: 'N3k:7.0.3.I2.1,N9k:7.0.3.I2.1'
+  os_version: 'N3k:7.0.3.I2.1, N9k:7.0.3.I2.1'
   get_command: 'show running eigrp all'
   get_value: 'feature eigrp'
   set_value: '<state> feature eigrp'
--- a/docs/README-develop-node-utils-APIs.md
+++ b/docs/README-develop-node-utils-APIs.md
@@ -159,6 +159,7 @@ The following basic command_reference parameters will be defined for each resour
  4. `default_value:` This is typically the "factory" default state of the property, expressed as an actual value (true, 12, "off", etc)
  5. `kind:` The data type of this property. If omitted, the property will be a string by default. Commonly used values for this property are `int` and `boolean`.
  6. `multiple:` By default a property is assumed to be found once or not at all by the `get_command`/`get_value` lookup, and an error will be raised if multiple matches are found. If multiple matches are valid and expected, you must set `multiple: true` for this property.
+ 7. `os_version:` By default a property is assumed to be supported on all version of software. If it is supported starting from a particular version, then this field can be used to specify the version for a particular product_id.
 
 There are additional YAML command parameters available which are not covered by this document. Please see the [README_YAML.md](../lib/cisco_node_utils/cmd_ref/README_YAML.md) document for more information on the structure and semantics of these files.
 The properties in this example require additional context for their `get_value` and `set_value` because they need to differentiate between different eigrp instances. This is done with the `context` parameter. (For more complex properties, you can define `get_context` and `set_context` separately if needed.) Most properties will also have a default value.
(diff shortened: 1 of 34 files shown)


docs/README-develop-node-utils-APIs.md
```diff
@@ -159,6 +159,7 @@ The following basic command_reference parameters will be defined for each resour
  4. `default_value:` This is typically the "factory" default state of the property, expressed as an actual value (true, 12, "off", etc)
  5. `kind:` The data type of this property. If omitted, the property will be a string by default. Commonly used values for this property are `int` and `boolean`.
  6. `multiple:` By default a property is assumed to be found once or not at all by the `get_command`/`get_value` lookup, and an error will be raised if multiple matches are found. If multiple matches are valid and expected, you must set `multiple: true` for this property.
+ 7. `os_version:` By default a property is assumed to be supported on all version of software. If it is supported starting from a particular version, then this field can be used to specify the version for a particular product_id.
```
mikewiebe: I would reword as follows:
```
This command_reference parameter is used to specify the product family and minimum os version within that product family required to support the property.  Any version that is greater then or equal to the specified os version within the product family will be supported.
```

lib/cisco_node_utils/cmd_ref/stp_global.yaml
```diff
@@ -78,7 +78,8 @@ bridge_assurance:
   default_value: true
 
 domain:
-  _exclude: [N3k, N9k-F, N9k]
+  _exclude: [N9k-F]
+  os_version: 'N3k:7.0.3.I6.1, N9k:7.0.3.I6.1'
```
mikewiebe: It would be nice if we could support either of the following:
`os_version: 'N3k:7.0.3.I6.1, N9k:7.0.3.I6.1'`
or
`os_version: 'N3k, N9k:7.0.3.I6.1`

tests/ciscotest.rb
```diff
@@ -172,6 +172,34 @@ def validate_property_excluded?(feature, property)
...
+    match = nil
+    hash.keys.each do |prid|
+      match = prid if pid.match(Regexp.new(prid.upcase))
+    end
+    return if match.nil?
+    ver = hash[match]
+    lim = im.split[0].tr('(', '.').tr(')', '.').chomp('.')
+    # due to a bug in Gem::Version, we need to append a letter
+    # to the version field if the to be compared version
+    # has a letter at the end
```
mikewiebe: Please add more specific examples in the comments with actual version numbers for reference.

docs/README-develop-node-utils-APIs.md
mikewiebe: Also please indicate that this is a `test_only` feature and is ignored outside the context of the minitests and must be used in conjunction with the `skip_incompat_version?` api.

lib/cisco_node_utils/node.rb
```diff
@@ -435,5 +435,10 @@ def boot
     def system
       config_get('show_version', 'system_image')
     end
+
+    def os_version_get(feature, property)
+      ref = @cmd_ref.lookup(feature, property)
```
mikewiebe: Could you simplify this to:
`@cmd_ref.lookup(feature, property).os_version`

tests/ciscotest.rb
```diff
@@ -172,6 +172,34 @@ def validate_property_excluded?(feature, property)
...
+    ver << 'a' if lim[-1, 1] =~ /[[:alpha:]]/
+    skip('software version not supported for this test') if
```
mikewiebe: I would reword this to add a bit more info:
```ruby
skip("The #{feature} #{property} property is only supported on OS version #{ver} or later on #{pid}")
```

tests/ciscotest.rb
```diff
@@ -172,6 +172,34 @@ def validate_property_excluded?(feature, property)
...
+
+    array = ov.split(',')
+    hash = {}
+    array.each do |e|
+      key_value = e.split(':')
+      hash[key_value[0].strip] = key_value[1]
+    end
+
+    match = nil
+    hash.keys.each do |prid|
```
mikewiebe: It would be good to use our `product_tag` api (https://github.com/cisco/cisco-network-node-utils/blob/develop/tests/ciscotest.rb#L466) as a way to make sure any product tags in the yaml file are supported.

This way we provide input validation if developers try to use a tag this not supported.

mikewiebe:
@saichint Done with the initial review.

mikewiebe:
@saichint Travis errors

PR #546, opened 2017-05-08
Span session fixes:

mikewiebe:
Summary of changes:
  * Completes the `source_interfaces` and `source_vlans` properties that were originally submitted under https://github.com/cisco/cisco-network-node-utils/pull/539
  * Add proper `_exclude` yaml entry for unsupported platforms

This update allows both `source_interfaces` and `source_vlans` to be set using a ruby hash which was requested by @tomcooperca.  The puppet manifest will use a hash instead of an array for these properties.

Tested on: N3k, N9k and N9k-F.  Verified to skip on N7k.

### DIFF
--- a/tests/test_span_session.rb
+++ b/tests/test_span_session.rb
@@ -89,7 +89,7 @@ class TestSpanSession < CiscoTestCase
     assert_equal(intlb.to_a.sort, span.source_interfaces.sort)
 
     # intla/c same size but 1 element different
-    intlc = { int2            => 'tx',
+    intlc = { int2            => 'both',
               int1            => 'rx',
               'port-channel1' => 'both',
               'sup-eth0'      => 'rx' }
--- a/tests/test_span_session.rb
+++ b/tests/test_span_session.rb
@@ -60,33 +60,95 @@ class TestSpanSession < CiscoTestCase
     assert_equal(span.description, desc)
   end
 
-  def test_session_source_interface
+  def test_session_source_interfaces
     span = SpanSession.new(4)
     po_int = Interface.new('port-channel1')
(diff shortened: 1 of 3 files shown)


saichint:
Looks good. Only a general question.
Does the source_vlans take overlapping ranges like %w(1 2-10 5-12) ? For ex. in stp_global, this was taken care in the puppet_module side. As long as CLI throws an error or puppet module takes care of this, it is fine. Or else this will create idempotent issue.

mikewiebe:
@saichint Yes, it does support overlapping ranges.  As you stated this will need to be handled in the puppet module to avoid the idempotence issue.

tests/test_span_session.rb
```diff
@@ -60,33 +60,95 @@ def test_session_description
...
+    intlb = { int1            => 'rx',
+              int2            => 'tx',
+              'port-channel1' => 'both',
+              'sup-eth0'      => 'rx' }
+
+    span.source_interfaces = intlb
+    assert_equal(intlb.to_a.sort, span.source_interfaces.sort)
+
+    # intla/c same size but 1 element different
+    intlc = { int2            => 'tx',
```
rahushen: Is this meant to check variable placement in a hash ?

saichint:
:+1:

rahushen:
@mikewiebe changes look good. Minor comment regarding one of the tests.

tests/test_span_session.rb
mikewiebe: Nice find. I updated the test and it caused me to find a bug in my code :).  Bug also fixed.

mikewiebe:
@rahushen Review comments addressed.

PR #547, opened 2017-05-08
Bgp af aa

saichint:
This PR is for adding a new provider bgp_af_aggr_addr which was requested in the enhancement https://github.com/cisco/cisco-network-puppet-module/issues/452

Also couple of methods "attach_prefix" and "extract_value" have been moved to Utils as they are being used in many providers and same code is getting copied in many places. 

All tests pass on all platforms.

### DIFF
--- a/lib/cisco_node_utils/cisco_cmn_utils.rb
+++ b/lib/cisco_node_utils/cisco_cmn_utils.rb
@@ -368,8 +368,17 @@ module Cisco
       value
     end # add_quotes
 
-    # extract value of property
-    def self.extract_value(prop, prefix=nil, match_method)
+    # This method is used in config_get for CLIs which
+    # have multiple properties in the same output
+    # Given a match_method which defines regex pattern and the
+    # match criteria, this method, extracts the value of
+    # a property with or without a prefix.
+    # For ex. if the regex pattern is something like:
+    # (?<action>\S+)?
+    # the property to extract is action and prefix is nil
+    # for (?<src_port>range \S+)
+    # the property to extract is src_port and prefix is range
+    def self.extract_value(match_method, prop, prefix=nil)
       prefix = prop if prefix.nil?
       mm = match_method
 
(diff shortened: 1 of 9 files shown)


lib/cisco_node_utils/cisco_cmn_utils.rb
```diff
@@ -367,5 +367,27 @@ def self.add_quotes(value)
         value.start_with?('"') && value.end_with?('"')
       value
     end # add_quotes
+
+    # extract value of property from aggregate-address
```
mikewiebe: This is a general purpose method right?  Should you use a more generic comment that does not mention aggregate-address?  

Thanks for moving these into `cisco_cmd_utils` so they can be used by any objects that need them.

saichint: Forgot to clean comments when moving from bgp_af_aggr_addr to cisco_cmn_utils

lib/cisco_node_utils/cisco_cmn_utils.rb
```diff
@@ -367,5 +367,27 @@ def self.add_quotes(value)
         value.start_with?('"') && value.end_with?('"')
       value
     end # add_quotes
+
+    # extract value of property
```
mikewiebe: IMO these utility methods need more documentation.  It would be good to provide some examples that show how/why to use these for developers that are trying to understand the various methods in the library.

lib/cisco_node_utils/cisco_cmn_utils.rb
```diff
@@ -367,5 +367,27 @@ def self.add_quotes(value)
...
+    def self.extract_value(prop, prefix=nil, match_method)
```
mikewiebe: If you switch the argument order of `prefix` and `match_method` you don't have to fill in the `nil` value when calling the method as follows.  Typically args with defaults should be at the end anyway.

This:
```ruby
val = Utils.extract_value('control', nil, probe_get)
```
Becomes:
```ruby
val = Utils.extract_value('control', probe_get)
```

tests/test_bgp_af_aa.rb
```diff
@@ -0,0 +1,108 @@
...
+# limitations under the License.
+
+require_relative 'ciscotest'
+require_relative '../lib/cisco_node_utils/bgp_af'
+require_relative '../lib/cisco_node_utils/bgp_af_aggr_addr'
+
+include Cisco
+# TestBgpAfAggrAddr - Minitest for BgpAFAggrAddr class
+class TestBgpAfAggrAddr < CiscoTestCase
+  @@pre_clean_needed = true # rubocop:disable Style/ClassVars
```
mikewiebe, line 22: Do you need to skip any platforms for this or is it supported on n3|7|9k platforms?

saichint: Works on all platforms including n5|6k

lib/cisco_node_utils/cisco_cmn_utils.rb
saichint: will move the match_method to 1st arg because prop and prefix should be together.

mikewiebe:
@saichint Just a couple of minor comments.  Otherwise looks good.

saichint:
Done with changes and tested.

PR #551, opened 2017-06-02
Object_group

saichint:
This PR is for adding node utils for object_group and object_group_entry. object_group is just for creating an object_group as a parent for object_group_entry. It has no properties. object_group_entry is similar to cisco_ace provider and it has 2 properties for creating ip address and port entries.
All minitest pass.

### DIFF
--- a/tests/test_object_group.rb
+++ b/tests/test_object_group.rb
@@ -1,0 +1,125 @@
+# Copyright (c) 2017 Cisco and/or its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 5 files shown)


tests/test_object_group.rb
```diff
@@ -0,0 +1,125 @@
...
+    @og_name_port = 'test-my-port'
+    no_og_list_my
+  end
+
+  def teardown
+    no_og_list_my
+    super
+  end
+
+  def no_og_list_my
```
mikewiebe: I think it would be simpler/cleaner to get the objects and call the destroy method here:

Here is a better example 
https://github.com/cisco/cisco-network-node-utils/blob/develop/tests/ciscotest.rb#L256

mikewiebe:
@saichint Looks good.  Just a minor comment in the test cleanup.

mikewiebe:
👍  Looks good.

PR #553, opened 2017-07-05
Fix bgp, ospf, and route map testcases

rahushen:
This PR is for fixing minitest errors on greenboro.

1. BGP, OSPF show commands fail due to feature not enabled.
2. Protocol order not maintained in running config for route maps.
3. additional properties added to BGP AF in I7.

All required tests pass


saichint:
:+1:

PR #555, opened 2017-08-14
cisco_ace fix

saichint:
This PR is for fixing ace issue:
when remark property is being set, log and established should be nil.
Added validation on the puppet module side as well.
All minitests pass


mikewiebe:
@saichint Changes look good to me.  Requested a review from @shermdog before merging.

shermdog (commented):
Looks good to me.  It would be nice to see a test in there to verify established and log are nil for remarks though.

PR #559, opened 2017-09-14
Bridge domain fixes

saichint:
This PR is for fixing bridge domain. Previously the code was depending on a space beyond the bridge-domain to differentiate between two strings of bridge-domain which is error prone. This PR corrects the problem by using show bridge-domain summary and show bridge-domain <bridge-domain id> commands.
The tests all pass on old and new versions of software releases.
Tested beaker as well and it is fine.

### DIFF
--- a/lib/cisco_node_utils/bridge_domain.rb
+++ b/lib/cisco_node_utils/bridge_domain.rb
@@ -129,7 +129,6 @@ module Cisco
     end
 
     def bd_name=(str)
-      str = '' unless str
       state = str.empty? ? 'no' : ''
       config_set('bridge_domain', 'bd_name', bd: @bd_ids, state: state,
                    name: str)
--- a/lib/cisco_node_utils/bridge_domain.rb
+++ b/lib/cisco_node_utils/bridge_domain.rb
@@ -124,15 +124,14 @@ module Cisco
     #                      PROPERTIES                      #
     ########################################################
 
-    # Bridge-Domain name assigning case
-    # bridge-domain 100
-    #   name bd100
     def bd_name
       match = config_get('bridge_domain', 'bd_name', bd: @bd_ids)
       match.each do |line|
         next unless line.include?('Name')
         regexp = Regexp.new('Name::\s(\S+)')
         name = regexp.match(line)
+        # if name is not set, it will be Bridge-Domain followed by bd
+        # like Bridge-domain101 for bd 101
         bname = name[1].include?('Bridge-Domain') ? default_bd_name : name[1]
         return bname
       end
@@ -149,9 +148,6 @@ module Cisco
       config_get_default('bridge_domain', 'bd_name')
     end
 
-    # Bridge-Domain type change to fabric-control
-    # bridge-domain 100
-    #   fabric-control
     # This type property can be defined only for one bd
     def fabric_control
       match = config_get('bridge_domain', 'fabric_control', bd: @bd_ids)
@@ -174,9 +170,6 @@ module Cisco
       config_get_default('bridge_domain', 'fabric_control')
     end
 
-    # Bridge-Domain Shutdown case
-    # bridge-domain 100
-    #   shutdown
     def shutdown
       match = config_get('bridge_domain', 'shutdown', bd: @bd_ids)
       match.each do |line|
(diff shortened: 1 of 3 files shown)


lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -124,11 +124,17 @@ def destroy
     #                      PROPERTIES                      #
     ########################################################
 
-    # Bridge-Domain name assigning case
-    # bridge-domain 100
-    #   name bd100
     def bd_name
-      config_get('bridge_domain', 'bd_name', bd: @bd_ids)
+      match = config_get('bridge_domain', 'bd_name', bd: @bd_ids)
```
mikewiebe: The getters for `bd_name` and `fabric_control` can be significantly simplified by using structured output.

For example:
```yaml
bd_name:
  kind: string
  get_data_format: nxapi_structured
  get_command: "show bridge-domain <bd>"
  get_context: ["TABLE_bd", "ROW_bd"]
  get_value: ["bd-id <bd>", "bd-name"]
  set_value: "<state> name <name>"
  default_value: ~
```

```ruby
def bd_name
  config_get('bridge_domain', 'bd_name', bd: @bd_ids)
end
```

I verified these changes work on both `helsinki` and `beverly-hills`

**Update:** `get_value:` can be further simplified as follows since you are filtering the show command using `<bd>`

`get_value: "bd-name"`

mikewiebe: **Note:** This will cause two of the tests to fail with the following:

```shell
Node under test:
  - name  - n7k-j
  - type  - N7K-C7009
  - image - bootflash:///n7000-s2-dk9.7.3.0.D1.1.bin

TestBridgeDomain#test_bd_name = 5.96 s = F

Finished in 7.861067s, 0.1272 runs/s, 0.1272 assertions/s.

  1) Failure:
TestBridgeDomain#test_bd_name [tests/test_bridge_domain.rb:109]:
Error: Bridge-Domain name not initialized to default.
Expected: nil
  Actual: "Bridge-Domain101"
```

This test needs to be fixed.  We should not be testing for `nil` here but rather the default name.

lib/cisco_node_utils/bridge_domain.rb
```diff
@@ -142,12 +148,17 @@ def default_bd_name
       config_get_default('bridge_domain', 'bd_name')
     end
 
-    # Bridge-Domain type change to fabric-control
-    # bridge-domain 100
-    #   fabric-control
     # This type property can be defined only for one bd
     def fabric_control
-      config_get('bridge_domain', 'fabric_control', bd: @bd_ids)
+      match = config_get('bridge_domain', 'fabric_control', bd: @bd_ids)
```
mikewiebe, line 143: You could probably also simplify this using structured output but I will let you work out the details.

mikewiebe:
@saichint Done with my first round of review.

lib/cisco_node_utils/bridge_domain.rb
saichint: done

lib/cisco_node_utils/bridge_domain.rb
saichint: I still did a small processing inside the object file rather than fixing the test code. The reason is that the default value is nil in the old code and this could create a backward compatibility issue. The other reason is that the return value is either the set value or "Bridge-domain" followed by bd. So the get will always return the "Bridge-domain" followed by bd if the name is not set. This can cause problems in the puppet module side as the default value is still nil there. So idempotent issues will arise unless we fix the provider code also. I don't see value in changing the code in multiple places. So just returning nil if the get value is "Bridge-domain" followed by bd, is simple one line change and will work just like before.

mikewiebe: I suppose it's better to normalize to something we control rather then let the box decide in this case.

mikewiebe:
Few more comments.

lib/cisco_node_utils/bridge_domain.rb
mikewiebe: Actually, had a chance to think about this some more.  I still believe we should change the default behavior even if it means changing puppet.  Here is my rational.

1. There likely won't be a backwards compatibility issue because we see the same default `bd_name` behavior in helsinki and beverly_hills.
2. Even if the default name format changes, it's still a default name and the actual name of the bridge-domain so we should return that as the value.
3. If users try to use the default name as the actual name in the manifest, it will get returned as nil.

saichint:
Done with my changes for review comments

mikewiebe:
👍

PR #560, opened 2017-09-22
remove ingress replication for n7k

saichint:
This PR is for getting rid of ingress replication for n7k as it is not supported by the platform. Another PR for puppet module will be opened later.


saichint:
UPDATE: tested with beaker and opening a PR for puppet module shortly. No more changes here.

mikewiebe:
👍

PR #562, opened 2017-10-04
Add default state property to interface object

mikewiebe:
**This updates adds the following:**
  * Class variable `@state_default` used to indicate weather or not interface is in a default state. This information is only tracked for ethernet interfaces and excludes other types of interfaces that can be created and destroyed (vlan, loopback, sub-interfaces, port-channels etc...)
  * Refactor of the `self.interfaces` method to issue the `show run interfaces` command once and gather default and non-default state info for each interface from the single show command.
  * Fixes to various interface properties exposed by this change.

**NOTE:**
These changes are part set of diffs in the cisco puppet module that allow the `cisco_interface` type to treat physical and logical interfaces as truly `ensurable` resources.  Physical interfaces can now be created and destroyed logically so that they can be managed in the same way as a logical interfaces.

Puppet Changes: https://github.com/cisco/cisco-network-puppet-module/pull/470

**Tested on:**
N9k(I2, I6 and I7), N3k(I6), N7k(D1 and 8.2)

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/interface.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface.yaml
@@ -40,6 +40,13 @@ create:
   set_context: ~
   set_value: "interface <name>"
 
+default:
+  multiple:
+  set_context: ~
+  set_value:   "default interface <name>"
+  get_command: "show running interface <name>"
+  get_value:   '/(.*)/'
+
 description:
   kind: string
   get_value: '/^description (.*)/'
--- a/lib/cisco_node_utils/cmd_ref/interface.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface.yaml
@@ -18,8 +18,11 @@ all_interfaces:
   multiple:
   ios_xr:
     get_command: 'show running all-interfaces'
+    get_value: '/^interface (.*)/'
+  nexus:
+    get_command: "show running-config interface | section '^interface'"
+    get_value: '/(.*)/'
   get_context: ~
-  get_value: '/^interface (.*)/'
 
 bfd_echo:
   _exclude: [ios_xr]
--- a/lib/cisco_node_utils/cmd_ref/interface_ospf.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface_ospf.yaml
@@ -6,6 +6,11 @@ _template:
   get_command: 'show running interface all'
   context: ["interface %s"]
 
+all_interfaces:
+  multiple:
+  get_context: ~
+  get_value: '/^interface (.*)/'
+
 area:
   get_value: '/^\s*ip router ospf (\S+) area (\S+)/'
   set_value: "%s ip router ospf %s area %s"
--- a/lib/cisco_node_utils/cmd_ref/interface_portchannel.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface_portchannel.yaml
@@ -7,6 +7,11 @@ _template:
   get_context: ['/^interface %s$/i']
   set_context: ["interface %s"]
 
+all_interfaces:
+  multiple:
+  get_context: ~
+  get_value: '/^interface (.*)/'
+
 bfd_per_link:
   kind: boolean
   get_value: '/^bfd per-link$/'
--- a/lib/cisco_node_utils/interface.rb
+++ b/lib/cisco_node_utils/interface.rb
@@ -141,7 +141,16 @@ module Cisco
     end
 
     def destroy
-      config_set('interface', 'destroy', name: @name)
+      if @name[/ethernet/]
+        config_set('interface', 'default', name: @name)
+      else
+        config_set('interface', 'destroy', name: @name)
+      end
+    end
+
+    def default?
+      state = config_get('interface', 'default', name: @name)
+      state.nil? ? true : false
     end
 
     def pvlan_enable
--- a/lib/cisco_node_utils/interface.rb
+++ b/lib/cisco_node_utils/interface.rb
@@ -1,6 +1,6 @@
 # November 2015, Chris Van Heuveln
 #
-# Copyright (c) 2015-2016 Cisco and/or its affiliates.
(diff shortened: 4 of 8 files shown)


mikewiebe:
@shermdog Still a WIP but wanted to get your eyes on it early if you have any comments.

mikewiebe:
@saichint @shermdog I have finished the testing/coding.  Just need a code review.

lib/cisco_node_utils/cmd_ref/interface_ospf.yaml
```diff
@@ -6,6 +6,11 @@ _template:
   get_command: 'show running interface all'
   context: ["interface %s"]
 
+all_interfaces:
```
saichint, line 9: Do we need this? We have the same property in interface.yaml (all_interfaces). can it be reused here?

lib/cisco_node_utils/cmd_ref/interface_portchannel.yaml
```diff
@@ -7,6 +7,11 @@ _template:
   get_context: ['/^interface %s$/i']
   set_context: ["interface %s"]
 
+all_interfaces:
```
saichint, line 10: see my comment above. applies here as well

lib/cisco_node_utils/interface.rb
```diff
@@ -132,6 +152,9 @@ def self.capabilities(intf, mode=:hash)
     def create
       feature_vlan_set(true) if @name[/(vlan|bdi)/i]
       config_set('interface', 'create', name: @name)
+      if @name[/ethernet/] && !@name[/ethernet.*\.\d+/]
```
saichint: This seems odd. Are we changing a property just to make it non-default even though the manifest may not specify it?

lib/cisco_node_utils/cmd_ref/interface_ospf.yaml
saichint: update: seems like you need it as you changed the object code to call the new yaml properties. what is the reason here? The yaml seems pretty much the same, may be something is missing. could you add a comment?

saichint:
@mikewiebe Couple of minor questions in the review as it was not clear what the need for all_interfaces and changing the description. if you can add some comments explaining them, then it is fine.

lib/cisco_node_utils/cmd_ref/interface_ospf.yaml
mikewiebe: There is a pretty big difference now.  The `self.all_interfaces` method in `interface.rb` uses the `show running-config interface | section '^interface'` command along with a filter `get_value: '/(.*)/'` that gets all of the interfaces and associated interface data.  We need all of the data so that we can keep track of all interfaces and weather or not the interface is in a default state.

For the other objects `interface_portchannel.rb` and `interface_ospf.rb` we only need a simple list of interface names and not the additional data under each interface.

lib/cisco_node_utils/interface.rb
mikewiebe: Yes, I added this so that creating an interface for physical interfaces and setting properties to default values is idempotent.  This will NOTE get called if the manifest does not specify it. I needed at least one property to be set to a non-default value under the physical interface for it to be in a non-default state when created.  Without this if the ethernet interface is in the `absent` state (default state) and we try and create an interface via `ensure => present` and/or set properties under the interface to `default` values it will not be idempotent.

mikewiebe: This seemed like a fairly harmless way to allow us to manage physical interfaces and maintain idempotence, but I plan to demo this to the customer and get there 👍  or 👎 before keeping it.

saichint:
:+1:

lib/cisco_node_utils/cmd_ref/interface.yaml
```diff
@@ -18,8 +18,11 @@ all_interfaces:
   multiple:
   ios_xr:
     get_command: 'show running all-interfaces'
+    get_value: '/^interface (.*)/'
+  nexus:
+    get_command: "show running-config interface | section '^interface'"
+    get_value: '/(.*)/'
   get_context: ~
-  get_value: '/^interface (.*)/'
```
chrisvanheuveln, line 22: @mikewiebe  `section` causes some issues for some providers; e.g. `interface_service_vni`. See my other comment below.

lib/cisco_node_utils/interface.rb
```diff
@@ -72,10 +79,23 @@ def self.interfaces(opt=nil)
       intf_list = config_get('interface', 'all_interfaces')
       return hash if intf_list.nil?
 
+      # Massage intf_list data into an array that is easy
+      # to work with.
+      intf_list.collect! { |x| x.strip || x }
+      intf_list.delete('')
+      intf_list = intf_list.join(' ').split('interface')
+      intf_list.delete('')
```
chrisvanheuveln, line 87: @mikewiebe This massage doesn't work right with the section option. 

```
      intf_list = node.config_get('interface', 'all_interfaces')
# => ["interface Ethernet5/48", "", "  service instance 22 vni", "", ""]

      intf_list.collect! { |x| x.strip || x }
      intf_list.delete('')
      intf_list = intf_list.join(' ').split('interface')
      intf_list.delete('')
      puts intf_list
# => [" Ethernet5/48 service instance 22 vni"]
```

There are also other NU providers that use `all_interfaces` (e.g. `interface_service_vni`) so this list massage needs to occur for all of those providers.

chrisvanheuveln: I was a little hasty. This code looks okay here but `interface_service_vni` needs similar massage code.

lib/cisco_node_utils/cmd_ref/interface.yaml
chrisvanheuveln: @mikewiebe per IRC I'm going to add a separate `all_interfaces` getter specific to vni. That seems to be the only provider having issues.

PR #563, opened 2017-10-10
Improve performance

saichint:
This PR is for improving node utils performance while getting the facts.
1. "Show version" is very slow command, replaced with "Show hardware"
2. Unnecessary http requests removed
3. Use "show system uptime" to get uptime instead of "show version" which is slow.

Tested on N3k, N7k and N9k. there are few errors which are unrelated to this PR.


mikewiebe:
👍

PR #564, opened 2017-10-24
N3600 support

saichint:
1. This PR is for adding support to N3600. 
2. Event_history properties for bgp are now version independent

Tests run on n3k, n7k, n9k and n9k-f and they pass

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/snmp_server.yaml
+++ b/lib/cisco_node_utils/cmd_ref/snmp_server.yaml
@@ -63,7 +63,11 @@ packet_size:
     default_value: 0
   N6k: *n5k_default_packet_size
   N7k: *n5k_default_packet_size
+  # On recent versions, packet size is fixed,
+  # so this change will take care of the fix but
+  # it could still fail in the older versions.
   N9k-F: *n3k_default_packet_size
+  N3k-F: *n3k_default_packet_size
 
 protocol:
   _exclude: [ios_xr]
--- a/lib/cisco_node_utils/cmd_ref/snmp_server.yaml
+++ b/lib/cisco_node_utils/cmd_ref/snmp_server.yaml
@@ -63,7 +63,7 @@ packet_size:
     default_value: 0
   N6k: *n5k_default_packet_size
   N7k: *n5k_default_packet_size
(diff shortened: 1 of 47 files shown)


lib/cisco_node_utils/cmd_ref/snmp_server.yaml
```diff
@@ -63,7 +63,7 @@ packet_size:
     default_value: 0
   N6k: *n5k_default_packet_size
   N7k: *n5k_default_packet_size
-  N9k-F: *n5k_default_packet_size
+  N9k-F: *n3k_default_packet_size
```
mikewiebe: Have you tested this change on both the original N9K-Fretta images and the latest?  I suspect it will break older image tests.

lib/cisco_node_utils/node.rb
```diff
@@ -372,6 +372,15 @@ def prod_qualifier(prod, inventory)
...
         end
+      when /N3K/
+        if @cmd_ref
+          ver = os_version
+        else
+          ver = get(command:     'show version',
+                    data_format: :nxapi_structured)['kickstart_ver_str']
+        end
+        # Append -F for fretta platform.
+        return 'N9k-3-F' if ver[/7.0\(3\)F/]
```
mikewiebe: I don't understand this change.  Why are you adding a `-3` in the middle?

tests/test_interface.rb
```diff
@@ -629,7 +629,7 @@ def test_mtu_invalid_loopback
   end
 
   def test_speed
-    skip_legacy_defect?('7.0.3.I5.2', 'CSCvd41419')
+    skip_legacy_defect?('7.0.3.I5|7.0.3.F3', 'CSCvd41419')
```
mikewiebe, line 632: Is this defect still present in `7.0.3.I5.3`?

mikewiebe:
@saichint Done with my initial review comments.

tests/test_interface.rb
saichint: yes

saichint:
Done with all changes. Added N3k-F to the product_id for the N3600 series. All tests pass amd skips are valid.

mikewiebe:
👍

PR #565, opened 2017-11-01
Feature/evpn multisite

rahushen:
- Added new evpn_multisite NU module
- Added new evpn_stormcontrol NU module
- Added new interface_evpn_multisite NU module
- Added new properties to vxlan_vtep, vxlan_vtep_vni, bgp_neighbor, bgp_neighbor_af
- Added tests for above changes

### DIFF
--- a/lib/cisco_node_utils/bgp_neighbor.rb
+++ b/lib/cisco_node_utils/bgp_neighbor.rb
@@ -504,4 +504,24 @@ module Cisco
       end
     end
   end # class
+
+  def default_peer_type
+    config_get_default('bgp_neighbor', 'peer_type')
+  end
+
+  def peer_type=(val)
+    if val == default_peer_type
+      @set_args[:state] = 'no'
+    else
+      Feature.nv_overlay_evpn_enable unless
+        Feature.nv_overlay_evpn_enabled?
+      @set_args[:peer_type] = val
+      @set_args[:state] = ''
+    end
+    config_set('bgp_neighbor', 'peer_type', @set_args)
+  end
+
+  def peer_type
+    config_get('bgp_neighbor', 'peer_type', @get_args)
+  end
 end # module
--- a/lib/cisco_node_utils/bgp_neighbor.rb
+++ b/lib/cisco_node_utils/bgp_neighbor.rb
@@ -503,25 +503,25 @@ module Cisco
         fail KeyError
       end
     end
-  end # class
 
-  def default_peer_type
-    config_get_default('bgp_neighbor', 'peer_type')
-  end
-
-  def peer_type=(val)
-    if val == default_peer_type
-      @set_args[:state] = 'no'
-    else
-      Feature.nv_overlay_evpn_enable unless
-        Feature.nv_overlay_evpn_enabled?
-      @set_args[:peer_type] = val
-      @set_args[:state] = ''
+    def default_peer_type
+      config_get_default('bgp_neighbor', 'peer_type')
     end
-    config_set('bgp_neighbor', 'peer_type', @set_args)
-  end
 
-  def peer_type
-    config_get('bgp_neighbor', 'peer_type', @get_args)
-  end
+    def peer_type=(val)
+      if val == default_peer_type
+        @set_args[:state] = 'no'
+      else
+        Feature.nv_overlay_evpn_enable unless
+          Feature.nv_overlay_evpn_enabled?
+        @set_args[:peer_type] = val
+        @set_args[:state] = ''
+      end
+      config_set('bgp_neighbor', 'peer_type', @set_args)
+    end
+
+    def peer_type
+      config_get('bgp_neighbor', 'peer_type', @get_args)
+    end
+  end # class
 end # module
--- a/lib/cisco_node_utils/bgp_neighbor.rb
+++ b/lib/cisco_node_utils/bgp_neighbor.rb
@@ -512,8 +512,7 @@ module Cisco
       if val == default_peer_type
         @set_args[:state] = 'no'
       else
-        Feature.nv_overlay_evpn_enable unless
-          Feature.nv_overlay_evpn_enabled?
+        Feature.nv_overlay_evpn_enable
         @set_args[:peer_type] = val
         @set_args[:state] = ''
       end
--- a/lib/cisco_node_utils/bgp_neighbor_af.rb
+++ b/lib/cisco_node_utils/bgp_neighbor_af.rb
@@ -756,5 +756,22 @@ module Cisco
     def default_weight
       config_get_default('bgp_neighbor_af', 'weight')
     end
+
+    def rewrite_rt_asn
+      config_get('bgp_neighbor_af', 'rewrite_rt_asn', @get_args)
+    end
+
+    def rewrite_rt_asn=(state)
+      if state
+        Feature.nv_overlay_evpn_enable unless
+          Feature.nv_overlay_evpn_enabled?
+      end
+      set_args_keys(state: (state ? '' : 'no'))
+      config_set('bgp_neighbor_af', 'rewrite_rt_asn', @set_args)
+    end
+
+    def default_rewrite_rt_asn
+      config_get_default('bgp_neighbor_af', 'rewrite_rt_asn')
+    end
   end
 end
--- a/lib/cisco_node_utils/bgp_neighbor_af.rb
+++ b/lib/cisco_node_utils/bgp_neighbor_af.rb
@@ -762,10 +762,7 @@ module Cisco
     end
 
     def rewrite_rt_asn=(state)
-      if state
-        Feature.nv_overlay_evpn_enable unless
-          Feature.nv_overlay_evpn_enabled?
-      end
+      Feature.nv_overlay_evpn_enable if state
       set_args_keys(state: (state ? '' : 'no'))
       config_set('bgp_neighbor_af', 'rewrite_rt_asn', @set_args)
     end
--- a/lib/cisco_node_utils/cmd_ref/bgp_neighbor.yaml
+++ b/lib/cisco_node_utils/cmd_ref/bgp_neighbor.yaml
@@ -115,7 +115,7 @@ password_type:
     default_value: 0
 
 peer_type:
-  _exclude: [ios_xr, N3k, N5k, N6k, N7k, N9k-F]
+  _exclude: [ios_xr, N3k, N3k-F, N5k, N6k, N7k, N9k-F]
   kind: string
   get_value: '/^peer-type (\S+)$/'
   set_value: '<state> peer-type <peer_type>'
--- a/lib/cisco_node_utils/cmd_ref/bgp_neighbor.yaml
+++ b/lib/cisco_node_utils/cmd_ref/bgp_neighbor.yaml
@@ -115,6 +115,7 @@ password_type:
     default_value: 0
 
 peer_type:
+  _exclude: [ios_xr, N3k, N5k, N6k, N7k, N9k-F]
   kind: string
   get_value: '/^peer-type (\S+)$/'
   set_value: '<state> peer-type <peer_type>'
--- a/lib/cisco_node_utils/cmd_ref/bgp_neighbor.yaml
+++ b/lib/cisco_node_utils/cmd_ref/bgp_neighbor.yaml
@@ -114,6 +114,12 @@ password_type:
     get_value: '/^password (\d+)/'
     default_value: 0
 
+peer_type:
+  kind: string
+  get_value: '/^peer-type (\S+)$/'
+  set_value: '<state> peer-type <peer_type>'
+  default_value: ""
+
 remote_as:
   get_value: '/^remote-as (\d*?.?\d+?)$/'
   set_value: '<state> remote-as <remote_as>'
--- a/lib/cisco_node_utils/cmd_ref/bgp_neighbor_af.yaml
+++ b/lib/cisco_node_utils/cmd_ref/bgp_neighbor_af.yaml
@@ -146,6 +146,13 @@ prefix_list_out:
   set_value: '<state> prefix-list <str> out'
   default_value: ''
 
+rewrite_rt_asn:
+  _exclude: [ios_xr]
+  kind: boolean
+  get_value: '/^rewrite-rt-asn$/'
+  set_value: '<state> rewrite-rt-asn'
+  default_value: false
+
 route_map_in:
   kind: string
   default_value: ''
--- a/lib/cisco_node_utils/cmd_ref/bgp_neighbor_af.yaml
+++ b/lib/cisco_node_utils/cmd_ref/bgp_neighbor_af.yaml
@@ -147,7 +147,7 @@ prefix_list_out:
   default_value: ''
 
 rewrite_rt_asn:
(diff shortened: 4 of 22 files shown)


mikewiebe:
@rahushen Please make sure to update the changelog.

lib/cisco_node_utils/bgp_neighbor.rb
```diff
@@ -503,5 +503,25 @@ def mode_cli_to_symbol(cli)
...
+    def default_peer_type
+      config_get_default('bgp_neighbor', 'peer_type')
+    end
+
+    def peer_type=(val)
+      if val == default_peer_type
+        @set_args[:state] = 'no'
+      else
+        Feature.nv_overlay_evpn_enable unless
+          Feature.nv_overlay_evpn_enabled?
```
mikewiebe: You don't actually need the `unless Feature.nv_overlay_evpn_enabled?` bit.  The `Feature.nv_overlay_evpn_enable` takes care of this.

https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/feature.rb#L193

lib/cisco_node_utils/bgp_neighbor_af.rb
```diff
@@ -756,5 +756,22 @@ def weight=(int)
...
     end
+
+    def rewrite_rt_asn
+      config_get('bgp_neighbor_af', 'rewrite_rt_asn', @get_args)
+    end
+
+    def rewrite_rt_asn=(state)
+      if state
+        Feature.nv_overlay_evpn_enable unless
+          Feature.nv_overlay_evpn_enabled?
```
mikewiebe: Same comment here.

lib/cisco_node_utils/cmd_ref/bgp_neighbor.yaml
```diff
@@ -114,6 +114,13 @@ password_type:
     get_value: '/^password (\d+)/'
     default_value: 0
 
+peer_type:
+  _exclude: [ios_xr, N3k, N5k, N6k, N7k, N9k-F]
```
mikewiebe: Sai just added a key for `N3k-F` which needs to be added to your `_exclude`.

lib/cisco_node_utils/cmd_ref/bgp_neighbor_af.yaml
```diff
@@ -146,6 +146,13 @@ prefix_list_out:
   set_value: '<state> prefix-list <str> out'
   default_value: ''
 
+rewrite_rt_asn:
+  _exclude: [ios_xr, N3k, N5k, N6k, N7k, N9k-F]
```
mikewiebe: `N3k-F`

lib/cisco_node_utils/cmd_ref/evpn_multisite.yaml
```diff
@@ -0,0 +1,17 @@
+# evpn_multisite.yaml
+---
+_exclude: [ios_xr, N3k, N5k, N6k, N7k, N9k-F]
```
mikewiebe: `N3k-F`

lib/cisco_node_utils/cmd_ref/evpn_stormcontrol.yaml
```diff
@@ -0,0 +1,18 @@
+# evpn_stormcontrol.yaml
+---
+_exclude: [ios_xr, N3k, N5k, N6k, N7k, N9k-F]
```
mikewiebe: `N3k-F`

lib/cisco_node_utils/cmd_ref/interface_evpn_multisite.yaml
```diff
@@ -0,0 +1,12 @@
+# interface_evpn_multisite
+---
+_exclude: [ios_xr, N3k, N5k, N6k, N7k, N9k-F]
```
mikewiebe: `N3k-F` and everywhere else it's needed :)

lib/cisco_node_utils/cmd_ref/evpn_multisite.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+_template:
+  get_command: "show running-config"
```
mikewiebe: Are there default values for these properties?  Does `show running all` display the defaults for them?

lib/cisco_node_utils/cmd_ref/evpn_stormcontrol.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+_template:
+  get_command: "show running-config"
```
mikewiebe: Are there default values for these properties?  Does `show running all` display the defaults for them?

lib/cisco_node_utils/cmd_ref/interface_evpn_multisite.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+_template:
+  get_command: "show running"
```
mikewiebe: Are there default values for these properties?  Does `show running all` display the defaults for them?

lib/cisco_node_utils/evpn_multisite.rb
```diff
@@ -0,0 +1,75 @@
...
+                " 'Integer' object"
+      fail ArgumentError, err_msg unless multisiteid.is_a?(Integer) ||
+                                         multisiteid.is_a?(String)
+      @multisiteid = multisiteid.to_i
+      @get_args = @set_args = { multisiteid: @multisiteid }
+      create if instantiate
+    end
+
+    def create
+      EvpnMultisite.enable
```
mikewiebe: Why not just call `Feature.nv_overlay_enable` here?

lib/cisco_node_utils/evpn_multisite.rb
```diff
@@ -0,0 +1,75 @@
...
+      @set_args[:state] = ''
+      config_set('evpn_multisite', 'multisite', @set_args)
+    end
+
+    def destroy
+      @set_args[:state] = 'no'
+      config_set('evpn_multisite', 'multisite', @set_args)
+    end
+
+    # enable feature nv overlay
```
mikewiebe: Seems redundant to me.  Don't see value in a wrapper method.

lib/cisco_node_utils/evpn_multisite.rb
```diff
@@ -0,0 +1,75 @@
...
+    def self.enable
+      Feature.nv_overlay_enable
+    end
+
+    def self.multisite
```
mikewiebe: Why is this a class method instead of an instance method?

lib/cisco_node_utils/evpn_stormcontrol.rb
```diff
@@ -0,0 +1,81 @@
...
+      @set_args[:state] = ''
+      config_stormcontrol
+    end
+
+    def destroy
+      @set_args[:state] = 'no'
+      config_stormcontrol
+    end
+
+    def self.broadcast
```
mikewiebe, line 58: Same question for the following methods that are class vs. instance methods.  What is the reason?

lib/cisco_node_utils/evpn_stormcontrol.rb
```diff
@@ -0,0 +1,81 @@
...
+      config_get('evpn_stormcontrol', 'unicast')
+    end
+
+    def level=(level)
+      err_msg = "level must be either a 'String' or an" \
+                " 'Integer' object"
+      fail ArgumentError, err_msg unless level.is_a?(Integer) ||
+                                         level.is_a?(String)
+      level = level.to_i
+      @set_args[:level] = level
```
saichint: Can we just do @set_args[:level] = level.to_i

lib/cisco_node_utils/evpn_multisite.rb
```diff
@@ -0,0 +1,75 @@
...
+      config_get('evpn_multisite', 'multisite')
+    end
+
+    def multisite=(multisiteid)
+      err_msg = "multisiteid must be either a 'String' or an" \
+                " 'Integer' object"
+      fail ArgumentError, err_msg unless multisiteid.is_a?(Integer) ||
+                                         multisiteid.is_a?(String)
+      multisiteid = multisiteid.to_i
+      @set_args[:multisiteid] = multisiteid
```
saichint: Can we just do @set_args[:multisiteid] = multisiteid.to_i

lib/cisco_node_utils/cmd_ref/evpn_multisite.yaml
saichint: Is there a simple command like "show running-config evpn" or "show running-config | section evpn" or something like that. That way it reduces the amount of output this produces.

lib/cisco_node_utils/cmd_ref/evpn_stormcontrol.yaml
saichint: Same comment as above

lib/cisco_node_utils/cmd_ref/interface_evpn_multisite.yaml
saichint: If you have to use "show running-config" then it is fine but at least use the same command (here it is "show running"). There is no difference between these two at the platform level but our cache is per command based and both of these are seen as 2 different ones and so it will not use cache, which is not desirable.

lib/cisco_node_utils/evpn_stormcontrol.rb
```diff
@@ -0,0 +1,81 @@
...
+      fail ArgumentError, err_msg unless type_list.include?(stormcontrol_type)
+      @stormcontrol_type = stormcontrol_type
+
+      err_msg = "level must be either a 'String' or an" \
+                " 'Integer' object"
+      fail ArgumentError, err_msg unless level.is_a?(Integer) ||
+                                         level.is_a?(String)
+      @level = level.to_i
+      @get_args = @set_args = { stormcontrol_type: @stormcontrol_type,
+                                level:             @level }
```
saichint, line 39: do we need @level? I don't see anywhere it is being used. If not needed, you can remove that line and just do level: level.to_i in the @getargs.

lib/cisco_node_utils/bgp_neighbor.rb
rahushen: Removed the check

lib/cisco_node_utils/bgp_neighbor_af.rb
rahushen: Removed the check

lib/cisco_node_utils/cmd_ref/bgp_neighbor.yaml
rahushen: added

lib/cisco_node_utils/cmd_ref/bgp_neighbor_af.yaml
rahushen: added

lib/cisco_node_utils/cmd_ref/evpn_multisite.yaml
rahushen: added

lib/cisco_node_utils/cmd_ref/evpn_stormcontrol.yaml
rahushen: added

lib/cisco_node_utils/cmd_ref/interface_evpn_multisite.yaml
rahushen: added

lib/cisco_node_utils/cmd_ref/evpn_multisite.yaml
rahushen: No. They don't show up in running config if they're not configured.

lib/cisco_node_utils/cmd_ref/evpn_stormcontrol.yaml
rahushen: No. They don't show up in the running-config if not configured.

lib/cisco_node_utils/cmd_ref/interface_evpn_multisite.yaml
rahushen: No. They don't show up in the running-config if not configured.

lib/cisco_node_utils/evpn_multisite.rb
rahushen: Yes, modified it.

lib/cisco_node_utils/evpn_multisite.rb
rahushen: Agreed. Removed it.

lib/cisco_node_utils/evpn_multisite.rb
rahushen: This allows me to inspect if multisite is configured on the device without having to create an instance variable of class 'EvpnMultisite'. As discussed, this may cause problems with the autogen methods while developing provider code for this module, but I'll revisit this at that point if changes need to be made.

lib/cisco_node_utils/evpn_stormcontrol.rb
rahushen: This allows me to inspect if stormcontrol is configured on the device without having to create an instance variable of class 'EvpnStormcontrol'. As discussed, this may cause problems with the autogen methods while developing provider code for this module, but I'll revisit this at that point if changes need to be made.

lib/cisco_node_utils/evpn_stormcontrol.rb
rahushen: I modified it to this:

@level = level.to_i
@set_args[:level] = @level

The reason being that the instance variable is updated here as well as the then used to set the value on the device.

lib/cisco_node_utils/evpn_multisite.rb
rahushen: @multisiteid = multisiteid.to_i
@set_args[:multisiteid] = @multisiteid

lib/cisco_node_utils/cmd_ref/evpn_multisite.yaml
rahushen: Modified this to 'show running-config | section multisite'

lib/cisco_node_utils/cmd_ref/evpn_stormcontrol.yaml
rahushen: Modified this to 'show running-config | section storm-control'

lib/cisco_node_utils/cmd_ref/interface_evpn_multisite.yaml
rahushen: Modified this to 'show running interface all' based on the fact that all the other interface cmd_ref modules use that command.

lib/cisco_node_utils/evpn_stormcontrol.rb
rahushen: I modified some code to use @level. I think, in general, it is good to have the instance variables set incase we expand upon this module in the future.

mikewiebe:
👍 Will defer to @saichint for final approval.

saichint:
:+1:

PR #566, opened 2017-11-02
fix system_image calls

saichint:
This PR is for using image_version instead of system_image calls. The latter uses bin file name which could be anything and so prone to errors.
All minitests pass on n9k-f platform where the problem is seen.


mikewiebe:
👍

PR #567, opened 2017-12-04
fix for virtual platforms and tacacs_global

saichint:
This PR is for fixing minitest errors on virtual platforms and skipping keyformat 6 for I2 releases

### DIFF
--- a/tests/ciscotest.rb
+++ b/tests/ciscotest.rb
@@ -227,8 +227,7 @@ class CiscoTestCase < TestCase
   end
 
   def virtual_platform?
-    s = @device.cmd('show version')
-    s[/demo/] ? true : false
+    node.product_id[/N9K-NXOSV|N9K-9000v/] ? true : false
   end
 
   def interfaces
--- a/tests/ciscotest.rb
+++ b/tests/ciscotest.rb
@@ -226,6 +226,11 @@ class CiscoTestCase < TestCase
     skip(msg) if Utils.image_version?(Regexp.new(pattern))
   end
 
+  def virtual_platform?
+    s = @device.cmd('show version')
+    s[/demo/] ? true : false
+  end
+
   def interfaces
     unless @@interfaces
       # Build the platform_info, used for interface lookup
--- a/tests/test_tacacs_global.rb
+++ b/tests/test_tacacs_global.rb
@@ -72,13 +72,15 @@ class TestTacacsGlobal < CiscoTestCase
     assert_equal(7, global.key_format)
     assert_equal('"WAWY_NZB"', global.key)
 
-    # second key change - modify key to type6
-    key_format = 6
-    # Must use a valid type6 password: CSCvb36266
-    key = 'JDYkqyIFWeBvzpljSfWmRZrmRSRE8'
-    global.encryption_key_set(key_format, key)
-    assert_equal(key_format, global.key_format)
-    assert_equal("\"#{key}\"", global.key)
+    unless Platform.image_version[/I2/]
+      # second key change - modify key to type6
+      key_format = 6
+      # Must use a valid type6 password: CSCvb36266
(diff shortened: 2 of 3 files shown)


tests/ciscotest.rb
```diff
@@ -226,6 +226,11 @@ def skip_legacy_defect?(pattern, msg)
     skip(msg) if Utils.image_version?(Regexp.new(pattern))
   end
 
+  def virtual_platform?
```
mikewiebe, line 229: The `@product_id` instance variable already has what you need since it containts the PID from the `show inventory` output so you don't need to query the device with another show command for this case.

```
camden-nx-1# show inventory 
NAME: "Chassis",  DESCR: "NX-OSv Chassis"                        
PID: N9K-NXOSV           ,  VID: V02 ,  SN: 9AK1QGSS07J          
```

Suggest you rewrite the method as follows:

```ruby
def virtual_platform?
  @product_id[/N9K-NXOSV/] ? true : false
end

File: tests/test_tacacs_global.rb [hunk 2]
```diff
@@ -72,13 +72,15 @@ def test_tacacs_global
...
     assert_equal('"WAWY_NZB"', global.key)
 
-    # second key change - modify key to type6
-    key_format = 6
-    # Must use a valid type6 password: CSCvb36266
-    key = 'JDYkqyIFWeBvzpljSfWmRZrmRSRE8'
-    global.encryption_key_set(key_format, key)
-    assert_equal(key_format, global.key_format)
-    assert_equal("\"#{key}\"", global.key)
+    unless Platform.image_version[/I2/]
```
mikewiebe, line 75: Do we have a defect we can reference here and call  the `skip_legacy_method` instead for I2 images?

tests/ciscotest.rb
saichint: Not really. The PID field is not constant like you mentioned. 
On some platforms, it shows "N9K-NXOSV" like you mentioned, and on some others it is
""N9K-9000v". We cannot depend on this. If you are sure these are the only 2 variations I can do @product_id[/NXOSV|9000v/] ? true : false
but we have to be 100% sure. Can you check with your group and let me know?

tests/test_tacacs_global.rb
saichint: I will check again, I could not find any defect on topic. But even if we, how do you call skip_legacy_defect here? We want all the test code to run just above these 4 lines and all the code below as well. Skip simply gets out of the test right away, right?

tests/ciscotest.rb
mikewiebe: Hmm... that's unfortunate.  Let me check with Jusheng and get back to you.  I really want to avoid another box query if we can.

tests/test_tacacs_global.rb
mikewiebe: We would need to pull this out into a separate test.  Are we certain this is only a problem on I2?  If so then it's probably ok like it is.  We should be moving away from I2 image validation soon.

tests/ciscotest.rb
saichint: Ok, I am now checking both NXOSV and 9000v

tests/test_tacacs_global.rb
saichint: I am leaving this test as it is, since I2 is going away soon. No point in writing separate tests just for a short time. In the meantime it will fail if it runs against I2 but that's fine

mikewiebe: I did not mean to remove the I2 check.  I still think that is good.  No point in always clicking on the email just to see that it's the same problem over and over again.  If we are not going to fix it then we should leave the I2 check.  I only meant that we don't need to separate this out into it's own test.

saichint: Ok added it back. I got confused by your comment. Sorry!

mikewiebe:
👍

PR #568, opened 2017-12-04
Fix n56k support

mikewiebe:
I discovered that we inadvertently broke our support for `n5k` and `n6k` devices.  This fix addresses the following:

1. The structured output for `show inventory` is slightly different between platforms.
2. The `sh hardware` command is not supported on `N5k` and `N6k` platforms.

**NOTE:** Currently running tests on N6k and N9k to validate the changes.


saichint:
:+1:

PR #569, opened 2017-12-07
(CISCO-60) Do not enable TACACS feature at init

shermdog:
Prior to this commit `feature tacacs+` was being enabled simply by initilizing the providers.

Moved enabling of TACACS to appropriate setters and stopped trying to query status for resources when disabled.


mikewiebe:
👍 Looks good and tests pass against various targets.  I added one comment to https://github.com/cisco/cisco-network-puppet-module/pull/479 so will merge once that is addressed.

PR #571, opened 2017-12-13
Create version 1.9.0-dev

mikewiebe:
This updates `lib/cisco_node_utils/version.rb` to use version `1.9.0-dev` in the develop branch.  When the gem is built and installed it will make it more obvious that this is a pre-release development gem.


saichint:
:+1:

PR #572, opened 2017-12-14
Part 2: More modifications for evpn_multisite

rahushen:
After adding the provider code for puppet, a few alterations needed to be made from a design perspective.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/bgp_neighbor_af.yaml
+++ b/lib/cisco_node_utils/cmd_ref/bgp_neighbor_af.yaml
@@ -201,6 +201,7 @@ soft_reconfiguration_in:
     get_value: '/^soft-reconfiguration inbound(?: always)?/'
     set_value: '<state> soft-reconfiguration inbound <always>'
   N3k: *soft_recon_always
+  N3k-F: *soft_recon_always
   N9k-F: *soft_recon_always
   N9k: *soft_recon_always
   else:
--- a/lib/cisco_node_utils/cmd_ref/bgp_neighbor_af.yaml
+++ b/lib/cisco_node_utils/cmd_ref/bgp_neighbor_af.yaml
@@ -146,6 +146,13 @@ prefix_list_out:
   set_value: '<state> prefix-list <str> out'
   default_value: ''
 
+rewrite_rt_asn:
+  _exclude: [ios_xr]
+  kind: boolean
+  get_value: '/^rewrite-rt-asn$/'
+  set_value: '<state> rewrite-rt-asn'
+  default_value: false
+
 route_map_in:
   kind: string
   default_value: ''
--- a/lib/cisco_node_utils/cmd_ref/bgp_neighbor_af.yaml
+++ b/lib/cisco_node_utils/cmd_ref/bgp_neighbor_af.yaml
@@ -146,11 +146,11 @@ prefix_list_out:
   set_value: '<state> prefix-list <str> out'
   default_value: ''
 
-rewrite_rt_asn:
+rewrite_evpn_rt_asn:
   _exclude: [ios_xr, N3k, N3k-F, N5k, N6k, N7k, N9k-F]
   kind: boolean
-  get_value: '/^rewrite-rt-asn$/'
-  set_value: '<state> rewrite-rt-asn'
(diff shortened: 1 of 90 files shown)


lib/cisco_node_utils/evpn_multisite.rb
```diff
@@ -34,17 +34,31 @@ def initialize(multisiteid, instantiate=true)
     end
 
     def create
-      Feature.nv_overlay_enable
+      # There is an issue with EX platform using
+      # Feature.nv_overlay_enable generating error code '400'
+      # Using config_set for enabling nv overlay
+      config_set('feature', 'nv_overlay', state: '')
+      sleep 1
```
mikewiebe, line 41: Why do we need the sleep?

lib/cisco_node_utils/evpn_multisite.rb
```diff
@@ -34,17 +34,31 @@ def initialize(multisiteid, instantiate=true)
...
       @set_args[:state] = ''
       config_set('evpn_multisite', 'multisite', @set_args)
     end
 
     def destroy
       @set_args[:state] = 'no'
+      # HACK: set time to a dummy value
```
mikewiebe, line 48: What are we working around in this case?

lib/cisco_node_utils/evpn_stormcontrol.rb
```diff
@@ -55,16 +55,19 @@ def destroy
...
-
-    def self.multicast
-      config_get('evpn_stormcontrol', 'multicast')
+    # Create a hash of all stormcontrol instances
+    def self.stormcontrol
+      hash = {}
+      type_list = %w(broadcast multicast unicast)
+      type_list.each do |type|
+        level = config_get('evpn_stormcontrol', type)
+        hash[type] = EvpnStormcontrol.new(type, level, false) if level
```
mikewiebe, line 64: Based on the documentation on the puppet side, `level` is a property, not a parameter. If it's needed to uniquely identify the resource it should be a parameter for puppet.

lib/cisco_node_utils/evpn_multisite.rb
rahushen: for configuring evpn_multisite, the command is as follows:

evpn multisite border-gateway 50.

To unconfigure, the command 'no evpn multisite border-gateway' doesn't work. It needs a multisite id which can be any integer. Ideally, the user while writing the maniest needn't have a multisite id while uncofiguring (ensure => absent). Hence, the hack.

lib/cisco_node_utils/evpn_stormcontrol.rb
rahushen: As discussed, the level is not required to uniquely identify the resource.

lib/cisco_node_utils/evpn_multisite.rb
rahushen: I just used the logic for self.nv_overlay_enable in feature.rb. I think this is to allow the device to actually turn the feature on after processing the configuration and avoid timing issues.

rahushen:
@mikewiebe ... Thanks for the first round of reviews. I've made changes and the code is now ready for another round of review.

lib/cisco_node_utils/cmd_ref/bgp_neighbor_af.yaml
```diff
@@ -146,11 +146,11 @@ prefix_list_out:
   set_value: '<state> prefix-list <str> out'
   default_value: ''
 
-rewrite_rt_asn:
+rewrite_evpn_rt_asn:
   _exclude: [ios_xr, N3k, N3k-F, N5k, N6k, N7k, N9k-F]
   kind: boolean
-  get_value: '/^rewrite-rt-asn$/'
-  set_value: '<state> rewrite-rt-asn'
+  get_value: '/^rewrite-evpn-rt-asn$/'
```
mikewiebe, line 152: Does this work on both I7.1 and I7.2?

rahushen: Yes, it works across both releases.

mikewiebe:
👍

PR #573, opened 2018-01-10
fixes for nu minitest failures

saichint:
This PR is for fixing node_ext and radius_global minitest failures. 
1. reset_reason and time are not supported on virtual platforms
2. radius-server key type 6 is not supported on older platforms.


mikewiebe:
👍

PR #575, opened 2018-01-29
TRM - Evpn multicast

rahushen:
Add node_utils logic for TRM (evpn multicast and changes to vrf_af to support mvpn).

### DIFF
--- a/tests/test_evpn_multicast.rb
+++ b/tests/test_evpn_multicast.rb
@@ -32,6 +32,14 @@ class TestEvpnMulticast < CiscoTestCase
     config('no advertise evpn multicast')
   end
 
+  def teardown
+    # disable feature ngmvpn and advertise evpn multicast
+    # after each test
+    config('no feature ngmvpn')
+    config('no advertise evpn multicast')
+    super
+  end
+
   def test_create_and_destroy
     mc = EvpnMulticast.new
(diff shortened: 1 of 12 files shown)


tests/test_evpn_multicast.rb
```diff
@@ -0,0 +1,56 @@
...
+
+require_relative 'ciscotest'
+require_relative '../lib/cisco_node_utils/evpn_multicast'
+require_relative '../lib/cisco_node_utils/feature'
+
+# TestEvpnMulticast - Minitest for EvpnMulticast class
+class TestEvpnMulticast < CiscoTestCase
+  @skip_unless_supported = 'evpn_multicast'
+
+  def setup
```
mikewiebe, line 27: We have seen issues in subsequent tests if we don't include a `teardown` method.   Suggest you add one to remove `feature ngmvpn` and `advertise evpn multicast` when you are done.

Example:
https://github.com/cisco/cisco-network-node-utils/blob/develop/tests/test_acl.rb#L31

mikewiebe:
@rahushen Minor test comment, otherwise 👍

rahushen:
@mikewiebe... updated the PR with test changes and also added ip multicast NU piece.

mikewiebe:
👍

PR #581, opened 2018-05-07
Set host_reachability correctly when setting ingress_replication

mikewiebe:
**Summary:**
In later versions of Nexus, the vxlan vtep host reachability must be set properly before setting the ingress replication feature.

This update modified the `ingress_replication` setter to check the should intent for `protocol` and ensures that the vtep host reachability is set properly to support the protocol setting.

**Testing:**
Tests pass on n7k, n9k(latest, greensboro, freeport and camden), n9k-fretta)

### DIFF
--- a/lib/cisco_node_utils/vxlan_vtep_vni.rb
+++ b/lib/cisco_node_utils/vxlan_vtep_vni.rb
@@ -132,19 +132,13 @@ module Cisco
     def remove_add_ingress_replication(protocol)
       # Note: ingress-replication is not supported on all platforms.
       # Use to_s.empty check to also handle nil check.
-      if ingress_replication.to_s.empty?
-        set_host_reachability(@set_args[:name], protocol)
-        set_args_keys(state: '', protocol: protocol)
-        config_set('vxlan_vtep_vni', 'ingress_replication', @set_args)
-      else
-        # Sadly, the only way to change between protocols is to
-        # first remove the existing protocol.
+      unless ingress_replication.to_s.empty?
         set_args_keys(state: 'no', protocol: ingress_replication)
         config_set('vxlan_vtep_vni', 'ingress_replication', @set_args)
-        set_host_reachability(@set_args[:name], protocol)
-        set_args_keys(state: '', protocol: protocol)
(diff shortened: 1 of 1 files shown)


saichint:
👍

lib/cisco_node_utils/vxlan_vtep_vni.rb
```diff
@@ -112,17 +112,36 @@ def ingress_replication
...
+        fail "Protocol #{protocol} currently not supported"
+      end
+      VxlanVtep.new(vtep_name).host_reachability = host_reachability
+    end
+
     def remove_add_ingress_replication(protocol)
       # Note: ingress-replication is not supported on all platforms.
       # Use to_s.empty check to also handle nil check.
       if ingress_replication.to_s.empty?
+        set_host_reachability(@set_args[:name], protocol)
```
rahushen: Maybe move this before if/else as it is exercised in both code paths

rahushen:
@mikewiebe .. Minor comment. Otherwise 👍

lib/cisco_node_utils/vxlan_vtep_vni.rb
mikewiebe: Good suggestion.

mikewiebe: Actually.. I tested it and it fails if I move it out of the if/else block.  The reason is that the two lines in the else block have to be executed first.

rahushen: would this work?

```
unless ingress_replication.to_s.empty?
  set_args_keys(state: 'no', protocol: ingress_replication)
  config_set('vxlan_vtep_vni', 'ingress_replication', @set_args)
end
set_host_reachability(@set_args[:name], protocol)
set_args_keys(state: '', protocol: protocol)
config_set('vxlan_vtep_vni', 'ingress_replication', @set_args)
```

mikewiebe: Yup.  That works.  Thanks for the suggestion.

rahushen (approved):

PR #582, opened 2018-06-18
Address default behavior change

mikewiebe:
**Summary:**
In recent image versions, there has been a default behavior change for `overlay_spt_only`.  In older image versions the feature defaults to off or `false` and in recent image versions the feature is now on by default or `true`.

This update:
* Updates the `overlay_spt_only` property to use the `show fabric multicast globals` command instead of `show running` which is not reliable for this property.
* Adds a `kind: boolean` check for both `overlay_distributed_dr` and `overlay_spt_only` properties.
* Adds `auto_default: false` to the `overlay_spt_only` property to make it easier to handle the different default states for different image versions.
* Adds a version check in `default_overlay_spt_only` 
* Separates the tests for both of the properties to make the results more granular.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/ip_multicast.yaml
+++ b/lib/cisco_node_utils/cmd_ref/ip_multicast.yaml
@@ -7,12 +7,16 @@ _template:
 
 overlay_distributed_dr:
   context: ~
+  kind: boolean
   get_value: '^ip multicast overlay-distributed-dr$'
   set_value: "<state> ip multicast overlay-distributed-dr"
   default_value: false
 
 overlay_spt_only:
   context: ~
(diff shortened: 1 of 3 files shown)


saichint:
PR seems fine.
One general question I have:
This provider is exclusive to N9k. But is there a software version check for older software version to skip it (below I7)? I did not see but may be I am looking at wrong place?

mikewiebe:
Yes, but it's part of the `ngmvpn` feature.

https://github.com/cisco/cisco-network-node-utils/blob/74af32f1b5dfe2c968cc7497b685edcfe23fcb86/lib/cisco_node_utils/cmd_ref/feature.yaml#L57

saichint:
👍

PR #585, opened 2018-07-10
Fix CISCO-65 and CISCO-68

saichint:
This PR is for fixing the following issues:
1. https://tickets.puppetlabs.com/browse/CISCO-65
On n5k, if there are multiple vlans, the puppet resource ```cisco_vlan``` command is failing. This is due to the fact that, the ```show vlan br | json``` returns an integer for ```vlan_id``` and on all other platforms, it returns a string. Converting index to string solves it.
2. https://tickets.puppetlabs.com/browse/CISCO-68
If there are multiple tacacs servers like (192.168.1.1, 192.168.1.11 etc.) the regex pattern is matching both servers for 192.168.1.1 and the value for ```port``` attribute, it is returning is an array which is causing issue. Basically the regex pattern check is wrong which does not take the white space into account properly.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/tacacs_server_host.yaml
+++ b/lib/cisco_node_utils/cmd_ref/tacacs_server_host.yaml
@@ -48,7 +48,7 @@ hosts:
 port:
   default_value: 49
   kind: int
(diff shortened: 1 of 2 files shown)


mikewiebe:
👍 @saichint Looks good to me.

@shermdog Can you take a look?

PR #586, opened 2018-08-03
(CISCO-71) Add support for banner/motd

shermdog:
This will provide support for new netdev_stdlib banner/motd type.
puppetlabs/netdev_stdlib#40

### DIFF
--- a/tests/test_banner.rb
+++ b/tests/test_banner.rb
@@ -1,0 +1,85 @@
+#
+# Minitest for Banner class
+#
+# Copyright (c) 2014-2018 Cisco and/or its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative 'ciscotest'
+require_relative '../lib/cisco_node_utils/banner'
+
+# TestBanner - Minitest for Banner node utility.
+class TestBanner < CiscoTestCase
+  def setup
+    # setup runs at the beginning of each test
+    super
+    no_motd
+  end
+
+  def teardown
+    # teardown runs at the end of each test
+    no_motd
+    super
+  end
+
+  def no_motd
+    # Turn the feature off for a clean test.
+    config('no banner motd')
+  end
+
+  # TESTS
+
+  def test_single_motd
+    id = 'default'
+
+    banner = Cisco::Banner.new(id)
+    assert_includes(Cisco::Banner.banners, id)
+    assert_equal(Cisco::Banner.banners[id], banner)
+
+    assert_equal(banner.default_motd, Cisco::Banner.banners['default'].motd)
+    assert_equal(banner.default_motd, banner.motd)
+
+    banner.motd = 'Test banner!'
+    assert_equal(Cisco::Banner.banners['default'].motd,
+                 'Test banner!')
+    assert_equal(Cisco::Banner.banners['default'].motd,
+                 banner.motd)
+
+    banner.motd = nil
+    assert_equal(banner.default_motd, Cisco::Banner.banners['default'].motd)
+    assert_equal(banner.default_motd, banner.motd)
+  end
+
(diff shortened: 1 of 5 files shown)


mikewiebe:
@shermdog Are you done with this PR and ready for me to review?

shermdog:
@mikewiebe I think it's close now.  I'm going to get some beaker tests together for the puppet module and put a PR up for that so it can all be done in one fell swoop.

shermdog:
@mikewiebe this should be good to merge.  Added PR for puppet module.

tests/test_banner.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def no_motd
+    # Turn the feature off for a clean test.
+    config('no banner motd')
+  end
+
+  # TESTS
+
+  def test_motd
```
mikewiebe: This test fails on my devices.

```shell
Node under test:
  - name    - n9k
  - type    - N9K-C9504
  - image   - bootflash:///nxos.9.2.1.bin
  - version - 9.2(1)

TestBanner#test_motd = 3.62 s = F

Finished in 3.625701s, 0.2758 runs/s, 1.1032 assertions/s.

  1) Failure:
TestBanner#test_motd [tests/test_banner.rb:49]:
Expected "User Access Verification\n" to be nil.

1 runs, 4 assertions, 1 failures, 0 errors, 0 skips
```

The behavior on my devcies is as follows.

```
n9k# conf t
Enter configuration commands, one per line. End with CNTL/Z.
n9k(config)# no banner motd
n9k(config)# 
n9k(config)# end
n9k# 
n9k# show banner motd 
User Access Verification
n9k# 
```

I see that you are setting the default to this but asserting for nil?

Should this be using `refute_nil` instead?

mikewiebe: Or.. why not check that it is set to the default value?

tests/test_banner.rb
```diff
@@ -0,0 +1,62 @@
...
+    assert_equal(Cisco::Banner.banners['default'].motd,
+                 'Test banner!')
+    assert_equal(Cisco::Banner.banners['default'].motd,
+                 banner.motd)
+
+    banner.motd = nil
+    assert_nil(Cisco::Banner.banners['default'].motd)
+    assert_nil(banner.motd)
+  end
+end
```
mikewiebe: It would be nice to have a multiline banner check test for versions `>=` `7.0(3)I7(4)`

shermdog: Do you have any examples handy on fencing this for that release?  Don't want to run up against unuspported platforms/versions.

mikewiebe: We can use the `step_unless_legacy_defect()` method.

Here is an example:

https://github.com/cisco/cisco-network-node-utils/blob/3b254e740769b700e15000196be5dd19065b2c75/tests/test_tacacs_global.rb#L80

Versions that we support and need to skip would be

```ruby
skip_versions = ['7.0.3.I[2-6]', '7.0.3.I7.[1-3]', '7.3', '8.[1-3]']
```

The fix for the n7k will likely go into `8.4`

mikewiebe: Don't worry about listing a defect number for now.  Just add a message indicating that multiline banners are not supported.

mikewiebe: Alternatively we could extend https://github.com/cisco/cisco-network-node-utils/blob/88e0d96a1014e79b12681d337ea4397c4ef55040/tests/ciscotest.rb#L225 to skip a list of versions since the intent of `step_unless_legacy_defect()` is to skip a specific step within a test case.

tests/test_banner.rb
shermdog: Good catch - missed updating this when I changed the behavior :)

shermdog:
@mikewiebe fixed the broken test, extended `skip_legacy_defect` to skip array of versions and added multiline banner test.

Tested against: `7.0(3)I7(1)`, `7.0(3)I7(4)`, `9.2(1)`

tests/test_banner.rb
```diff
@@ -0,0 +1,85 @@
...
+                 banner.motd)
+
+    banner.motd = nil
+    assert_equal(banner.default_motd, Cisco::Banner.banners['default'].motd)
+    assert_equal(banner.default_motd, banner.motd)
+  end
+
+  def test_multiline_motd
+    skip_versions = ['7.0.3.I[2-6]', '7.0.3.I7.[1-3]', '7.3', '8.[1-3]']
+    skip_legacy_defect?(skip_versions, '')
```
mikewiebe: Please update the message to be something along the lines of: `multiline banner configuration using nxapi not supported`

mikewiebe:
@shermdog One minor comment and then should be good to merge

PR #589, opened 2018-08-23
(CISCO-75/76) Fix interface_hsrp_group

shermdog:
Removed default values for authentication.  There are not defaults that are sane in this context.  Users should supply explicit values.

### DIFF
--- a/lib/cisco_node_utils/interface_hsrp_group.rb
+++ b/lib/cisco_node_utils/interface_hsrp_group.rb
@@ -1,7 +1,7 @@
 #
 # October 2016, Sai Chintalapudi
 #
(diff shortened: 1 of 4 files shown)


lib/cisco_node_utils/interface_hsrp_group.rb
```diff
@@ -96,6 +96,12 @@ def destroy
     # authentication md5 key-string ABCXYZ timeout 22
     # authentication md5 key-string 7 12345678901234567890 timeout 22
     # authentication md5 key-string 7 123456789 compatibility timeout 22
+    # if passwd
```
mikewiebe, line 99: Is this comment block intentional?

PR #593, opened 2018-08-30
(FM-7109) Add ability to dynamically specify environment

shermdog:
This PR adds basic functionality for passing in environment data at run time
This is the building block for Puppet Device support, but should not affect other consumers of the gem.

### DIFF
--- a/ext/mkrf_conf.rb
+++ b/ext/mkrf_conf.rb
@@ -37,7 +37,7 @@ begin
   os == 'ios_xr' || deps << Gem::Dependency.new('net_http_unix',
                                                 '~> 0.2', '>= 0.2.1')
   # NX-OS doesn't need gRPC
-  os == 'nexus' || deps << Gem::Dependency.new('grpc', '~> 0.12')
+  os == 'nexus' || deps << Gem::Dependency.new('grpc', '~> 1.14.1')
 
   deps.each do |dep|
     installed = dep.matching_specs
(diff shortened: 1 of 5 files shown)


shermdog:
cc @willmeek

willmeek (approved):
Great work @shermdog

ext/mkrf_conf.rb
```diff
@@ -37,7 +37,7 @@
   os == 'ios_xr' || deps << Gem::Dependency.new('net_http_unix',
                                                 '~> 0.2', '>= 0.2.1')
   # NX-OS doesn't need gRPC
-  os == 'nexus' || deps << Gem::Dependency.new('grpc', '~> 0.12')
+  os == 'nexus' || deps << Gem::Dependency.new('grpc', '~> 1.14.1')
```
mikewiebe, line 40: Are you testing XR in your environment?

mikewiebe:
@shermdog Looks good.  Just a question about XR support.

ext/mkrf_conf.rb
shermdog: We are not testing / using this code w/ XR at all (to my knowledge)

mikewiebe: Ok.. just wondering why the version bump?  We have considered removing support for this.

shermdog: When using Puppet Device it installs both gems.  This is basically a temporary fix to unblock agentless.  I agree we should probably remove the gRPC / XR stuff in the future.

PR #595, opened 2018-09-06
CISCO-74 Fix puppet resource interface defaults

mikewiebe:
Fixes problem where `cisco_interface` returns illegal default values

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/interface.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface.yaml
@@ -258,7 +258,7 @@ ipv4_proxy_arp:
 ipv4_redirects_loopback:
   kind: boolean
   nexus:
-    default_only: false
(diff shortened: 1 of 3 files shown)


lib/cisco_node_utils/cmd_ref/interface.yaml
```diff
@@ -384,6 +384,7 @@ purge_config:
   set_value:   "default interface <name>"
   get_command: "show running interface"
   get_value:   '/(.*)/'
+  default_value: ~
```
saichint, line 387: Is there any reason this is needed? The default is not defined here anyway

saichint: Sorry my bad, this is needed. ignore it

saichint:
👍

shermdog (approved):

PR #596, opened 2018-09-07
Fix for Cisco-73

saichint:
This PR is for fixing issue cisco-73
https://tickets.puppetlabs.com/browse/CISCO-73

Basically the current ace provider implementation does not take care of icmp properly.
However, upon further inspection of CLIs, it is found that the CLIs differ between different versions of software and also between platforms.
* ```redirect``` can be just a keyword or it can be used with interfaces or they could both be used simultaneously.
* the position of ```redirect``` changes on the show run based on the software version.
* ```fragments``` is put a different location than other options even though the behavior is similar to other options.
* ```set_erspan_gre_proto``` and ```set_erspan_dscp``` positions change based on the software versions.

Due to all these weird quirks, the code checks some of the parameters on 2nd pass due to hit and miss regex patterns for the ```show run aclmgr``` output.

* For all the single word icmp parameter options, the name ```proto_option``` (instead of icmp...something) is given because this can be used to other protocols going forward and also because for setting the configuration the same pattern is used for all protocols (unlike get regex where it is different for icmp). 
* However, only ```icmp``` supports the ```proto_option``` in the near term.

Tested on n7k, n3k and n9k.

### DIFF
--- a/lib/cisco_node_utils/ace.rb
+++ b/lib/cisco_node_utils/ace.rb
@@ -105,18 +105,44 @@ module Cisco
                  ' *(?<proto>\d+|\S+)'\
                  ' *(?<src_addr>any|host \S+|[:\.0-9a-fA-F]+ [:\.0-9a-fA-F]+|[:\.0-9a-fA-F]+\/\d+|addrgroup \S+)'\
                  ' *(?<dst_addr>any|host \S+|[:\.0-9a-fA-F]+ [:\.0-9a-fA-F]+|[:\.0-9a-fA-F]+\/\d+|addrgroup \S+)'\
+                 ' *(?<proto_option>\S+)?'\
+                 ' *(?<precedence>precedence \S+)?'\
                  ' *(?<redirect>redirect \S+)?'\
                  ' *(?<dscp>dscp \S+)?'\
-                 ' *(?<packet_length>packet-length (range \d+ \d+|(lt|eq|gt|neq) \d+))?'\
-                 ' *(?<precedence>precedence \S+)?'\
-                 ' *(?<set_erspan_dscp>set-erspan-dscp \d+)?'\
-                 ' *(?<set_erspan_gre_proto>set-erspan-gre-proto \d+)?'\
                  ' *(?<time_range>time-range \S+)?'\
+                 ' *(?<packet_length>packet-length (range \d+ \d+|(lt|eq|gt|neq) \d+))?'\
                  ' *(?<ttl>ttl \d+)?'\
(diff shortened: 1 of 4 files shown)


lib/cisco_node_utils/ace.rb
```diff
@@ -74,6 +74,9 @@ def ace_get
       remark = Regexp.new('(?<seqno>\d+) remark (?<remark>.*)').match(str)
       return remark unless remark.nil?
 
+      # for icmp things are different
```
mikewiebe: *nit* `Specialized icmp protocol handling`

shermdog (changes requested):
Squash the commits and I'm :+1:

shermdog (changes requested):
Squash the commits and I'm :+1:

saichint:
@shermdog waiting for Rahul to give 👍

lib/cisco_node_utils/ace.rb
```diff
@@ -96,6 +99,60 @@ def ace_get
       regexp.match(str)
     end
 
+    # icmp ace getter
+    def icmp_ace_get(str)
+      # rubocop:disable Metrics/LineLength
```
rahushen, line 104: Do we need this ?

lib/cisco_node_utils/ace.rb
```diff
@@ -96,6 +99,60 @@ def ace_get
...
+          return temp if check_redirect_repeat(str)
+          return regexp_no_proto_option.match(str)
+        end
+      # the reserved keywords check
+      elsif po.strip.match(/precedence$|dscp$|time-range$|packet-length$|ttl$|vlan$|set-erspan-gre-proto$|set-erspan-dscp$|log$/)
+        return regexp_no_proto_option.match(str)
+      else
+        return temp
+      end
+      # rubocop:enable Metrics/LineLength
```
rahushen, line 153: Same as above.

rahushen (commented):
Looks good. Minor comments

lib/cisco_node_utils/ace.rb
saichint: Sorry, I am not following the comment. Are you asking if this method is needed? The answer is yes.

saichint: If you are asking if the disable line length is needed, then yes it is required because of the regex line length more than 80 chars (just like the ace_get method where the original regex pattern is defined.

lib/cisco_node_utils/ace.rb
saichint: Please see comment above

PR #598, opened 2018-09-25
Fix interface default issues

saichint:
When the interface defaults were defined, there are few issues creeped in.
1. couple of typos
2. test issues where the interface needs to be put in switchport_mode for the corresponding properties to test.

### DIFF
--- a/tests/test_interface.rb
+++ b/tests/test_interface.rb
@@ -346,6 +346,8 @@ class TestInterface < CiscoTestCase
     # Validate the collection
     inttype_h.each_value do |v|
       interface = v[:interface]
+      next if interface.name[/vlan|loopback/]
+      interface.switchport_mode = :access
 
       assert_equal(v[:access_vlan], interface.access_vlan,
                    'Error: Access vlan value not correct')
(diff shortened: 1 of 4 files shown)


tests/test_interface.rb
```diff
@@ -1328,8 +1330,6 @@ def interface_test_data
         default_shutdown:    true,
         switchport:          :disabled,
         default_switchport:  :disabled,
-        access_vlan:         DEFAULT_IF_ACCESS_VLAN,
```
mikewiebe, line 1331: Why are you removing these?

mikewiebe:
@saichint One comment, otherwise 👍

mikewiebe:
@saichint Please add a note indicating which platforms/image versions you tested.

tests/test_interface.rb
saichint: For loopback and vlan interfaces, access_vlan is not an option. For ethernet and port-channel it is available if the switchport mode is turned on.

saichint:
Tested on n9k running 9.2.1 image and on n7k running 8.3.1 image

PR #599, opened 2018-10-12
Fixes for n9k-f and n3k-f

saichint:
* This PR is for fixing minitests and software versioning scheme which changed from older versions to latest versions on n3k-f and n9k-f.
* Some of the features like vtp, acl, vxlan_vtp_vni etc. have a different behavior for these platforms and this is addressed.

All tests pass on n9k, n9k-f, n3k-f

### DIFF
--- a/lib/cisco_node_utils/node.rb
+++ b/lib/cisco_node_utils/node.rb
@@ -358,20 +358,13 @@ module Cisco
     def prod_qualifier(prod, inventory)
       case prod
       when /N9K/
-        # Two datapoints are used to determine if the current n9k
+        # one datapoint is used to determine if the current n9k
         # platform is a fretta based n9k or non-fretta.
         #
-        # 1) Image Version == 7.0(3)F*
-        # 2) Fabric Module == N9K-C9*-FM-R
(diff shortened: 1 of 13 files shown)


lib/cisco_node_utils/node.rb
```diff
@@ -358,32 +358,27 @@ def product_id
     def prod_qualifier(prod, inventory)
       case prod
       when /N9K/
-        # Two datapoints are used to determine if the current n9k
+        # one datapoint is used to determine if the current n9k
```
mikewiebe: You should be able to collapse this into a single code block:

```ruby
   when /N(3|9)K/
     # one datapoint is used to determine if the current n(3|9)k
     # platform is a fretta based n(3|9)k or non-fretta.
     #
     # Module == *FM-R
     inventory.each do |row|
       if row['productid'][/(-R|FM-R)/]
         # Append -F for fretta platform.
         return prod.concat('-F') unless prod[/-F/]
       end
     end
   end
```

mikewiebe:
One comment, otherwise 👍

PR #600, opened 2018-11-14
Add new CLIs for Vxlan Vtep 

saichint:
This PR is for adding new CLIs for vxlan_vtep and vxlan_vtep_vni. 
Tested on all relevant platforms and they are ok.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
+++ b/lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
@@ -12,6 +12,36 @@ all_interfaces:
   get_context: ~
   get_value: '/^interface (.*)$/'
 
+global_ingress_replication_bgp:
+  _exclude: [N3k-F, N5k, N6k, N7k, N9k-F]
+  os_version: 'N9k:9.2'
+  kind: boolean
+  get_value: '/^global ingress-replication protocol bgp$/'
+  set_value: '<state> global ingress-replication protocol bgp'
+  default_value: false
+
+global_mcast_group_l2:
+  _exclude: [N3k-F, N5k, N6k, N7k]
+  os_version: 'N9k, N9k-F:9.2'
+  get_value: '/^global mcast-group (\S+) L2$/'
+  set_value: '<state> global mcast-group <ip> L2'
+  default_value: ~
+
+global_mcast_group_l3:
+  _exclude: [N3k-F, N5k, N6k, N7k, N9k-F]
+  os_version: 'N9k:9.2'
+  get_value: '/^global mcast-group (\S+) L3$/'
+  set_value: '<state> global mcast-group <ip> L3'
+  default_value: ~
+
+global_suppress_arp:
+  _exclude: [N3k-F, N5k, N6k, N7k]
+  os_version: 'N9k, N9k-F:9.2'
+  kind: boolean
+  get_value: '/^global suppress-arp$/'
+  set_value: '<state> global suppress-arp'
+  default_value: false
+
 host_reachability:
   get_value: '/^host-reachability protocol (\S+)/'
   set_value: '<state> host-reachability protocol <proto>'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,20 @@
 Changelog
 =========
 
+## [v1.11.0]
+
+### New Cisco Resources
+
+### Added
+* Extend vxlan_vtep_vni with attribute:
+  * `suppress_arp_disable`
+
+### Changed
+
+### Removed
+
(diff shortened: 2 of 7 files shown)


lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
```diff
@@ -12,6 +12,36 @@ all_interfaces:
   get_context: ~
   get_value: '/^interface (.*)$/'
 
+global_ingress_replication_bgp:
+  _exclude: [N3k-F, N5k, N6k, N7k, N9k-F]
+  os_version: 'N9k:9.2'
```
mikewiebe, line 17: This should be `N9k:9.1` right?

lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
```diff
@@ -12,6 +12,36 @@ all_interfaces:
...
+  kind: boolean
+  get_value: '/^global ingress-replication protocol bgp$/'
+  set_value: '<state> global ingress-replication protocol bgp'
+  default_value: false
+
+global_mcast_group_l2:
+  _exclude: [N3k-F, N5k, N6k, N7k]
+  os_version: 'N9k, N9k-F:9.2'
```
mikewiebe, line 25: I don't understand why this property is different for the `_exclude` and `os_version`

lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
```diff
@@ -12,6 +12,36 @@ all_interfaces:
...
+global_mcast_group_l3:
+  _exclude: [N3k-F, N5k, N6k, N7k, N9k-F]
+  os_version: 'N9k:9.2'
+  get_value: '/^global mcast-group (\S+) L3$/'
+  set_value: '<state> global mcast-group <ip> L3'
+  default_value: ~
+
+global_suppress_arp:
+  _exclude: [N3k-F, N5k, N6k, N7k]
+  os_version: 'N9k, N9k-F:9.2'
```
mikewiebe, line 39: Same comment here about the difference and why.

lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
saichint: I am not sure whether there is anything like 9.1
I know there is only 9.2.1 and 9.2.2
Please check again

lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
saichint: The exclude option says that this property is not applicable for N3k-F, N5k, N6k and N7k. It is applicable to N9k and N9k-F, however, only if this platform is running version 9.2 or above. The os_version is for checking the minimum version capability.

lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
saichint: See above

lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
mikewiebe: Good point. 👍

lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
mikewiebe: Good point 👍

lib/cisco_node_utils/cmd_ref/vxlan_vtep.yaml
mikewiebe: That I understand.  The question I have is... Is this property supported on N9k running versions older then `9.2`?

saichint: No. only from 9.2

mikewiebe: I don't remember the way the `os_version` string lookup works but I thought that you had to specify a version string for each platform.

`os_version: 'N9k:9.2, N9k-F:9.2`

saichint: Not really. If the version is different for different platforms, then what you say is correct. If the version is same for multiple platforms, you can use comma notation.
Check this:
https://github.com/cisco/cisco-network-node-utils/blob/develop/lib/cisco_node_utils/cmd_ref/vpc.yaml#L98

PR #601, opened 2018-11-20
Allow node cache to be reset.

da-ar:
As each class inherits directly from `node_util`, each class has its own cache of `node`. This removes the per class cache.
Additionally a reset has been added that can be called prior to `Environment` allowing the new environment settings to be used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+## [Unreleased]
+
+### Added
+  * `reset_instance` method to node. Allows a single instance of nodeutils to reset the environment cache.
+
+### Removed
+  * Removed cache in `node_util.node`, which gave every inheriting class it's own cache.
+
 ## [v1.10.0]
 
 ### New Cisco Resources
--- a/lib/cisco_node_utils/node.rb
+++ b/lib/cisco_node_utils/node.rb
@@ -237,6 +237,11 @@ module Cisco
       @instance ||= new
     end
 
+    # Allow instance cache to be reset
+    def self.reset_instance
+      @instance = nil
+    end
+
     def initialize
       @client = Cisco::Client.create
       @cmd_ref = nil
--- a/lib/cisco_node_utils/node_util.rb
+++ b/lib/cisco_node_utils/node_util.rb
@@ -21,7 +21,7 @@ module Cisco
   # NodeUtil - generic functionality for node utility subclasses to use
   class NodeUtil
     def self.node
-      @node ||= Cisco::Node.instance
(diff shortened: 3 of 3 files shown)


da-ar:
How this will be used: https://github.com/da-ar/cisco-network-puppet-module/commit/c0b320d09be40eb5a6455074982152406bb2158a

willmeek (approved):

mikewiebe:
@da-ar Not sure I fully understand the reason for this change.  Can you provide more context?  Thanks!

da-ar:
@mikewiebe So the reason for this change is that the module will only be instantiated once and used to access many different devices.  The current code allows multiple environments to be created but you cannot specify the specific connection you want to use and class based caches mean that old connections are retained when an environment is changed. This PR will allow a previously set environment to be manually cleared and a new environment to take its place.

For example, this would break like this:

```ruby
Cisco::Environment.add_new('default', host1)
Cisco::DomainNames.domainnames['something'] = value1
#connect to second device
Cisco::Environment.add_new('default', host2)
Cisco::DomainNames.domainnames['something'] = value2 # changes are made to host1
```

Take the following code path:
domain_name.domainnames > config_get > NodeUtil.config_get > node.config_get

`node` is [Cisco::Node.instance](https://github.com/cisco/cisco-network-node-utils/blob/4e0e66d76dded6aa437993fd04748dde3db6ffaa/lib/cisco_node_utils/node_util.rb#L24) 

Which is [Cisco::Node.new](https://github.com/cisco/cisco-network-node-utils/blob/4e0e66d76dded6aa437993fd04748dde3db6ffaa/lib/cisco_node_utils/node.rb#L236)

`node.config_get` calls out to `get` which accesses [@client.get](https://github.com/cisco/cisco-network-node-utils/blob/4e0e66d76dded6aa437993fd04748dde3db6ffaa/lib/cisco_node_utils/node.rb#L288). `@client` is set when `Cisco::Node` is [initalized](https://github.com/cisco/cisco-network-node-utils/blob/4e0e66d76dded6aa437993fd04748dde3db6ffaa/lib/cisco_node_utils/node.rb#L241) and does not pass an environment, so it [defaults to nil](https://github.com/cisco/cisco-network-node-utils/blob/4e0e66d76dded6aa437993fd04748dde3db6ffaa/lib/cisco_node_utils/client/client.rb#L85)

mikewiebe:
@da-ar Thanks for the explanation.  Looks pretty straight forward but I kicked of a test against our internal testbed just to be safe.  I will have results in the morning and will post back here.

mikewiebe:
@da-ar Test results looked good. 👍

PR #605, opened 2019-01-08
Extend nxapi client with https support

Thomas-Franklin:
* `transport` will control if `use_ssl` is true or not
* `verify_mode` has been added for when https mode is used, it selects based on `verify_mode` param being set to either 'none', 'peer', 'fail-no-peer' and 'client-once'

The following environment tests were used to verify https/http and host:port combinations.

```
nxapi_remote_http:
  host: localhost
  port: 7070
  telnet_port: 2222
  username: vagrant
  password: vagrant
host_and_port:
  host: localhost:7070
  telnet_port: 2222
  username: vagrant
  password: vagrant
nxapi_remote_https:
  host: localhost
  telnet_port: 2222
  transport: https
  port: 4431
  transport: https
  username: vagrant
  password: vagrant
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Changelog
 ### New Cisco Resources
 
 ### Added
+* Extend nxapi client for https support
+   * `use_ssl` will be true when `transport` is `https`
+   * now makes use of `port` for custom nxapi ports
 * Extend router_ospf_vrf with attribute:
    * `redistribute`
 * `reset_instance` method to node. Allows a single instance of nodeutils to reset the environment cache.
--- a/tests/basetest.rb
+++ b/tests/basetest.rb
@@ -85,6 +85,38 @@ class TestCase < Minitest::Test
     self.class.password
   end
 
+  def self.telnet_port
+    Cisco::Environment.environment[:telnet_port] || 23
+  end
+
+  def telnet_port
+    self.class.telnet_port
+  end
+
+  def self.port
+    Cisco::Environment.environment[:port] || address.split(':')[1]
+  end
+
+  def port
+    self.class.port
+  end
+
+  def self.transport
+    Cisco::Environment.environment[:transport] || 'http'
+  end
+
+  def transport
+    self.class.transport
+  end
+
+  def self.verify_mode
+    Cisco::Environment.environment[:verify_mode] || 'none'
+  end
+
+  def verify_mode
+    self.class.verify_mode
+  end
+
   def setup
     # Hack - populate environment from user-entered values from basetest.rb
     if Cisco::Environment.environments.empty?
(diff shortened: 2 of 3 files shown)


willmeek (approved):

mikewiebe:
@Thomas-Franklin Thanks for getting this PR open.  I will review today.

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ Changelog
 ### New Cisco Resources
 
 ### Added
+* Extend nxapi client for https support
+   * `use_ssl` will be true when `transport` is `https`
+   * `verify_mode` controls the OpenSSL session verification,
+      - 'none' for `OpenSSL::SSL::VERIFY_NONE`
+      - 'peer' for `OpenSSL::SSL::VERIFY_PEER`
+      - 'client-once' for `OpenSSL::SSL::VERIFY_CLIENT_ONCE`
```
mikewiebe: @Thomas-Franklin Did have a chance to test/verify these different modes?  I only used `OpenSSL::SSL::VERIFY_NONE`.   We need to make sure and test all of the modes before claming support for each.

tests/basetest.rb
```diff
@@ -85,20 +85,56 @@ def password
     self.class.password
   end
 
+  def self.telnet_port
```
mikewiebe, line 88: Why are you exposing a telnet port?  Are you using a non-default port in your test environment?

mikewiebe:
@Thomas-Franklin I would like to add a set of minitests to cover the following scenarios.  Specifically the following scenarios.

1. transport: http / https
2. port: deafult and non-default port
3. verify_modes

I was thinking something along the lines of `tests/test_nxapi_transport.rb` that performs a simple test across each.  If you prefer to do it in a different PR we can merge this and add the tests in a separate PR.

tests/basetest.rb
Thomas-Franklin: Yea I am using non-default in my local test environment

CHANGELOG.md
Thomas-Franklin: Unfortunately I have not been able too, will try and do further testing tomorrow and see if I can test the alternative modes and add the minitests suggested as part of this PR

Thomas-Franklin:
> @Thomas-Franklin I would like to add a set of minitests to cover the following scenarios. Specifically the following scenarios.
> 
> 1. transport: http / https
> 2. port: deafult and non-default port
> 3. verify_modes
> 
> I was thinking something along the lines of `tests/test_nxapi_transport.rb` that performs a simple test across each. If you prefer to do it in a different PR we can merge this and add the tests in a separate PR.

For the requested tests just wondering what you would expect to be covered in `tests/test_nxapi_transport.rb`, my understanding on the minitests were that it is driven by an environment file, i.e. 

```
http:
  host: <host>
  username: user
  password: pass

https:
  host: <host>
  transport: https
  username: user
  password: pass
```

Then any test could be ran with the `-e <environment>` param, so `bundle exec ruby tests/file.rb -e https` would run the specified test file using the https transport.

mikewiebe:
> > @Thomas-Franklin I would like to add a set of minitests to cover the following scenarios. Specifically the following scenarios.
> > 
> > 1. transport: http / https
> > 2. port: deafult and non-default port
> > 3. verify_modes
> > 
> > I was thinking something along the lines of `tests/test_nxapi_transport.rb` that performs a simple test across each. If you prefer to do it in a different PR we can merge this and add the tests in a separate PR.
> 
> For the requested tests just wondering what you would expect to be covered in `tests/test_nxapi_transport.rb`, my understanding on the minitests were that it is driven by an environment file, i.e.
> 
> ```
> http:
>   host: <host>
>   username: user
>   password: pass
> 
> https:
>   host: <host>
>   transport: https
>   username: user
>   password: pass
> ```
> Then any test could be ran with the `-e <environment>` param, so `bundle exec ruby tests/file.rb -e https` would run the specified test file using the https transport.

@Thomas-Franklin Correct. I suppose you are right.  I was envisioning a simple test that uses ospf as an example and cycles through the various transports but it probably makes more sense to just add this to our CI environment and test one of the suites against the different options from canned environments.

PR #607, opened 2019-01-15
(maint) Bump Ruby version 2.4.1

shermdog:
This commit bumps the minimum Ruby version to 2.4.1
Also removed deprecated Fixnum

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,11 +4,12 @@ sudo: false
 language: ruby
 cache: bundler
 rvm:
-  - 2.2.2
-  - 2.1.6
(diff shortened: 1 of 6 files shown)


shermdog:
I haven't run this through any testing or CI, but wanted to get it into the review cycle.  Users always get hung up on warnings :)

With the EOL of Puppet 4.x the minimum version of Ruby we support now is 2.4.1

shermdog:
Looks like Travis changed Bundler behavior recently: https://docs.travis-ci.com/user/languages/ruby/#bundler-20

Also updated to take that into account.

PR #584, opened 2018-06-28
(FM-7136) exclude add_quotes from entire 7.3 series

shermdog:
There was an exclusion set in add_quotes for 7.3.0 in #519.  It appears that this exclusion needs to apply to later version of 7.3:

```

[root@localhost ~]# cat tac_global.pp
--
tacacs_global { 'default':
key              => 'wwgfus-pwb',
key_format       => '7',
source_interface => ['mgmt0'],
timeout          => '7',
}
```
```
[root@localhost ~]# puppet apply -d tac_global.pp
 
Notice: /Stage[main]/Main/Tacacs_global[default]/key: key changed 'unset' to 'wwgfus-pwb'
Notice: /Stage[main]/Main/Tacacs_global[default]/key_format: key_format changed '' to '7'
 
Debug: Set state using data format 'cli'
Debug:   to value(s):
tacacs-server key 7 "wwgfus-pwb"
Debug: Input (cli_conf): ' tacacs-server key 7 "wwgfus-pwb"'
Debug: Sending HTTP request to NX-API at localhost:
{"accept-encoding"=>["gzip;q=1.0,deflate;q=0.6,identity;q=0.3"], "accept"=>["*/*"], "user-agent"=>["Ruby"], "cookie"=>["nxapi_auth=admin:local"], "content-type"=>["application/json"]}
{"ins_api":{"version":"1.0","type":"cli_conf","chunk":"0","sid":"1","input":" tacacs-server key 7 \"wwgfus-pwb\"","output_format":"json"}}
Debug: HTTP Response: OK
{
"ins_api":	{
"type":	"cli_conf",
"version":	"1.2",
"sid":	"eoc",
"outputs":	{
"output":	{
"msg":	"Success",
"code":	"200",
"body":	{
}
}
}
}
}
Debug: Result for ' tacacs-server key 7 "wwgfus-pwb"': Success
```
```
[root@localhost ~]# puppet resource tacacs_global
--
tacacs_global { 'default':
key              => '"\ wwgfus-pwb\"',
key_format       => '7',
source_interface => ['mgmt0'],
timeout          => '7',
}
```
```
cisco-c5672up# sh run tacacs+ all
--
 
!Command: show running-config tacacs+ all
!Time: Thu Jun 28 21:38:19 2018
 
version 7.3(2)N1(1)
feature tacacs+
 
tacacs-server key 7 "\"wwgfus-pwb\""
```

@mikewiebe can you folks test this on your side against a few different 7.3 trains?

PR #609, opened 2019-01-17
Minitest fixes

chrisvanheuveln:
Most of the fixes include version checks and/or syntax handling fixes for various versions.

### DIFF
--- a/tests/test_tacacs_global.rb
+++ b/tests/test_tacacs_global.rb
@@ -75,7 +75,9 @@ class TestTacacsGlobal < CiscoTestCase
     global.encryption_key_set(nil, key)
     # Device encypts key - verify return value
     assert_equal(7, global.key_format)
-    assert_equal('"WAWY_NZB"', global.key)
+    key = 'WAWY_NZB'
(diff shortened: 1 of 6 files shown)


tests/test_tacacs_global.rb
```diff
@@ -75,7 +75,9 @@ def test_tacacs_global
     global.encryption_key_set(nil, key)
     # Device encypts key - verify return value
     assert_equal(7, global.key_format)
-    assert_equal('"WAWY_NZB"', global.key)
+    key = 'WAWY_NZB'
+    key = "\"#{key}\"" unless Platform.image_version[/7\.3/]
```
chrisvanheuveln, line 79: This is wrong. The getter should always add quotes and it's not for some reason. I'll debug and fix.

chrisvanheuveln: Nevermind. My initial thoughts were correct and my changes are fine. No quotes for 7.3 images.

Ref: #519 #584

mikewiebe:
👍

PR #611, opened 2019-02-01
Rel200/fix radius global

mikewiebe:


mikewiebe:
@chrisvanheuveln approves!

PR #612, opened 2019-02-04
Add C3064 exclude for feature-set fex

mikewiebe:
The `install feature-set fex` command is excluded on all of our n3k platforms except `C3048`.  This command is not supported on the 3048 platform so this defines the platform and adds the exclude in `feature.yaml`.


chrisvanheuveln:
You're actually excluding `3048`

mikewiebe:
> You're actually excluding `3048`

Thanks.. I updated the summary for historical accuracy.

PR #618, opened 2019-02-06
Minitest: router_bgp fixes for n6k

chrisvanheuveln:
* `rd` needs `nv overlay evpn` to expose cli
* nexus does not support nsr
* added a common method to check for attrs that are excluded from cmd_ref
  * `cmd_ref.lookup` returns type `UnsupportedCmdRef`

### DIFF
--- a/tests/ciscotest.rb
+++ b/tests/ciscotest.rb
@@ -214,6 +214,12 @@ class CiscoTestCase < TestCase
       Gem::Version.new(lim) < Gem::Version.new(ver)
   end
 
+  def skip_if_UnsupportedCmdRef(feature_name, attribute_name) # rubocop:disable Style/MethodName
+    # Check if attr is excluded by the cmd_ref yaml
+    skip("UnsupportedCmdRef attribute: '#{attribute_name}'") if
+      cmd_ref.lookup(feature_name, attribute_name).is_a?(UnsupportedCmdRef)
+  end
+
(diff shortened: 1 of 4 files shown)


tests/ciscotest.rb
```diff
@@ -214,6 +214,12 @@ def skip_incompat_version?(feature, property)
       Gem::Version.new(lim) < Gem::Version.new(ver)
   end
 
+  def skip_if_UnsupportedCmdRef(feature_name, attribute_name) # rubocop:disable Style/MethodName
```
mikewiebe, line 217: This is a nice addition.

tests/test_router_bgp.rb
```diff
@@ -256,24 +257,18 @@ def test_asnum_dot
   end
 
   def test_nsr
+    skip_if_UnsupportedCmdRef('bgp', 'nsr')
```
mikewiebe, line 260: Looks like the nsr stuff was added by the XR team when they were still doing puppet.  I see that NSR is supported for  OSPFv2 on Nexus but there is no mention of BGP.  Did you find documentation that states it's not supported for BGP or is it the absence of the CLI?

chrisvanheuveln: Neither. Our yaml excludes it so I added this check to obey the yaml. (sushirol exluded it from nexus when he added the attr)

PR #619, opened 2019-02-07
minitest n6k: test_vni dependency

chrisvanheuveln:
* n6k has feature-set fabricpath dependency
* Added getter/setter to feature.rb
* test_vn_segment_vlan_based was being skipped before; added fabricpath_enable to it and now it runs on n6k

Tested both testcases on n6k, n7k, n9k

### DIFF
--- a/tests/test_feature.rb
+++ b/tests/test_feature.rb
@@ -173,7 +173,8 @@ class TestFeature < CiscoTestCase
 
   def test_vn_segment_vlan_based
     vxlan_linecard?
-    Feature.nv_overlay_enable unless node.product_id[/N3/]
+    Feature.fabricpath_enable if node.product_id[/N(5|6)/]
+    Feature.nv_overlay_enable unless node.product_id[/N(3|5|6)/]
     feature('vn_segment_vlan_based')
   rescue RuntimeError => e
     hardware_supports_feature?(e.message)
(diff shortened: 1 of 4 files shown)


tests/test_feature.rb
```diff
@@ -222,6 +224,28 @@ def test_feature_set_fabric
...
+      assert_nil(Feature.fabricpath_enabled?)
+      assert_raises(Cisco::UnsupportedError) { Feature.fabricpath_enable }
+      return
+    end
+    vdc_limit_f3_no_intf_needed(:set)
+    fs = 'feature-set fabricpath'
+    # Get current state of the feature-set
+    feature_set_installed = Feature.fabricpath_installed?
+
+    config("no #{fs} ; no install #{fs}") if feature_set_installed
```
mikewiebe, line 238: *nit* Why not just call `Feature.fabricpath_installed?` directly here?

mikewiebe (approved):
Minor nit.

tests/test_feature.rb
chrisvanheuveln: This test is just a copy/paste/rename from the other feature-set tests. I *think* the reason I used `config()` here back in the day was  because it's marginally faster to send both commands at once.

mikewiebe: I was not concerned about `config()`.  I was suggesting you get rid of line 235 `feature_set_installed = Feature.fabricpath_installed?` and just do this.

```ruby
config("no #{fs} ; no install #{fs}") if Feature.fabricpath_installed?
```

PR #621, opened 2019-02-07
vxlan minitest fixes

chrisvanheuveln:
I modified `hardware_supports_feature?` to just return a status instead of doing only a skip or flunk. This allows the callers to cache the status and act on it themselves. In the vxlan tests the `setup` takes ~12s (per testcase) to process dependencies before it fails so this change allows us to short-circuit dependency loading and shave a few minutes off of each test overall.

Tested on n6k, n7k, n9k, n3k

### DIFF
--- a/tests/test_vxlan_vtep.rb
+++ b/tests/test_vxlan_vtep.rb
@@ -24,9 +24,11 @@ include Cisco
 class TestVxlanVtep < CiscoTestCase
   @skip_unless_supported = 'vxlan_vtep'
   @@pre_clean_needed = true # rubocop:disable Style/ClassVars
+  @@feature_unsupported = false # rubocop:disable Style/ClassVars
 
   def setup
     super
+    skip('setup: Feature is unsupported on this device') if @@feature_unsupported
     skip('Platform does not support MT-full or MT-lite') unless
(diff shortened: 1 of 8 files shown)


tests/test_vxlan_vtep.rb
```diff
@@ -24,16 +24,22 @@
...
     skip('Platform does not support MT-full or MT-lite') unless
       VxlanVtep.mt_full_support || VxlanVtep.mt_lite_support
     Interface.interfaces(:nve).each { |_nve, obj| obj.destroy }
     vdc_limit_f3_no_intf_needed(:set)
     feature_cleanup if @@pre_clean_needed
     Feature.nv_overlay_enable
     @@pre_clean_needed = false # rubocop:disable Style/ClassVars
+  rescue RuntimeError => e
+    # Skip locally to shortcut the dependencies above for the next test
+    @@feature_unsupported = hardware_supports_feature?(e.message, status_only: true) # rubocop:disable Style/ClassVars
```
mikewiebe, line 41: Ok, so if I am reading this correctly, the first test will still go through the dependency list but then skips will go quickly after that.

chrisvanheuveln: correct

mikewiebe (approved):

PR #626, opened 2019-04-04
(maint) example of a ssh client

Thomas-Franklin:
This client had ran through the acceptance tests of the ciscopuppet module, only noticable issue was the timing, but this is mainly due to how the acceptance tests are small manifests of individual components vs an actual manifest

### DIFF
--- a/lib/cisco_node_utils/client/ssh.rb
+++ b/lib/cisco_node_utils/client/ssh.rb
@@ -12,25 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Namespace for Cisco-specific code
-module Cisco
-  # Namespace for Cisco Client shim
-  class Client
-  end
+require_relative 'client'
+
+# Fail gracefully if submodule dependencies are not met
+begin
+  require 'net/ssh/telnet'
+rescue LoadError => e
+  raise unless e.message =~ /-- net\/ssh\/telnet/
+  # If grpc is not installed, raise an error that client understands.
+  raise LoadError, "Unable to load net/ssh/telnet -- #{e}"
 end
 
-require_relative 'client/client'
-
-# Try to load known extensions
-extensions = ['client/ssh',
-              'client/nxapi',
-              'client/grpc',
-             ]
-extensions.each do |ext|
-  begin
-    require_relative ext
-  rescue LoadError => e
-    # ignore missing client-(grpc|nxapi), they're not always required
-    raise unless e.message =~ /#{Regexp.escape(ext)}/
-  end
+# Namespace for Cisco SSH specific code
+class Cisco::Client::SSH < Cisco::Client
 end
+
+# Auto-load all Ruby files in the subdirectory
+Dir.glob(__dir__ + '/ssh/*.rb') { |file| require file }
--- a/lib/cisco_node_utils/client/ssh/client.rb
+++ b/lib/cisco_node_utils/client/ssh/client.rb
@@ -1,0 +1,167 @@
+# SSH client library.
+#
+# November 2014, Glenn F. Matthews
+#
+# Copyright (c) 2014-2016 Cisco and/or its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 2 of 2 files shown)


lib/cisco_node_utils/client/ssh.rb
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) 2015 Cisco and/or its affiliates.
```
mikewiebe, line 1: *nit* Update Copyright

lib/cisco_node_utils/client/ssh/client.rb
```diff
@@ -0,0 +1,167 @@
+# SSH client library.
+#
+# November 2014, Glenn F. Matthews
```
mikewiebe, line 3: Update Date and Author

lib/cisco_node_utils/client/ssh/client.rb
```diff
@@ -0,0 +1,167 @@
...
+#
+# Copyright (c) 2014-2016 Cisco and/or its affiliates.
```
mikewiebe, line 5: Copyright

lib/cisco_node_utils/client/ssh.rb
```diff
@@ -0,0 +1,31 @@
...
+# limitations under the License.
+
+require_relative 'client'
+
+# Fail gracefully if submodule dependencies are not met
+begin
+  require 'net/ssh/telnet'
+rescue LoadError => e
+  raise unless e.message =~ /-- net\/ssh\/telnet/
+  # If grpc is not installed, raise an error that client understands.
```
mikewiebe, line 22: Comment references grpc.  Cut and paste error?

lib/cisco_node_utils/client/ssh/client.rb
```diff
@@ -0,0 +1,167 @@
...
+  #   used to enter the desired CLI sub-mode
+  # @param values [String, Array<String>] One or more commands
+  #   to enter within the CLI sub-mode.
+  # @param kwargs data-format-specific args
+  def set(data_format: :cli,
+          context: nil,
+          values: nil,
+          **_kwargs)
+    # we don't currently support nxapi_structured for configuration
+    fail Cisco::RequestNotSupported if data_format == :nxapi_structured
```
mikewiebe, line 75: I would prefer that we change this to  `structured_data` that can be used for both `ssh` and `nxapi` or any future clients.  I know this would result in more sweeping changes but we should probably by hiding the connection details from our `get_data_format` keys in the feature yaml files.

The `cach_flush` method also uses terms that are very nxapi specific so I think we should have a larger discussion around how we design for multiple clients.

lib/cisco_node_utils/client/ssh/client.rb
```diff
@@ -0,0 +1,167 @@
...
+    end
+
+    debug("Input (#{type}): \'#{command}\'")
+    if cache_enable? && @cache_hash[type] && @cache_hash[type][command]
+      return @cache_hash[type][command]
+    end
+
+    # send the request and get the response
+    debug("Sending SSH request to device at #{@address}")
+    response = Net::SSH.start(@host, @username, password: @password, port: @port, timeout: @timeout, @verify_host_key => false) do |connection|
```
mikewiebe, line 151: What happens in the case where passwordless SSH is setup between the `puppet device` node and the switch?  Is the `@password` simply ignored in this context if set or `nil`?

mikewiebe (commented):
@Thomas-Franklin Thanks for taking the time to start this work!

lib/cisco_node_utils/client/ssh.rb
Thomas-Franklin: A lot of this was just a direct cut and paste, just to get something thrown together for a quick POC. Feel free to close the PR as we will probably pick it up in the future and address these issues.

lib/cisco_node_utils/client/ssh/client.rb
Thomas-Franklin: I believe it should function the same, and is something we would need to verify. But happy to close this for the time being, as it is something we will be getting time to address in the future hopefully, and just needed something to prove it out for the time-being.

mikewiebe:
Closing this as a reference point for future work.

PR #630, opened 2019-07-08
interface: performance fixes

chrisvanheuveln:
#### Summary
Modified class method `Interface.interfaces` to allow queries for single interfaces.

Prior behavior was to initially capture and cache the state of all interfaces on the device. This approach is acceptable when managing all or most of the interfaces on the device, but performance may be unnecessarily slow when a user is managing only a handful of interfaces on a device that contains many hundreds or even thousands of interfaces.

This design change will be paired with a change to the Puppet module. The intent on the Puppet side is to allow the user to specify an interface count threshold in the manifest, which will determine when to use the cache-all approach vs single interface query.

#### Details
interface perf: `<show_name>` vs `<name>` for proper get_command usage

- Created a unique tag `<show_name>` to differentiate the `<name>` value used by the `get_command:` strings. This was necessary because the `context:` strings also used `<name>`, and not changing this causes the show command string to be interface specific at all times, even when we want the complete list of interfaces.

Add'l info:
 - `Interface.interfaces` uses `'all_interfaces'` `cmd_ref` values, and returns either a hash of all interfaces or a hash with just one interface.
 - The `command_reference` code searches for the same `context:` value in the hash, regardless of hash depth.
 - When there are multiple items in the hash we want to use `get_command:` without an interface specified, because that uses cached data instead of querying the device for each interface.
 - When there is a single interface in the hash we want to use the interface specific `get_command:` form, which returns a small amount of data very quickly.
 - Command output is cached and data is used for all getter calls - as long as each getter specifies the same `get_command:` pattern. If a different show command pattern is used it triggers a new query for the unique get_command string.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/interface.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface.yaml
@@ -3,9 +3,9 @@
 _template:
   context: ["interface <name>"]
   ios_xr:
-    get_command: "show running interface"
+    get_command: "show running interface <name>"
   nexus:
-    get_command: "show running interface all"
+    get_command: "show running interface <name> all"
 
 access_vlan:
   _exclude: [ios_xr]
@@ -20,7 +20,7 @@ all_interfaces:
     get_command: 'show running all-interfaces'
     get_value: '/^interface (.*)/'
   nexus:
-    get_command: "show running-config interface | section '^interface'"
+    get_command: "show running-config interface <name> | section '^interface'"
     get_value: '/(.*)/'
   get_context: ~
 
@@ -35,7 +35,7 @@ capabilities:
   _exclude: [ios_xr]
   multiple:
   get_context: ["<name>"]
-  get_command: "show interface capabilities"
+  get_command: "show interface <name> capabilities"
   get_value: '/(.*)/'
   default_value: [] # :raw default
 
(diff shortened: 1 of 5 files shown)


lib/cisco_node_utils/cmd_ref/interface.yaml
```diff
@@ -3,9 +3,9 @@
 _template:
   context: ["interface <name>"]
   ios_xr:
-    get_command: "show running interface"
+    get_command: "show running interface <name>"
```
mikewiebe, line 6: I would not make any updates to XR.  They are not needed at this point and we have discussed just removing the XR support.

lib/cisco_node_utils/cmd_ref/interface.yaml
```diff
@@ -35,7 +35,7 @@ capabilities:
   _exclude: [ios_xr]
   multiple:
   get_context: ["<name>"]
-  get_command: "show interface capabilities"
+  get_command: "show interface <name> capabilities"
```
mikewiebe, line 38: Why does this one use `<name>` vs `<show_name>`?

lib/cisco_node_utils/cmd_ref/interface.yaml
chrisvanheuveln: ack

lib/cisco_node_utils/cmd_ref/interface.yaml
chrisvanheuveln: This is actually a completely different command from the other `get_command` strings and it also overrides the `_template` `get_command` string. Since it's a different command there is no cache re-use. 
We use the capabilities method to help with minitest / beaker only, and otherwise it's not used. I added the `<name>` field just for efficiency but again it's unrelated to the normal getters. 
ref:   https://github.com/cisco/cisco-network-node-utils/blob/intf-perf-fix/lib/cisco_node_utils/interface.rb#L151

mikewiebe (approved):
Changes look good.

PR #631, opened 2019-07-10
interface_ospf: interface lookup performance fixes

chrisvanheuveln:
#### Summary
Part 2 of the interface lookup performance work.
Ref #630 

- Modified class method `InterfaceOspf.interfaces` to allow queries for single interfaces.
- `interface_ospf.yaml` is one of the early implementations and it was still using positional fields instead of named fields. This created a larger effort to implement the interface performance changes.
- `interface_ospf.rb` is the only consumer of the `InterfaceOspf` class.


lib/cisco_node_utils/interface_ospf.rb
```diff
@@ -23,53 +23,62 @@
...
       fail ArgumentError unless ospf_name.length > 0
       fail ArgumentError unless area.length > 0
 
       # normalize
-      int_name = int_name.downcase
-      @interface = Interface.interfaces[int_name]
-      fail "interface #{int_name} does not exist" if @interface.nil?
+      @intf_name = intf_name.downcase
+      fail "interface #{@intf_name} does not exist" if
+        Interface.interfaces[@intf_name].nil?
```
chrisvanheuveln: This is probably a problem - because it's going to get all interfaces by default. Let me see if there's a solution here.

chrisvanheuveln: Latest commit addresses this issue. You can continue the review now.

lib/cisco_node_utils/interface_ospf.rb
```diff
@@ -143,109 +150,128 @@ def default_message_digest_encryption_type
...
     def hello_interval=(interval)
-      config_set('interface_ospf', 'hello_interval',
-                 @interface.name, '', interval.to_i)
+      # Previous behavior always sets interval and ignores 'no' cmd
+      @set_args.merge!(state: '', interval: interval.to_i)
+      config_set('interface_ospf', 'hello_interval', @set_args)
+      set_args_keys_default
+    end
+
+    def destroy_hello_interval
```
mikewiebe: Minor but can you combine `destroy_hello_interval` and `destroy_dead_interval` into one helper method instead of two?  Code is identical except for `dead_interval`|`hello_interval` string.

chrisvanheuveln: done

chrisvanheuveln:
I hosed the branch on this PR. Creating a new branch and abandoning this PR.

PR #635, opened 2019-07-18
Add Interface.interface_count api

chrisvanheuveln:
This api simply returns a count of interfaces based on what is found in show running.

Tested on all current regression platforms.


mikewiebe:
👍

PR #636, opened 2019-07-22
Interface.interfaces selection failure when 'interface' substrings present

chrisvanheuveln:
Symptom:
- Interface output with an `'interface'` substring causes the `intf_list` to include invalid interface id's.
  - e.g. the following will cause `split` to create an erroneous `interface 'hsrp'` from the line that follows the line with the substring on it:
```
interface portchannel100
  hsrp bfd
  hsrp use-bia scope interface
                     ^^^^^^^^^
  hsrp mac-refresh 350
```

  - result:
```
pry(Cisco::Interface)> intf_list
=> [" port-channel100 hsrp bfd hsrp version 2 hsrp delay minimum 100 reload 200 hsrp use-bia scope ",
 " hsrp mac-refresh 350 ",
   ^^^^ (bug: this is now a separate entry in intf_list)
  ...
```

- Found with `cisco_hsrp/test_interface_hsrp.rb`

### DIFF
--- a/lib/cisco_node_utils/interface.rb
+++ b/lib/cisco_node_utils/interface.rb
@@ -84,8 +84,9 @@ module Cisco
       rescue CliError => e
         # ignore logical interfaces that may not exist yet;
         # invalid interface types should still raise
-        raise unless single_intf &&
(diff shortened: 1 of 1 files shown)


mikewiebe:
👍

PR #637, opened 2019-07-23
interface_ospf: interface lookup performance fixes

chrisvanheuveln:
**Summary**
Part 2 of the interface lookup performance work.
Ref #630

- Modified class method InterfaceOspf.interfaces to allow queries for single interfaces.
- interface_ospf.yaml is one of the early implementations and it was still using positional fields instead of named fields. This created a larger effort to implement the interface performance changes.
- interface_ospf.rb is the only consumer of the InterfaceOspf class.

### DIFF
--- a/lib/cisco_node_utils/cmd_ref/interface_ospf.yaml
+++ b/lib/cisco_node_utils/cmd_ref/interface_ospf.yaml
@@ -63,7 +63,7 @@ message_digest_key_id:
   default_value: 0
 
 message_digest_key_set:
-  set_value: '<state> ip ospf message-digest-key <keyid> <algtype> <enctype> <enc>'
+  set_value: '<state> ip ospf message-digest-key <keyid> <algtype> <enctype> <password>'
 
 message_digest_password:
   default_value: ~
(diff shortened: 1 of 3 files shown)


lib/cisco_node_utils/cmd_ref/interface_ospf.yaml
```diff
@@ -63,7 +63,7 @@ message_digest_key_id:
   default_value: 0
 
 message_digest_key_set:
-  set_value: "%s ip ospf message-digest-key %s %s %s %s"
+  set_value: '<state> ip ospf message-digest-key <keyid> <algtype> <enctype> <enc>'
```
mikewiebe: whats the diff between `enctype` and `enc`?

chrisvanheuveln: `enctype` == encoding type
`enc` == encoding (aka password)

I have no idea why the original coder used that var name. 'password' would have been more descriptive.

mikewiebe: Should we change it to `password`?  Is it more involved then  a search/replace?

chrisvanheuveln: oooooh all right. 
Could also change it to `secret_code_word` or `magic_fancy_word` if that flows better.

mikewiebe (approved):

PR #638, opened 2019-07-23
rescue intf lookups when intf is non-existent

chrisvanheuveln:
When calling the various `interfaces` methods with a `single_intf`, the `config_get` will raise when an interface is not present (e.g. loopbacks, svis, etc).

These methods should return an empty hash when the interface does not exist, just as the previous code did when it collected all interfaces.

Tested on 9k and 7k.


mikewiebe:
👍

PR #642, opened 2019-08-07
Fix interface_ospf setter

mikewiebe:
Fixes a problem where the setter was trying to set the keyid to `0` which is not a supported keyid value.


mikewiebe:
@chrisvanheuveln provided over the shoulder review and approves.

PR #643, opened 2019-08-08
interface.rb: show_name N7k workaround fixes

chrisvanheuveln:
Additional workarounds to solve the problem with using `show run int all | section `^interface <intf_name_pattern>'`

The issue here was that the interface name in the manifest may not match the literal interface name string on the device (note this is only an N7K issue). When doing `show run int <intf_name> all` (on non-N7K's) it doesn't matter if there are case differences; when doing a section command we must match exactly because there is no case-insensitive option for `section`.


chrisvanheuveln:
self-merge since Mike is unavailable.

PR #651, opened 2019-12-12
interface_hsrp_group: interface lookup enhancement

chrisvanheuveln:
The original code collected a list of every interface on the device, then walked each of those looking for hsrp groups. My change uses `show running hsrp` to collect only those interfaces that already have hsrp configs.

UT: minitest & beaker hsrp tests are 100% Pass


mikewiebe:
👍  Changes look fine.  Curious how much faster this is?

chrisvanheuveln:
> how much faster

Hard to say for sure without some large scale tests. This change eliminates the `Interfaces.interfaces` call altogether -- which was parsing and creating objects for every single interface even though we just needed a list of interface names to check -- but with thousands of interfaces I would expect a significant decrease in processing time now.

